PR #396, opened 2014-11-04
Spec that AudioNodes from different contexts can't be connected together

padenot:


padenot:
This contained an unrelated commit, and can go in on its own.

PR #472, opened 2015-01-30
Fix issue 471: Default panner is equal-power.

rtoy:
Primarily fixes the typo in the description of the panningModel
attribute that says HRTF is the default. Also updated a couple of
places where the attribute value was written "equal-power" when the
actual attribute is "equalpower".


padenot:
Thanks, I'll go and write the Gecko patch now.

PR #479, opened 2015-02-06
#6 Support selective disconnection

hoch:
For preview: http://hoch.github.io/web-audio-api/#idl-def-AudioNode

### DIFF
--- a/index.html
+++ b/index.html
@@ -1150,6 +1150,7 @@ way that other EventTargets accept events.
       nodeA.connect(param);
     </pre>
   </dd>
+
   <dt>void disconnect()</dt>
   <dd>
   <dl class=parameters>
@@ -1158,6 +1159,47 @@ way that other EventTargets accept events.
       This parameter is an index describing which output of the
       <a><code>AudioNode</code></a> to disconnect. If this parameter is
       out-of-bounds, an IndexSizeError exception MUST be thrown.
+    </dd>
+  </dl>
+  </dd>
+
+  <dt>void disconnect()</dt>
+  <dd>
+  <dl class=parameters>
+    <dt>AudioNode destination</dt>
+    <dd>
+      The <code>destination</code> parameter is the <a><code>AudioNode</code></a> 
+      to disconnect. If <code>destination</code> <code>AudioNode</code> is not 
+      connected, an InvalidAccessError exception MUST be thrown.
+    </dd>
+    <dt>optional unsigned long output = 0</dt>
+    <dd>
+      The <code>output</code> parameter is an index describing which output of the
+      <a><code>AudioNode</code></a> from which to disconnect. If this paremeter 
+      is out-of-bound, an IndexSizeError exception MUST be thrown.
+    </dd>
+    <dt>optional unsigned long input = 0</dt>
+    <dd>
+      The <code>input</code> parameter is an index describing which input of
+      the destination <a><code>AudioNode</code></a> to disconnect. If this
+      parameter is out-of-bounds, an IndexSizeError exception MUST be thrown.
+    </dd>
+  </dl>
+  </dd>
+
+  <dt>void disconnect()</dt>
+  <dd>
+  <dl class=parameters>
+    <dt>AudioParam destination</dt>
+    <dd>The <code>destination</code> parameter is the <a><code>AudioParam</code></a> 
+      to disconnect.</dd> If <code>destination</code> <code>AudioParam</code> is 
+      not connected, an InvalidAccessError exception MUST be thrown.
+    <dt>optional unsigned long output = 0</dt>
+    <dd>
+      The <code>output</code> parameter is an index describing which output of the
+      <a><code>AudioNode</code></a> from which to disconnect. If the 
+      <code>parameter</code> is out-of-bound, an IndexSizeError exception MUST 
+      be thrown.
     </dd>
   </dl>
   </dd>
--- a/index.html
+++ b/index.html
@@ -1153,12 +1153,19 @@ way that other EventTargets accept events.
 
   <dt>void disconnect()</dt>
   <dd>
+    <p>Disconnects all outgoing connections from <a><code>AudioNode</code></a>.
+    </p>
+  </dd>
+
+  <dt>void disconnect()</dt>
+  <dd>
   <dl class=parameters>
-    <dt>optional unsigned long output = 0</dt>
+    <dt>unsigned long output</dt>
     <dd>
       This parameter is an index describing which output of the
-      <a><code>AudioNode</code></a> to disconnect. If this parameter is
-      out-of-bounds, an IndexSizeError exception MUST be thrown.
+      <a><code>AudioNode</code></a> to disconnect. It disconnects all outgoing 
+      connections from the given output. If this parameter is out-of-bounds, an 
+      IndexSizeError exception MUST be thrown.
     </dd>
   </dl>
   </dd>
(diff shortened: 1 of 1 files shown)


adelespinasse:
I think the first form (the old one) is still underspecified. Does it disconnect all connections from the given output, if there's more than one? Or only one (which one)? I think it's all of them, but it doesn't say so.

hoch:
@adelespinasse I think it should disconnect all outgoing connections out of the specified output. When the output index becomes '0' when it is undefined, it simply disconnects everything from output#0. Although this is the current behavior (that why I left it as it was), I agree that we can be more specific about this. Thanks!

hoch:
Actually I have found a small syntactic concern:
~~~
splitter.connect(merger, 0, 0);
splitter.connect(merger, 1, 1);
splitter.connect(merger, 2, 2);

splitter.disconnect(merger);
~~~
What should happen? and is the behavior consistent with how this statement looks like?
Should have brought this up before I file the pull request, but hopefully we can talk about this.

rtoy:
On Mon, Feb 9, 2015 at 11:17 AM, Hongchan Choi <notifications@github.com>
wrote:

> Actually I have found a small syntactic concern:
>
> splitter.connect(merger, 0, 0);
> splitter.connect(merger, 1, 1);
> splitter.connect(merger, 2, 2);
>
> splitter.disconnect(merger);
>
>
​Just to make it clear, splitter.disconnect(merger) is (as proposed) the
same as splitter.disconnect(merger, 0, 0).

Being able to use splitter.disconnect(merger) to disconnect all outputs of
the splitter appeals to me to disconnect completely one node from all
downstream connections instead of having to disconnect each output
individually.  (Granted, only the splitter and script processor and
audioworker can have more than one output.)
​


>  What should happen? and is the behavior consistent with how this
> statement looks like?
> Should have brought this up before I file the pull request, but hopefully
> we can talk about this.
>
> —
> <https://github.com/WebAudio/web-audio-api/pull/479#issuecomment-73571045>
> .
>



-- 
Ray

hoch:
Here's one more corner case:
~~~
splitter.connect(merger, 1, 1);
splitter.connect(audioParam, 1);

splitter.disconnect(merger, 1);
~~~
When the output has mixed destination type (Node and AudioParam) - what should happen? I am pretty certain that we know what we want to happen, but I think the behavior should also be specced.

Also what should happen when we do the following?
~~~
splitter.connect(merger, 1, 1);
splitter.connect(audioParam, 1);

splitter.disconnect(merger);
~~~
And does it make sense syntactically? According to 

I am trying to push this change to the spec, but there are some cases we need to think through. Any feedback from WG would be great.

cwilso:
I have a preference for

     splitter.disconnect(merger);

disconnecting all connections (not just the first one).  Similarly, 

    splitter.disconnect(merger,0);

needs to disconnect all connections from that output.

hoch:
@cwilso Then how do we resolve the ambiguity between these two based on the current IDL? 
~~~
splitter.disconnect(merger);
splitter.disconnect(merger, 0, 0);
~~~

Maybe IDL should be changed to accommodate this?
~~~
void disconnect (AudioNode destination); // disconnect all connections to destination
void disconnect (AudioNode destination, unsigned long output = 0, optional unsigned long input = 0);
~~~

hoch:
Come to think of it, we already have this problem in the current spec.

~~~
splitter.connect(merger, 0, 0);
splitter.connect(merger, 1, 1);
splitter.disconnect();
~~~

After the disconnection, the path between splitter#1 -> merger#1 will still remain. It works as specified in the spec, but not syntactically great. Not sure if this is what we want and how it is implemented. It is only applicable to the case where Splitter and Merger are involved, but just wanted to point out it's there.

hoch:
I would love to hear what WG think - if we decide to make this method a bit more useful, the IDL should be changed and that might break the current design. Here is the one I came up with - I basically removed all the default/optional values from the argument:

~~~
// disconnect all outgoing connections from the node.
disconnect();
// disconnect all outgoing connections from the given output.
disconnect(unsigned long output);
// disconnect all connections to the destination (AudioNode || AudioParam) from all outputs.
disconnect(destination);
// disconnect all connections to the destination from the given output.
disconnect(destination, unsigned long output);
// disconnect a single connection to the destination between given input and output
disconnect(destination, unsigned long output, unsigned long input);
~~~

Splitter and merger are the only reason for this chaos. When we deal with other nodes (which have single input and output), we're fine with default zero index in the argument. I am not sure if we need this level of complexity for very small use cases.

Any thought?

rtoy:
On Tue, Feb 10, 2015 at 9:23 AM, Hongchan Choi <notifications@github.com>
wrote:

> I would love to hear what WG think - if we decide to make this method a
> bit more useful, the IDL should be changed and that might break the current
> design. Here is the one I came up with - I basically removed all the
> default/optional values from the argument:
>
> // disconnect all outgoing connections from the node.
> disconnect();
>
>
​This is an incompatible change, right?  Previously, this disconnected
output 0 and input 0.  I think the incompatibility risk is low, perhaps low
enough to ignore.​


>
> // disconnect all outgoing connections from the given output.
> disconnect(unsigned long output);
> // disconnect all connections to the destination (AudioNode || AudioParam) from all outputs.
> disconnect(destination);
> // disconnect all connections to the destination from the given output.
> disconnect(destination, unsigned long output);
> // disconnect a single connection to the destination between given input and output
> disconnect(destination, unsigned long output, unsigned long input);
>
> Splitter and merger are the only reason for this chaos. When we deal with
> other nodes (which have single input and output), we're fine with default
> zero index in the argument. I am not sure if we need this level of
> complexity for very small use cases.
>
> Any thought?
>
> —
> <https://github.com/WebAudio/web-audio-api/pull/479#issuecomment-73742236>
> .
>



-- 
Ray

cwilso:
I think if the index is unspecified, all indices (inbound and outbound)
should be disconnected.

hoch:
>  I think the incompatibility risk is low, perhaps low enough to ignore.​

I agree. That's why I brought it up again - to get WG feedback on this breaking change.

hoch:
> I think if the index is unspecified, all indices (inbound and outbound) should be disconnected.

Yes, that is what this proposal is for. So indices will not be optional or defaulted to 0. User has to specify the index to pinpoint the disconnection. Otherwise it will disconnect everything from the given output, or to the given destination.

adelespinasse:
I'm not actually sure that would be a breaking change, because I think existing implementations may be broken already.

I'm basing this on a quick experiment in Firefox, where I can observe connections in the Web Audio inspector. I'm using the beta channel Firefox (currently 36.0), so I guess there could be a partial/incorrect implementation of the new spec or something, but that doesn't seem likely... I think what I'm seeing would probably be the same as recent release versions.

I create a channel splitter with the default 6 outputs and some nodes to connect it to. I connect different outputs of the splitter to different nodes (or to different inputs of a channel merger, but the inspector doesn't show the individual connections in that case so it's a little harder to see.)

Any of the following disconnects all outputs from the splitter node, not just from one of its outputs:

splitter.disconnect()
splitter.disconnect(0)
splitter.disconnect(node2);  // one of the nodes it's connected to
splitter.disconnect(audioContext.destination);  // a node it's not connected to

Calling splitter.disconnect(1) throws an error ("TypeError: obj.toString(...).match(...) is null"), whether or not output 1 is connected.

So it seems like, in Firefox at least, disconnect is incapable of doing anything besides disconnecting all outputs of the node. So I would say it's definitely good to keep that behavior for disconnect() (with no arguments). It MIGHT be a breaking change to make disconnect(0) only disconnect output zero, although only for splitter nodes, and I don't know how likely it is that that form has been used much.

Like I said, I haven't tested other browsers, but most audio applications try to at least support Chrome and Firefox, so it seems unlikely that there's too much code out there that is completely broken in Firefox.

hoch:
> So I would say it's definitely good to keep that behavior for disconnect() (with no arguments)

The current spec says that no argument means 'index 0.' Disconnecting everything with no argument is not something we can keep (even if FF already works that way) because that is not what the spec says.

Anyhow, I guess we all agree that `disconnect()` or `disconnect(destination)` should be a shortcut of disconnecting everything out of the node, or to the destination.

I am expecting to hear more opinions!

hoch:
It is a separate issue but we have the similar issue with `connect()`:

~~~
splitter.connect(merger);
~~~

This will only connect `splitter->output#0` to `merger->input#0`.

hoch:
I made some changes in the IDL of `AudioNode.disconnect()` method. Although the compatibility risk is fairly low, it is still a behavior change in the method.

http://hoch.github.io/web-audio-api/#idl-def-AudioNode

Please take a look the preview.

cwilso:
@hoch, the latest iteration LGTM, except I think the text under destination in the second-to-last 2 forms should say:

If there is no connection to the destination FROM THE GIVEN OUTPUT, an InvalidAccessError exception...

If there is no connection to the destination FROM THE GIVEN OUTPUT TO THE GIVEN INPUT, an InvalidAccessError exception...

hoch:
@cwilso Thanks. Will fix that.

joeberkovitz:
The current IDL (and the breaking change to disconnect() with no arguments causing it to disconnect everything) looks good to me.

padenot:
Same, looks good.

hoch:
@cwilso @joeberkovitz @padenot 

Thanks for the review!

PR #485, opened 2015-02-17
Address issue 477, making the max array size 2048, not 4096.

rtoy:
The maximum size of the arrays for a PeriodicWave is 2048, not 4096.  All current implementations ignore anything above 2048 anyway.


cwilso:
LGTM.

PR #490, opened 2015-03-06
Use HTTPS URLs where possible

mathiasbynens:
This also fixes the mixed content issues on https://webaudio.github.io/web-audio-api/ due to the MathJax script being loaded over HTTP.

Preview: https://mathiasbynens.github.io/web-audio-api/


padenot:
Looks good, thanks! This also seem to include a bunch of trailing white space fixes, but we might as well take it.

Issue #459, opened 2014-12-16
ChannelMergerNode in a cycle has unspecified behaviour


mark-buer:
What happens if the output of a ChannelMergerNode is fed back into it's input (via a DelayNode)?
The specification does not state the expected runtime behaviour.

As an illustration, please consider the following JS snippet where a ChannelMergerNode has two connected inputs - an oscillator, and the ChannelMergerNode itself (via a DelayNode):
```javascript
var audioContext;
if (!audioContext) {
    if (window.AudioContext) {
        audioContext = new AudioContext();
    } else if (window.webkitAudioContext) {
        audioContext = new webkitAudioContext();  
    } else {
        throw new Error('AudioContext missing');
    }
}

var o = audioContext.createOscillator();
var m = audioContext.createChannelMerger(2);
var d = audioContext.createDelay();

o.connect(m, 0, 0);
d.connect(m, 0, 1);
m.connect(d);
m.connect(audioContext.destination);
d.delayTime.value = 0.5;
o.start(0);
```

The snippet produces silence in Blink and seems to correctly produce a tone in Mozilla.

joeberkovitz:
Fixed by changing behavior of ChannelMergerNode #304

Issue #304, opened 2014-03-31
Difficult to address specific channel in ChannelMergerNode


tszynalski:
Suppose you want to write a function that plays a mono sample on a given channel (left, right, center, etc.). The intuitive way of doing it would be to connect a BufferSourceNode with the mono sample to the appropriate input of a ChannelMergerNode. For example:
```
//connect to input #5 = right surround
bufferSource.connect(mergerNode, 0, 5);
```
Unfortunately, a ChannelMergerNode only takes into account **connected inputs**, so the above code would ignore the unconnected inputs 0-4, take the first connected input (#5), and output a mono signal.

The current solution is to do the following:
```
var silence = audioContext.createBufferSource();
//creating a dummy AudioBufferSourceNode which is treated as one channel of silence
silence.connect(mergerNode, 0, 0);
silence.connect(mergerNode, 0, 1);
silence.connect(mergerNode, 0, 2);
silence.connect(mergerNode, 0, 3);
silence.connect(mergerNode, 0, 4);
bufferSource.connect(mergerNode, 0, 5);
```
Note that when working with 5.1-channel audio, we have to connect **5 dummy silent nodes** every time we want to play a sample on a single channel.

In addition, I believe there is nothing in the spec that prevents a browser from disconnecting and disposing of the dummy nodes as soon as the function exits and the silence variable is no longer referenced. This would cause the mergerNode to start outputting a mono signal, since it would receive only one input. So we would also have to make sure the "silence nodes" exist in a different (longer-lasting) scope than bufferSource.

It would be better to have the possibility of 1-1 mapping between ChannelMergerNode inputs and the channels in the output signal. We could tell a ChannelMergerNode what kind of output signal we want (mono, stereo, 5.1, etc.), connect a single input and the remaining inputs would be filled with silence.
It could work like this:

```
mergerNode.outputSignal = "5.1";
bufferSource.connect(mergerNode, 0, 5); //connect source to right surround channel
```
or
```
mergerNode.outputSignal = "5.1";
bufferSource.connect(mergerNode, 0, "SR");
```

The default setting of the "outputSignal" property could be "auto". This would correspond to the current way it works – automatically determine the number of channels by summing the channels on all connected inputs.
Setting outputSignal to "5.1" would have the following effects:
- take the first 6 inputs (# 0-5)
- interpret each input as mono (downmix if necessary)
- interpret unconnected inputs as one channel of silence
[...]

PR #491, opened 2015-03-06
Changed behavior of ChannelMergerNode

hoch:
As we discussed on a WG teleconf 3/5/2015, I wrote up the spec change based on my proposal in the discussion. FYI, this spec change is to address issues:
- #304 
- #459 

You can preview the PR here: 
http://hoch.github.io/web-audio-api/#h3_the-channelmergernode-interface


hoch:
Thanks!

PR #492, opened 2015-03-09
Added missing names to the WG memeber list

hoch:
This PR is to reflect recent changes in WG member names and roles in the spec.


hoch:
Thanks!

Issue #460, opened 2014-12-16
currentTime in AudioContext is not progressing while suspended anymore


obeletski:
I noticed that some updates to Web Audio API specification regarding currentTime in suspend state should be done:

**currentTime attribute of AudioContext**
> currentTime of type double, readonly

> This is a time in seconds which starts at zero when the context is created and increases in real-time. All scheduled times are relative to it. This is not a "transport" time which can be started, paused, and re-positioned. It is always moving forward. A GarageBand-like timeline transport system can be very easily built on top of this (in JavaScript). This time corresponds to an ever-increasing hardware timestamp

The latest draft of the specification actually enables us to pause and resume timeline of AudioContext and the transport timeline that is behind it. It would be good to reflect that in the description of the currentTime attribute.

**suspend/resume methods of AudioContext**
> Suspends(resumes) the progression of time in the audio context

Would be good to use the same style of referencing to the timeline of AudioContext as in the other parts of the spec - "same time coordinate system as AudioContext's currentTime" to avoid potential confusion with the "transport" time.

maximg:
Note that the description of currentTime explicitly states that it increases in real-time, without mentioning the suspended state. This is confusing at least and seems to contradict the description of the suspended state. It will help if the spec is updated to mention that currentTime only increases when the context is  running.

joeberkovitz:
#500 is fix for this

billhofmann:
Can this be closed as fixed by #500?

joeberkovitz:
I believe so although #500 is still not merged yet.

mdjp:
#500 now merged, closing.

PR #500, opened 2015-03-24
Clarify that currentTime does not increase when the context is not "running"

padenot:
This fixes issue #460


padenot:
@rtoy, this is rebased and includes fixes for your comments (https://github.com/WebAudio/web-audio-api/commit/eb0b41455dc3d01b14f8ab72ac08d229d6a0be33)

rtoy:
lgtm

PR #501, opened 2015-03-24
Use rejected promise instead of exception for suspend/resume/close.

### LINKED ISSUE #441: suspend/resume/close should use rejected promise for offline contexts
Currently the spec says that suspend/resume/close on an offline context should throw an exception.  However, since these methods also return a promise, it seems more appropriate not to throw an exception but return a promise that is rejected with an InvalidStateError DOM exception.

padenot:
This also specs what exception should be thrown when suspend and resume are
called on a close context.

This fixes #441.


rtoy:
Looks good, but I guess it needs to be rebased.

padenot:
This will be better fixed by merging in the rewrite of suspend/resume/close that sits in my fork on top of the Processing Model changes. Closing ths.

PR #514, opened 2015-04-27
441 invalid access error for offline suspend/resume/close

rtoy:
Use InvalidAccessError instead of InvalidStateError when using suspend/resume/close for an offline context.


padenot:
Hrm, why? We are clearly calling the method in an invalid state here, aren't we ?

hoch:
These changes are not valid anymore. Now we can call suspend/resume/close on OfflineAudioContext. They should return a rejected promise with DOM exception if something goes wrong. In that case, I believe the DOM exception in the rejection should be `invalidStateError`.

rtoy:
Closing this pull request since it's no longer valid.

Issue #447, opened 2014-11-25
Define built-in oscillator output more carefully


hoch:
This might be an issue for implementors, but I think it is worth to raise the flag to hear WG's opinion.

I was fiddling with a JS test platform for Web Audio API and found that the sound produced by oscillator nodes between browser actually has difference in loudness (RMS level) - and in some cases it is more than audible.

I understand we cannot have sample-accurate consistency across the browsers, but I think we need some guideline for loudness level and have to build a common test platform around it. (which we already discussed)

~~~js
var mod = context.createOscillator();
var idx = context.createGain();
var car = context.createOscillator();

mod.connect(idx);
idx.connect(car.frequency);
car.connect(context.destination);

mod.frequency.value = 1000;
idx.gain.value = 220;
car.type = 'sawtooth';
car.frequency.value = 440;

mod.start(0.0);
car.start(0.0);
~~~

The code above produces (RMS to dBFS):
Chrome/Safari: -2.14dB
FireFox: -1.84dB

Although it is just 0.3dB difference, but if you change frequency, modulation index or waveform type, the difference varies. I think we don't want to see this type of variety across the browser.

mdjp:
Would be exposed by proposed independent test suite as proposed by the WG

joeberkovitz:
Resolution: The output of a built-in oscillator must be equivalent to the output of a PeriodicWave whose Fourier coefficients yield the built-in wave type, with normalization as per the resolution of #91.

karlt:
On Mon, 01 Jun 2015 07:48:00 -0700, Joe Berkovitz wrote:

> Resolution: The output of a built-in oscillator must be equivalent to the
> output of a PeriodicWave whose Fourier coefficients yield the built-in wave
> type, with normalization as per the resolution of #91.

I'm not clear exactly what that means.

A PeriodicWave with an infinite number of coefficients would be
required to generate the basic waveform.  Is that what is meant
here?

Normalization would have no effect.  Band-limiting would be
applied to the specified infinite-resolution waveform which
already has range ±1, and so the discrete waveform would have a
slightly different range, as expected.  That all makes sense to
me.

It wouldn't make sense to normalize based on an inverse FFT of a
Fourier series truncated to the maximum size of arrays supported
by PeriodicWave.  That would mean a change the magnitude if the
maximum size of arrays is changed, in
https://github.com/WebAudio/web-audio-api/issues/386 for example.

rtoy:
[...]

PR #545, opened 2015-06-15
Add optional normalization parameter to createPeriodicWave

### LINKED ISSUE #91: (WaveTableNormalization): WaveTable normalization in createWaveTable not clearly defined
> Originally reported on W3C Bugzilla [ISSUE-17370](https://www.w3.org/Bugs/Public/show_bug.cgi?id=17370) Tue, 05 Jun 2012 11:55:13 GMT
> Assigned to 

Audio-ISSUE-84 (WaveTableNormalization): WaveTable normalization in createWaveTable not clearly defined [Web Audio API]

http://www.w3.org/2011/audio/track/issues/84

Raised by: Philip Jägenstedt
On product: Web Audio API

https://dvcs.w3.org/hg/audio/raw-file/tip/webaudio/specification.html#dfn-createWaveTable

"The created WaveTable will be used with an Oscillator and will represent a normalized time-domain waveform having maximum absolute peak value of 1. Another way of saying this is that the generated waveform of an Oscillator will have maximum peak value at 0dBFS. Conveniently, this corresponds to the full-range of the signal values used by the Web Audio API. Because the WaveTable will be normalized on creation, the real and imag parameters represent relative values."

This does not clearly define what normalization must be performed. Should/can the normalization be performed analytically, or should one period be generated and the maximum (absolute) value in that period be found? These approaches may not arrive at the exact same normalization level, so a sample-exact test suite would not be possible.

rtoy:
Fix #91 

This adds the optional normalization parameter to createPeriodicWave, defines how normalization is done and how the time-domain waveform can be created from the Fourier coefficients.

### DIFF
--- a/index.html
+++ b/index.html
@@ -835,7 +835,7 @@ or <a href="#widl-AudioContext-decodeAudioData-Promise-AudioBuffer--ArrayBuffer-
       Conveniently, this corresponds to the full-range of the signal values used by
       the Web Audio API.  Because the PeriodicWave is normalized by default on creation,
       the <code>real</code> and <code>imag</code> parameters represent
-      <em>relative</em> values. If normalization is disabled via the <code>normalize</code>
+      <em>relative</em> values. If normalization is disabled via the <code>disableNormalization</code>
       parameter, this normalization is disabled, and the time-domain waveform has the amplitudes as
       given by the Fourier coefficients.
     <p>
--- a/index.html
+++ b/index.html
@@ -828,14 +828,16 @@ or <a href="#widl-AudioContext-decodeAudioData-Promise-AudioBuffer--ArrayBuffer-
       <a href="https://en.wikipedia.org/wiki/Fourier_series">Fourier series</a>
       representing the partials of a periodic waveform.  The created
       <a><code>PeriodicWave</code></a> will be used with an
-      <a><code>OscillatorNode</code></a> and will represent a <em>normalized</em>
+      <a><code>OscillatorNode</code></a> and, by default,  will represent a <em>normalized</em>
       time-domain waveform having maximum absolute peak value of 1.  Another way of
       saying this is that the generated waveform of an
       <a><code>OscillatorNode</code></a> will have maximum peak value at 0dBFS.
       Conveniently, this corresponds to the full-range of the signal values used by
-      the Web Audio API.  Because the PeriodicWave will be normalized on creation,
+      the Web Audio API.  Because the PeriodicWave is normalized by default on creation,
       the <code>real</code> and <code>imag</code> parameters represent
-      <em>relative</em> values.
+      <em>relative</em> values. If normalization is disabled via the <code>normalize</code>
+      parameter, this normalization is disabled, and the time-domain waveform has the amplitudes as
+      given by the Fourier coefficients.
     <p>
       As <a>PeriodicWave</a> objects maintain their own representation, any
       modification of the arrays uses as the <code>real</code> and
@@ -864,6 +866,12 @@ or <a href="#widl-AudioContext-decodeAudioData-Promise-AudioBuffer--ArrayBuffer-
         not exist in the Fourier series.  The second element (index 1) represents
         the fundamental frequency.  The third element represents the first
         overtone, and so on.
+      </dd>
+      <dt>optional boolean normalize</dt>
+      <dd>
+        The <dfn id="dfn-normalize">normalize</dfn> parameter indicates whether the periodic wave
+        should be normalized or not.  By default the periodic wave is normalized.  If the parameter
+        is not given, the periodic wave is normalized.
       </dd>
     </dl>
     </dd>
@@ -4468,7 +4476,7 @@ to determine a <em>computedFrequency</em> value:
 </dl>
 
 <section>
-<h2>Basic waveform phase</h2>
+<h2>Basic Waveform Phase</h2>
 <p>
   The idealized mathematical waveforms for the various oscillator types are defined here. In summary,
   all waveforms are defined mathematically to be an odd function with a positive slope at time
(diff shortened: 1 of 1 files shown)


domenic:
This seems like it exhibits two antipatterns:

- [The boolean trap](http://ariya.ofilabs.com/2011/08/hall-of-api-shame-boolean-trap.html): context-less boolean arguments
- Incorrect defaults: defaults for booleans should always be false.

May I suggest `createPeriodicWave(..., { denormalized: true })` instead of `createPeriodicWave(..., false)`?

hoch:
That's a good point. However, we might have to consider the different name than `dernomaized`. What we are trying to do here is to avoid the internal normalization. I think what denormalizing means does not quite fit here.

Perhaps we should do `createPeriodicWave(..., { normalize: true })` and change the default behavior of the method since the internal normalization was known to be problematic in some cases. Some people will not like it though.

domenic:
Yeah, I thought denormalized might not be right. And yeah, it would be cleanest to change the default. Otherwise you have to come up with an inverted name, which can be awkward for words without natural antonyms.

rtoy:
I am opposed to changing the default normalization.

domenic:
Right, I didn't meant to imply you should actually make a backward-incompatible change, just that that would be "easiest" from a naming perspective. Coming up with an inverted name seems most straightforward.

rtoy:
Ok.  I don't have any strong opinions either way on the other suggestions. I'd even be ok with createUnnormalizedPeriodicWave(r, i).

foolip:
> Incorrect defaults: defaults for booleans should always be false.

I like it when this is the case, but does it matter if the default is true in cases where the inverted name would be awkward?

domenic:
> I like it when this is the case, but does it matter if the default is true in cases where the inverted name would be awkward?

Yes, I think it's better to have a slightly-awkward name than to violate the expectation that a boolean defaults to false.

foolip:
FWIW, it wouldn't be the first dictionary member to default to true, and the existing cases in canvas don't seem terrible to me:
https://html.spec.whatwg.org/#canvasrenderingcontext2dsettings
https://www.khronos.org/registry/webgl/specs/latest/1.0/#5.2

Of course, it would depend on what the actual suggestion is, and if there are actual usability problems with boolean dictionary members defaulting to true, not just a style issue.

rtoy:
If we write the Fourier series in complex form, I think this basically works out to be the same as solving for the roots of an order N polynomial with complex coefficients.  Doing the inverse transform and finding the max is much easier to do and doesn't require finding and writing a good polynomial root solver, especially since the order of the polynomial could be as high as 2048 or more.

karlt:
If there is no reason why a PeriodicWave must be used with a
particular AudioContext, then would it make sense to instead
provide a PeriodicWave constructor?

The new constructor could then have the desired
default-no-normalization behaviour, and the createPeriodicWave()
factory method could be retained as long as legacy support is
required.

Or is a factory method preferred to reduce pollution of the global
namespace?

hoch:
I can second the idea of having PeriodicWave detached from the context, but:
```js
var p1 = new PeriodicWave({ normalization: false });
var p2 = context.createPeriodicWave({ normalization: true });
```
This doesn't look great in my opinion. Probably we really need to try hard to come up with a clever name.

rtoy:
How about we just name the parameter disableNormalization?

hoch:
+1

padenot:
The key of the dict, you mean ? Or just the parameter in the prototype ?

rtoy:
Just the parameter.  I don't foresee other items we would want to add to the dict, so a dict seems like overkill.

domenic:
No. Use a dict. The Boolean trap is real.

rtoy:
Ugh. Ok.

karlt:
Dealing with the normalization problem on PeriodicWave is good if
we can assume that normalization of built-in oscillator types has
no effect because they are already normalized.

However, it wouldn't solve the problem with built-in oscillator
types if they are going to be based on a similar normalization
using a magnitude from an arbitrary Fourier truncation of the
specified waveform.  If that is the intention, then a solution is
required on OscillatorNode instead of PeriodicWave.
https://github.com/WebAudio/web-audio-api/issues/447

Normalization of built-in waves is not required because they are
already normalized, so I hope we don't need to solve that problem.

rtoy:
I don't quite follow.

But in any case, the waveforms for oscillators are always normalized, and to be consistent with PeriodicWaves, the normalization must be consistent with the normalization of PeriodicWaves. We are not allowing oscillators the option of not being normalized.  If a user wants that, they can use a periodic wave without normalization.

rtoy:
Please take another look. I've updated the spec to use a dictionary.

index.html
```diff
@@ -828,14 +828,16 @@ <h2 id="AudioContext">The AudioContext Interface</h2>
...
+      <a><code>OscillatorNode</code></a> and, by default,  will represent a <em>normalized</em>
       time-domain waveform having maximum absolute peak value of 1.  Another way of
       saying this is that the generated waveform of an
       <a><code>OscillatorNode</code></a> will have maximum peak value at 0dBFS.
       Conveniently, this corresponds to the full-range of the signal values used by
-      the Web Audio API.  Because the PeriodicWave will be normalized on creation,
+      the Web Audio API.  Because the PeriodicWave is normalized by default on creation,
       the <code>real</code> and <code>imag</code> parameters represent
-      <em>relative</em> values.
+      <em>relative</em> values. If normalization is disabled via the <code>normalize</code>
```
domenic: s/normalize/disableNormalization?

domenic:
LGTM, thank you so much for taking the time to make this API ergonomic :)

padenot:
Yep, looks good.

hoch:
@domenic Is there any method in the other Web APIs that uses a JavaScript object (dict) as an optional descriptor? What if `new Float32Array(1)` is passed as an argument? For example:

~~~
context.createPeriodicWave(new Float32Array(5), new Float32Array(5), new Float32Array(5));
~~~

How should we handle the third argument?

domenic:
Yes,  I suggest searching for the keyword "dictionary" in your favorite spec (e.g. DOM or HTML) and you will find plenty of examples.

hoch:
@domenic Thanks! It helped.

karlt:
> However, it wouldn't solve the problem with built-in oscillator types if they are going to be based on a similar normalization using a magnitude from an arbitrary Fourier truncation of the specified waveform. If that is the intention, then a solution is required on OscillatorNode instead of PeriodicWave.

https://github.com/WebAudio/web-audio-api/issues/2046 is reconsidering this.

PR #551, opened 2015-06-17
Define FFT for AnalyserNode

### LINKED ISSUE #470: Define FFT for AnalyzerNode
The FFT is not specified for the AnalyzerNode.  There is at least the mathmatician's definition and the engineer's definition which are slightly different.  And the FFT has at least 3 different possible scalings (1, 1/N, 1/sqrt(N)), so that means at least 6 different possible implementations.

I believe all implementations use the engineering definition with a scale factor of 1:

X(k) = sum(x(n)*exp(-j*2*pi*k*n/N), n = 0, N - 1)

rtoy:
Fix #470.

The section for the Blackman window is also updated to use
mathematical notation.

### DIFF
--- a/index.html
+++ b/index.html
@@ -3578,27 +3578,42 @@ When the <dfn id="current-frequency-data">current frequency data</dfn> are
 computed, the following operations are to be performed:
 <ol>
   <li> <a href="#blackman-window">Apply a Blackman window</a> to the time domain input data </li>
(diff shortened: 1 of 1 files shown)


rtoy:
Hope you don't mind @padenot, but I've added a definition for the FFT in the AnalyserNode

padenot:
No that's cool. This makes sense to express mathematically (unlike some other things in the spec where it makes sense to keep the algorithm in code).

Issue #509, opened 2015-04-08
AudioParam rates for BiquadFilterNode are inconsistant


pendragon-andyh:
Section 2.23 of the API specification contains the following sentences:
* Both frequency and detune are **a-rate** parameters and are used together to determine ...
* **All** attributes of the BiquadFilterNode are **k-rate** AudioParam.

padenot:
We can just remove then mention of a-rate, I think.

@rtoy, @jipodine do you think there is a use case to have a-rate filters ?

rtoy:
I don't see a reason for a-rate filters, but perhaps I lack imagination.

On Fri, Apr 17, 2015 at 8:48 AM, Paul Adenot <notifications@github.com>
wrote:

> We can just remove then mention of a-rate, I think.
>
> @rtoy <https://github.com/rtoy>, @jipodine <https://github.com/jipodine>
> do you think there is a use case to have a-rate filters ?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/WebAudio/web-audio-api/issues/509#issuecomment-94013359>
> .
>



-- 
Ray

luketeaford:
Audio rate filters are extremely important for synthesis. It's a well known technique for creating rough timbres. Even a very subtle amount can add a pleasing depth to the sound.

hoch:
@luketeaford Can you provide an example that really makes difference? Although it is specified k-rate, the filter commonly has internal de-zippering on parameters (which operate on top of coefficients) and the output signal usually smoothed out by the nature of memory inside.

Is this about manipulating actual coefficients sample-by-sample basis? If that's the case, it is hard for me to see its usefulness.

luketeaford:
I'm not sure because that is beyond my mathematical understanding. I am doing (to the best of my knowledge anyway) arate filter modulation on my web audio synth and I think the way it sounds is correct.

rtoy:
If it's web audio and if it's using the biquad filter, then it can't be
a-rate. I'm pretty sure every browser does only k-rate.

On Fri, Apr 17, 2015 at 11:29 AM, Luke Teaford <notifications@github.com>
wrote:

> I'm not sure because that is beyond my mathematical understanding. I am
> doing (to the best of my knowledge anyway) arate filter modulation on my
> web audio synth and I think the way it sounds is correct.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/WebAudio/web-audio-api/issues/509#issuecomment-94046438>
[...]

Issue #244, opened 2013-09-30
Add an way to request mathematically correct oscillators


padenot:
When using an OscillatorNode as an LFO, it can be useful to get rid of the ripples inherent to high frequency limiting to get precise modulation.

This issue is about designing an API to request mathematically correct oscillator, and define them using a mathematical definition (in terms of phase, waveform shape and peak amplitude).

mdjp:
This is a niche use case which can be covered by the Audio Worklet. Closing.

PR #552, opened 2015-06-17
Fix #447: Define oscillator carefully.

### LINKED ISSUE #447: Define built-in oscillator output more carefully

rtoy:
Define the oscillators for the builtin types to produce the same
results as if a PeriodicWave with the appropriate Fourier coefficients
were used to create the oscillator.


rtoy:
As noted in https://github.com/WebAudio/web-audio-api/issues/509#issuecomment-122955065, we mention allowing implementations to do less costly algorithms.  However, since we're defining the OscillatorNode in terms of the PeriodicWave, perhaps we also need to specify how many coefficients are used.  Currently, I think all implementations use the (previous) max of 2048.  Since we're also allowing PeriodicWave to have at least 8192 coefficients, the number of coefficients being used, especially at higher sample rates is unspecified.  This can make a difference

karlt:
rtoy writes:

> Since we're also allowing
> PeriodicWave to have at least 8192 coefficients, the number of coefficients
> being used, especially at higher sample rates is unspecified.  This can make a
> difference

Yes, best to not apply any additional normalization at all, which
is equivalent to applying the max(abs()) normalization like Webkit
used in PeriodicWave but using the limit as the number of
coefficients approaches infinity.

i.e. use the equivalent of the normalization of the best possible
implementation by simply not applying any normalization at all.

Anything else is making it more complicated than necessary.

rtoy:
On Mon, Jul 20, 2015 at 7:31 PM, Karl Tomlinson <notifications@github.com>
wrote:

> rtoy writes:
>
> > Since we're also allowing
> > PeriodicWave to have at least 8192 coefficients, the number of
> coefficients
> > being used, especially at higher sample rates is unspecified. This can
> make a
> > difference
>
> Yes, best to not apply any additional normalization at all, which
> is equivalent to applying the max(abs()) normalization like Webkit
> used in PeriodicWave but using the limit as the number of
> coefficients approaches infinity.
>
> i.e. use the equivalent of the normalization of the best possible
> implementation by simply not applying any normalization at all.
>

​Since we cannot do the infinite sum and since the waveform is supposed to
be band-limited, there will always be ringing causing the waveform to
exceed 1 in amplitude.

Your approach means oscillators will always exceed the nominal audio
amplitude.  I don't think this is what we want.

But at least if you really want such oscillators you can do that now by
creating your own PeriodicWave with the right coefficients with
normalization disabled.

And for the record, Chrome will use 8192 coefficients for sample rates
above 88.2 kHz, 2048 for rates between 24 kHz and 88.2 kHz, and 1024 for
rates at or below 24 kHz.  Yes, these limits are somewhat arbitrary, but
the 2048 limits preserves backward compatibility for the typical 44.1 and
48 kHz sample rates.​


>
> Anything else is making it more complicated than necessary.
>
> —
> <https://github.com/WebAudio/web-audio-api/pull/552#issuecomment-123131862>
> .
>



-- 
Ray

karlt:
rtoy writes:

> Your approach means oscillators will always exceed the nominal audio
> amplitude.

Yes.  I don't see that as a problem, but band-limited square waves
already do that because the normalization happens using a
magnitude from a different truncation to the generator.

https://bugzilla.mozilla.org/show_bug.cgi?id=1102949#c2

(Changing the normalization wrt the base frequency would make
things even less predictable.)

Even if oscillator peaks didn't exceed the nominal amplitude,
clients still would need to reduce the magnitude of oscillator
outputs if they intend to mix with any other sound and send to
hardware.

> But at least if you really want such oscillators you can do that now by
> creating your own PeriodicWave with the right coefficients with
> normalization disabled.

The basic oscillator types are currently spec'd to match what I
want.  The proposal here is to change that.

> And for the record, Chrome will use 8192 coefficients for sample rates
> above 88.2 kHz, 2048 for rates between 24 kHz and 88.2 kHz, and 1024 for
> rates at or below 24 kHz.  Yes, these limits are somewhat arbitrary, but
> the 2048 limits preserves backward compatibility for the typical 44.1 and
> 48 kHz sample rates.​

Having a predictable magnitude of the frequency components would be useful.

rtoy:
On Tue, Jul 21, 2015 at 5:08 PM, Karl Tomlinson <notifications@github.com>
wrote:

> rtoy writes:
>
> > Your approach means oscillators will always exceed the nominal audio
> > amplitude.
>
> Yes. I don't see that as a problem, but band-limited square waves
> already do that because the normalization happens using a
> magnitude from a different truncation to the generator.
>

​Do you have an example of where the band-limited square waves exceed 1 in
amplitude?  They shouldn't.

I don't understand what you mean by "using a magnitude from a different
truncation to the generator".
​


>
> https://bugzilla.mozilla.org/show_bug.cgi?id=1102949#c2
>
> (Changing the normalization wrt the base frequency would make
> things even less predictable.)
>
> Even if oscillator peaks didn't exceed the nominal amplitude,
> clients still would need to reduce the magnitude of oscillator
> outputs if they intend to mix with any other sound and send to
> hardware.
>

​Yes, but that's true for any kind of mixing and having oscillators not
exceed 1 makes the mixing easier for the causal user.  Without
normalization, changing the oscillator type would cause a possibly
different gain to be used when mixing.
​


>
> > But at least if you really want such oscillators you can do that now by
> > creating your own PeriodicWave with the right coefficients with
> > normalization disabled.
>
> The basic oscillator types are currently spec'd to match what I
> want. The proposal here is to change that.
>

​You want the mathematical oscillators?  Isn't that
https://github.com/WebAudio/web-audio-api/issues/244, which we already
decided to move to v.next?

And with AudioWorkers, users can implement their own oscillators as desired.
​


>
> > And for the record, Chrome will use 8192 coefficients for sample rates
> > above 88.2 kHz, 2048 for rates between 24 kHz and 88.2 kHz, and 1024 for
> > rates at or below 24 kHz. Yes, these limits are somewhat arbitrary, but
> > the 2048 limits preserves backward compatibility for the typical 44.1 and
> > 48 kHz sample rates.​
>
> Having a predictable magnitude of the frequency components would be useful.
>

​Sorry, I don't understand.  If normalization is disabled, you know exactly
what the magnitude of the frequency components are.  The only issue that
isn't specified is how many coefficients are used to generate the time
domain waveform.  Perhaps we should.​


> —
> <https://github.com/WebAudio/web-audio-api/pull/552#issuecomment-123513628>
> .
>



-- 
Ray

karlt:
rtoy writes:

> ​Do you have an example of where the band-limited square waves exceed 1 in
> amplitude?  They shouldn't.

http://people.mozilla.org/~karlt/square-peaks.html
(They should.)

> I don't understand what you mean by "using a magnitude from a different
> truncation to the generator".

Normalization is independent of frequency and happens before the
band-limiting that depends on fundamental and Nyquist frequencies.
The band-limited waveforms are different from the waveforms used
to calculate the normalization factor, and so have different peaks.

> ​Yes, but that's true for any kind of mixing and having oscillators not
> exceed 1 makes the mixing easier for the causal user.  Without
> normalization, changing the oscillator type would cause a possibly
> different gain to be used when mixing.

There are different energy levels in the different oscillator
types.  I don't think we can argue that peak-based-normalization
simplifies mixing of different types.

> ​You want the mathematical oscillators?  Isn't that
> https://github.com/WebAudio/web-audio-api/issues/244, which we already
> decided to move to v.next?

244 was attempting to remove ripples from the band-limiting AIUI.
That is not what I'm raising here.

Band-limiting of the built-in oscillators doesn't change the
magnitude of the frequency components well below Nyquist.
I'm concerned that the proposal here might be to change the
magnitude of those frequency components through an arbitrary
normalization algorithm.

> The only issue that
> isn't specified is how many coefficients are used to generate the time
> domain waveform.  Perhaps we should.​

That is a key issue because the number of coefficients that are
selected from the infinite series affects the magnitude that is
used for normalization.

Instead of selecting an arbitrary truncation, producing arbitrary
magnitudes, I would use the full series.  i.e. normalization is
performed on the continuous (infinite resolution) waveform.

rtoy:
On Thu, Jul 23, 2015 at 3:59 PM, Karl Tomlinson <notifications@github.com>
wrote:

> rtoy writes:
>
> > ​Do you have an example of where the band-limited square waves exceed 1
> in
> > amplitude? They shouldn't.
>
> http://people.mozilla.org/~karlt/square-peaks.html
> (They should.)
>

​Thanks for the demo.  You are right.  And in retrospect, normalizing the
amplitude of the fundamental waveform is not obviously guaranteeing that
the other band-limited waveforms are also peak-limited to 1.​
​

>
> > I don't understand what you mean by "using a magnitude from a different
> > truncation to the generator".
>
> Normalization is independent of frequency and happens before the
> band-limiting that depends on fundamental and Nyquist frequencies.
> The band-limited waveforms are different from the waveforms used
> to calculate the normalization factor, and so have different peaks.
>
> > ​Yes, but that's true for any kind of mixing and having oscillators not
> > exceed 1 makes the mixing easier for the causal user. Without
> > normalization, changing the oscillator type would cause a possibly
> > different gain to be used when mixing.
>
> There are different energy levels in the different oscillator
> types. I don't think we can argue that peak-based-normalization
> simplifies mixing of different types.
>

​Yes, certainly the loudness will be different, even without normalization.
(I think).
​


>
> > ​You want the mathematical oscillators? Isn't that
> > https://github.com/WebAudio/web-audio-api/issues/244, which we already
> > decided to move to v.next?
>
> 244 was attempting to remove ripples from the band-limiting AIUI.
> That is not what I'm raising here.
>

​That's not my understanding of the issue 244, but not relevant here.
​


>
> Band-limiting of the built-in oscillators doesn't change the
> magnitude of the frequency components well below Nyquist.
> I'm concerned that the proposal here might be to change the
> magnitude of those frequency components through an arbitrary
> normalization algorithm.
>

​At least in Chrome, the band-limiting doesn't change the Fourier
coefficients in any way.  We just drop the higher coefficients from the
series; all others are unchanged.

>
> > The only issue that
> > isn't specified is how many coefficients are used to generate the time
> > domain waveform. Perhaps we should.​
>
> That is a key issue because the number of coefficients that are
> selected from the infinite series affects the magnitude that is
> used for normalization.
>
> Instead of selecting an arbitrary truncation, producing arbitrary
> magnitudes, I would use the full series. i.e. normalization is
> performed on the continuous (infinite resolution) waveform.
>

​What kind of normalization is that?  If you did (mathmatically) the
infinite sum, you get exactly the mathematical waveforms described in the
spec, all of which have amplitude 1.  Thus, no normalization.  And then
band-limiting will produce ringing that causes the output to exceed 1.

I think I'm just too stupid to grasp what you really want.
​


> —
> <https://github.com/WebAudio/web-audio-api/pull/552#issuecomment-124261330>
> .
>



-- 
Ray

karlt:
On Fri, 24 Jul 2015 10:06:13 -0700, rtoy wrote:

>> Instead of selecting an arbitrary truncation, producing arbitrary
>> magnitudes, I would use the full series. i.e. normalization is
>> performed on the continuous (infinite resolution) waveform.
>>
>
> ​What kind of normalization is that?  If you did (mathmatically) the
> infinite sum, you get exactly the mathematical waveforms described in the
> spec, all of which have amplitude 1.  Thus, no normalization.

Yes, the waveforms described in the spec are already normalized
and so further normalization should have no effect, and be
equivalent to no normalization.

That is how I think the spec should stay, because that is simple,
clear, and unaffected by arbitrary approximation.

It's not clear how to apply the normalization approach of
PeriodicWave for waveforms with an infinite number of
coefficients.  If built-in waveforms are going to be specified in
terms of PeriodicWave, then I think it is better to specify that
it is equivalent to a PeriodicWave with no normalization so that
this aspect is clear.

(Beyond this, I'm also not sure about the idea of specifying
built-in waveforms in terms of PeriodicWave because I suspect it
is not the most efficient implementation for quality
representation of these waveforms at low fundamental frequencies,
but perhaps better generators can still generate equivalent
waveforms more efficiently.  I don't think the maximum number of
Fourier coefficients should be specified to resolve performance
problems with PeriodicWave, because I suspect other algorithms do
not have this problem.)

> And then
> band-limiting will produce ringing that causes the output to exceed 1.

Yes, the output will exceed 1.

I don't see that as a problem because oscillators should not be
connected to speakers at full volume.

If that really were a problem, then we would need a normalization
that provides peaks with abs <= 1 for all fundamental frequencies.
Such normalization would be different from that of PeriodicWave.

If backward compatibility with the existing implementation is the
goal, then for clarity I would suggest specifying a factor/divisor
that is independent of sample rate.  Clients can then know the
distance between flats of a square wave or slope of a sawtooth.

For square and sawtooth I would suggest calculating that
mathematically from the peak to peak distance of a band-limited
step function.  I expect that would provide the limit of the peaks
of square and sawtooth waves as the number of Fourier coefficients
approaches infinity.  (The coefficients approach zero at the same
rate as the slopes of the sine terms approach infinity, so the
limit of the peak value of the discrete wave as the number of
coefficients approaches infinity is not the same as the peak value
of the continuous wave.)

For triangle and sine, the factor can be 1 and still be
consistent enough with the existing implementation.

rtoy:
On Sun, Jul 26, 2015 at 9:04 PM, Karl Tomlinson <notifications@github.com>
wrote:

> On Fri, 24 Jul 2015 10:06:13 -0700, rtoy wrote:
>
> >> Instead of selecting an arbitrary truncation, producing arbitrary
> >> magnitudes, I would use the full series. i.e. normalization is
> >> performed on the continuous (infinite resolution) waveform.
> >>
> >
> > ​What kind of normalization is that? If you did (mathmatically) the
> > infinite sum, you get exactly the mathematical waveforms described in the
> > spec, all of which have amplitude 1. Thus, no normalization.
>
> Yes, the waveforms described in the spec are already normalized
> and so further normalization should have no effect, and be
> equivalent to no normalization.
>
> That is how I think the spec should stay, because that is simple,
> clear, and unaffected by arbitrary approximation.
>

​The spec has never (AFAIK) actually said that.  It's always said that
these are the basic waverforms and it is up to the implementation to make
sure these are band-limited.
​


>
> It's not clear how to apply the normalization approach of
> PeriodicWave for waveforms with an infinite number of
> coefficients. If built-in waveforms are going to be specified in
> terms of PeriodicWave, then I think it is better to specify that
> it is equivalent to a PeriodicWave with no normalization so that
> this aspect is clear.
>

​PeriodicWave's can't have a infinite number of coefficients so I don't
understand why you keep mentioning infinite number of coefficients.  I also
don't see how the normalization algorithm (in the pull request) is really
affected by the number of coefficients.  You compute the infinite sum
instead of the finite sum and find the max amplitude.​


>
> (Beyond this, I'm also not sure about the idea of specifying
> built-in waveforms in terms of PeriodicWave because I suspect it
> is not the most efficient implementation for quality
> representation of these waveforms at low fundamental frequencies,
>

​I think this is a done deal.  I seem to recall that we decided at the last
F2F to define the oscillator in terms of PeriodicWave.  I don't want to
reopen this, but if you do, please file a new issue instead of adding to
this.
​


> but perhaps better generators can still generate equivalent
> waveforms more efficiently. I don't think the maximum number of
> Fourier coefficients should be specified to resolve performance
> problems with PeriodicWave, because I suspect other algorithms do
> not have this problem.)
>

​That is certainly possible, but I think the burden is on you to provide
these other algorithms and propose them as the method to use for the
oscillators and also reconcile it with PeriodicWave.

I find defining oscillators in terms of PeriodicWave very nice.​


>
> > And then
> > band-limiting will produce ringing that causes the output to exceed 1.
>
> Yes, the output will exceed 1.
>
> I don't see that as a problem because oscillators should not be
> connected to speakers at full volume.
>
> If that really were a problem, then we would need a normalization
> that provides peaks with abs <= 1 for all fundamental frequencies.
> Such normalization would be different from that of PeriodicWave.
>
> If backward compatibility with the existing implementation is the
> goal, then for clarity I would suggest specifying a factor/divisor
> that is independent of sample rate. Clients can then know the
> distance between flats of a square wave or slope of a sawtooth.
>
> For square and sawtooth I would suggest calculating that
> mathematically from the peak to peak distance of a band-limited
> step function. I expect that would provide the limit of the peaks
> of square and sawtooth waves as the number of Fourier coefficients
> approaches infinity. (The coefficients approach zero at the same
> rate as the slopes of the sine terms approach infinity, so the
> limit of the peak value of the discrete wave as the number of
> coefficients approaches infinity is not the same as the peak value
> of the continuous wave.)
>

​I'm pretty sure that infinite sum of the Fourier series is in fact the
exact waveforms.  Thus, if you did the infinite sum for the square wave,
you do in fact get a square wave that is exactly +1 or -1 (and 0 at the
jump).  No normalization needed. It's only when the sum is finite do you
get the ringing.​

​

>
> For triangle and sine, the factor can be 1 and still be
> consistent enough with the existing implementation.
>
> —
> <https://github.com/WebAudio/web-audio-api/pull/552#issuecomment-125079733>
> .
>



-- 
Ray

karlt:
rtoy writes:

> On Sun, Jul 26, 2015 at 9:04 PM, Karl Tomlinson <notifications@github.com>
> wrote:
>
> ​I also
> don't see how the normalization algorithm (in the pull request) is really
> affected by the number of coefficients.  You compute the infinite sum
> instead of the finite sum and find the max amplitude.​

I had been assuming that
https://github.com/WebAudio/web-audio-api/issues/447#issuecomment-113010229
and this pull request might imply that the intention was that
normalization from PeriodicWave would change the amplitude of the
built-in waveforms, and I was concerned that change might come
from using a finite sum for the normalization factor.

If that is not the case and the infinite sum is intended for
normalization, are we agreed that infinite sum should be taken as
1 for the built-in waveforms and so the normalization factor has no
effect?

rtoy:
On Mon, Jul 27, 2015 at 2:58 PM, Karl Tomlinson <notifications@github.com>
wrote:

> rtoy writes:
>
> > On Sun, Jul 26, 2015 at 9:04 PM, Karl Tomlinson <
> notifications@github.com>
> > wrote:
> >
> > ​I also
> > don't see how the normalization algorithm (in the pull request) is really
> > affected by the number of coefficients. You compute the infinite sum
> > instead of the finite sum and find the max amplitude.​
>
> I had been assuming that
> https://github.com/WebAudio/web-audio-api/issues/447#issuecomment-113010229
> and this pull request might imply that the intention was that
> normalization from PeriodicWave would change the amplitude of the
> built-in waveforms, and I was concerned that change might come
> from using a finite sum for the normalization factor.
>

​Ah ha!  The intent of all of this is to codify existing practice (more or
less).  Chrome (and Safari) has always used PeriodicWave's as the basis for
the built-in oscillator types and the normalization has been done this way
for a long time.  The normalization uses the finite sum.  The difference
now is that the finite sum used to be for 2048 coefficients (always).  Now,
with the possibility of more coefficients, the normalization will use a
possibility of a different number of coefficients, but always finite.


> If that is not the case and the infinite sum is intended for
> normalization, are we agreed that infinite sum should be taken as
> 1 for the built-in waveforms and so the normalization factor has no
> effect?
>

​No, we're not agreed. :-)  Your proposal here is that, essentially, the
waveforms should never be normalized; you should get the actual sum, even
if you have huge coefficients producing a huge amplitude signal.  Perhaps
that's a different, but closely related, issue.
​


> —
> <https://github.com/WebAudio/web-audio-api/pull/552#issuecomment-125353889>
> .
>



-- 
Ray

karlt:
rtoy writes:

> ​No, we're not agreed. :-)  Your proposal here is that, essentially, the
> waveforms should never be normalized; you should get the actual sum, even
> if you have huge coefficients producing a huge amplitude signal.

Yes, although the built-in waveforms are mathematically defined,
which defines their magnitudes, so that there are no huge
coefficients producing huge amplitude signals.  See
http://webaudio.github.io/web-audio-api/#basic-waveform-phase

Also note that "the actual waveforms produced by the oscillator
may differ to prevent aliasing affects" but changes of magnitude
are not preventing aliasing affects.

> Perhaps that's a different, but closely related, issue.

Yes, my comments in this thread are re the magnitude issue and not
in opposition to the precise wording of this pull request.

I put my comments here in reply to
https://github.com/WebAudio/web-audio-api/pull/552#issuecomment-122971946
because of the potential relationship between approximation and
normalization.

I think it is fine to specify the desired waveforms and permit
implementations to approximate for performance reasons, as is
already specified.

I don't think it is a good idea to introduce limits on the quality
of the waveforms through limits on the number of coefficients.

rtoy:
On Mon, Jul 27, 2015 at 7:02 PM, Karl Tomlinson <notifications@github.com>
wrote:

> rtoy writes:
>
> > ​No, we're not agreed. :-) Your proposal here is that, essentially, the
> > waveforms should never be normalized; you should get the actual sum, even
> > if you have huge coefficients producing a huge amplitude signal.
>
> Yes, although the built-in waveforms are mathematically defined,
> which defines their magnitudes, so that there are no huge
> coefficients producing huge amplitude signals. See
> http://webaudio.github.io/web-audio-api/#basic-waveform-phase


​I was actually thinking of periodic waves used for the oscillator. I
believe you are basically arguing for no normalization at all, which would
mean I could have arbitrarily large coefficients for the periodic wave that
would produce arbitrarily large amplitude outputs from the oscillator.

I think we agree that if disableNormalization is true, this is the desired
result.​


>
>
> Also note that "the actual waveforms produced by the oscillator
> may differ to prevent aliasing affects" but changes of magnitude
> are not preventing aliasing affects.
>

​Good point.​


>
> > Perhaps that's a different, but closely related, issue.
>
> Yes, my comments in this thread are re the magnitude issue and not
> in opposition to the precise wording of this pull request.
>
> I put my comments here in reply to
> https://github.com/WebAudio/web-audio-api/pull/552#issuecomment-122971946
> because of the potential relationship between approximation and
> normalization.
>
> I think it is fine to specify the desired waveforms and permit
> implementations to approximate for performance reasons, as is
> already specified.
>

> I don't think it is a good idea to introduce limits on the quality
> of the waveforms through limits on the number of coefficients.
>


​But recall that the spec explicitly says the waveforms must be
band-limited.  Hence the mathematical waveforms are precluded. The number
of coefficients must be limited in some fashion.


-- 
Ray

karlt:
rtoy writes:

> ​But recall that the spec explicitly says the waveforms must be
> band-limited.  Hence the mathematical waveforms are precluded. The number
> of coefficients must be limited in some fashion.

Yes, for each particular fundamental frequency, it is important to
limit the number of terms, discarding from the specified
mathematical waveforms those terms that cannot be represented by
the sample rate.  I think this is sufficiently covered by "care
must be taken to discard (filter out) the high-frequency
information higher than the Nyquist frequency".  It is not
necessary to impose further limits than that on the number of
frequency components in the built-in waveforms.

rtoy:
On Tue, Jul 28, 2015 at 4:04 PM, Karl Tomlinson <notifications@github.com>
wrote:

> rtoy writes:
>
> > ​But recall that the spec explicitly says the waveforms must be
> > band-limited. Hence the mathematical waveforms are precluded. The number
> > of coefficients must be limited in some fashion.
>
> Yes, for each particular fundamental frequency, it is important to
> limit the number of terms, discarding from the specified
> mathematical waveforms those terms that cannot be represented by
> the sample rate. I think this is sufficiently covered by "care
> must be taken to discard (filter out) the high-frequency
> information higher than the Nyquist frequency". It is not
> necessary to impose further limits than that on the number of
> frequency components in the built-in waveforms.
>

​Yes.  I don't have any problems if implementations want to use more terms
as long as the result is band-limited.​


> —
> <https://github.com/WebAudio/web-audio-api/pull/552#issuecomment-125777614>
> .
>



-- 
Ray

rtoy:
What's the final conclusion here?

This PR is just to clarify that the builtin oscillator types must look as if they were generated from a PeriodicWave (normalized) with an appropriate number of coefficients.

karlt:
I don't have a strong objection to the sentence added here, but I don't think
it is necessary and it doesn't resolve
https://github.com/WebAudio/web-audio-api/issues/447

joeberkovitz:
@rtoy will come up with a concrete proposal including the exact PeriodicWave coefficient series yielding the exact equivalent to each builtin oscillator type.

rtoy:
The description of the coefficients were done in pull request #569.

I think all that really remains is that we need to say the basic oscillators must produce the same output as if a PeriodicWave were used with the appropriate coefficients (in #569), with normalization enabled.

padenot:
I think we can merge this.

PR #553, opened 2015-06-18
Fix #547: Define index for setValueCurveAtTime

### LINKED ISSUE #547: SetValueCurveAtTime interpolation not specified
The description in http://webaudio.github.io/web-audio-api/#widl-AudioParam-setValueCurveAtTime-void-Float32Array-values-double-startTime-double-duration shows how to calculate the value from the values array using an index of N * (t - startTime) / duration into the values array.  This index is very likely not an integer, so it's not clear how to select the actual index.

In chrome, the index is basically floor(0.5 + N * (t - startTime) / duration).

rtoy:
The indexing into the values array was not well-defined.  Specify that
the floating-point index is rounded to the nearest integer to index
into the array.  No interpolation is done.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1699,18 +1699,19 @@ The following rules will apply when calling these methods:
         calculated according to the <em>values</em> parameter.
       </p>
       <p>
-        During the time interval: <code>startTime</code> &lt;= t &lt;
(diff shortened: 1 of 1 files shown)


padenot:
(I had to rebase a little, but I think it's correct).

PR #556, opened 2015-06-19
Fix #323: Add IIRFilterNode

### LINKED ISSUE #323: Biquad: Possibility to set the filters coefficients and not just Q/freq/gain
Currently it is only possible to design a filter by using the Q/freq/gain parameters.
It seems that the solution would be to show the setNormalizedCoefficients function of the biquad node on the API.

void Biquad::setNormalizedCoefficients(double b0, double b1, double b2, double a0, double a1, double a2) ->  http://dxr.mozilla.org/mozilla-central/source/content/media/webaudio/blink/Biquad.cpp

rtoy:
This adds a new IIRFilterNode that implements a general IIR filter.
The filter coefficients are fixed and cannot be changed.  The maximum
number of coefficients is 20.

Is "IIRFIlterNode" the desired name?  What is the limit on the number of coefficients?  Should there be any?  In the constructor, the order of the arguments is the b (numerator coefficients) and a (denominator coefficients).  Is this the correct order?

### DIFF
--- a/index.html
+++ b/index.html
@@ -778,7 +778,7 @@ or <a href="#widl-AudioContext-decodeAudioData-Promise-AudioBuffer--ArrayBuffer-
       <dl class=parameters>
         <dt> Float32Array feedforward</dt>
 	<dd>
-          An array of the feedforward (numerator) coefficients for the tranfer function of the IIR
+          An array of the feedforward (numerator) coefficients for the transfer function of the IIR
           filter. The maximum length of this array is 20.  If all of the values are zero, an
           InvalidStateError MUST be thrown.  A NotSupportedError MUST be thrown if the array length
           is 0 or greater than 20.
--- a/index.html
+++ b/index.html
@@ -4386,13 +4386,15 @@ The number of channels of the output always equals the number of channels of the
     <dt>Float32Array magResponse</dt>
     <dd>
       This parameter specifies an output array receiving the linear magnitude
-      response values.
+      response values.  If this array is shorter than <code>frequencyHz</code> a NotSupportedError
+      MUST be signaled. 
     </dd>
     <dt>Float32Array phaseResponse</dt>
-   <dd>
-    This parameter specifies an output array receiving the phase response values
-    in radians.
-   </dd>
+    <dd>
+      This parameter specifies an output array receiving the phase response values
+      in radians.  If this array is shorter than <code>frequencyHz</code> a NotSupportedError
+      MUST be signaled.
+    </dd>
   </dl>
   </dd>
 </dl>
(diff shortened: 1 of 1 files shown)


mdjp:
Typo in line 781 /tranfer/transfer

joeberkovitz:
Do we need size checking on output arrays for `getFrequencyResponse()`?

rtoy:
Yes, we should mention constraints on the output arrays (and the input array).  BiquadFilterNode needs this too.  I'll file a separate issue on that.

rtoy:
A couple of things I would like to change.

 * Instead of `Float32Array` args to `createIIRFilter`, I'd like `sequence<float>` instead.  Then you can simply say `createIIRFilter([1,-.9],[1])` instead of doing `new Float32Array(2)` and so on.
 * Octave and matlab's `filter` function (IIR filtering) has the arguments in the opposite order.  That is, the feedback (numerator) coefficients are the first arg, and the feedback (denominator) coefficients are the second.

joeberkovitz:
These both sound like good changes to me: both the move to `sequence<float>` and the argument reordering to preserve the reading order of "numerator / denominator".

rtoy:
How about making the arguments to getFrequencyResponse also be
sequence<float> too instead of Float32Array?  This makes it different from
BiquadFilterNode, but we can change BiquadFilterNode.getFrequencyResponse
to do the same since it's backward compatible.

On Mon, Aug 31, 2015 at 7:27 AM, Joe Berkovitz <notifications@github.com>
wrote:

> These both sound like good changes to me: both the move to sequence<float>
> and the argument reordering to preserve the reading order of "numerator /
> denominator".
>
> —
> <https://github.com/WebAudio/web-audio-api/pull/556#issuecomment-136387647>
> .
>



-- 
Ray

joeberkovitz:
I think that the getFrequencyResponse case seems a little different since there is no particular upper limit on the number of frequencies passed in. The array could potentially be quite large if the caller is trying to grab a high-resolution spectrum, which maybe makes this a little bit like extracting frequency response data from AnalyzerNode. So I see more of a case for the Float32Array type here...

rtoy:
Ok.  I thought this might be a small convenience.

In any case, ES6 allows you to do Float32Array.from([1,2,3,4]), so it's not
too much of an inconvenience.

On Mon, Aug 31, 2015 at 9:52 AM, Joe Berkovitz <notifications@github.com>
wrote:

> I think that the getFrequencyResponse case seems a little different since
> there is no particular upper limit on the number of frequencies passed in.
> The array could potentially be quite large if the caller is trying to grab
> a high-resolution spectrum, which maybe makes this a little bit like
> extracting frequency response data from AnalyzerNode. So I see more of a
> case for the Float32Array type here...
>
> —
> <https://github.com/WebAudio/web-audio-api/pull/556#issuecomment-136427023>
> .
>



-- 
Ray

PR #562, opened 2015-07-03
fix #509, change k-rate to a-rate.

### LINKED ISSUE #509: AudioParam rates for BiquadFilterNode are inconsistant

billhofmann:
simple one line change

### DIFF
--- a/index.html
+++ b/index.html
@@ -4050,7 +4050,7 @@ The number of channels of the output always equals the number of channels of the
 </dl>
 
 <p>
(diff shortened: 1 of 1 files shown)


rtoy:
Thanks for the fix!

PR #576, opened 2015-08-24
Update wording for AudioBufferSourceNode start (called only once) and stop (can be called multiple times).

billhofmann:
Modification to intro paragraph, start method, when parameter.


padenot:
Thanks !

billhofmann:
Just picking easy fixes :)

PR #578, opened 2015-08-25
Clarify use of AudioBufferSourceNode and provide recommendation for non-memory assets

billhofmann:
and split the paragraph...


padenot:
Thanks !

Issue #344, opened 2014-07-25
cancelScheduledValues should set the current value as a parameter change


cwilso:
cancelScheduledValues should explicitly create a parameter value change in the event list of the current value, at that time.  This will enable smooth ramps from the current value.

joeberkovitz:
TPAC resolution: We will define `cancelScheduledValues` as deleting any extant automation events scheduled at/after the automation time, Clarify the spec to point out that this may cause discontinuities due to the disappearance of ramps whose `endTime` is scheduled after the cancellation point.

When cancelling automation and there is a desire to continue the parameter from the current value, recommend the following sequence:

```
var currentValue = param.value;
param.cancelScheduledValues(ctx.currentTime);
setValueAtTime(currentValue, ctx.currentTime);
```

mark-buer:
That is a very disappointing decision.

Perhaps the spec should also state this advisory to consumers of the API:
"The automation methods are designed for set-and-forget automations. If your use-case demands glitch-free audio in the face of changing automations, such as those required in many interactive applications, the automation methods are probably not useful."
Or words to that effect.

Out of curiosity, what happens if an audio render event occurs part-way-through the suggested value continuation code sequence?
```javascript
var currentValue = param.value;
// !!! Audio render occurs *right now* on the audio thread !!!
param.cancelScheduledValues(ctx.currentTime);
setValueAtTime(currentValue, ctx.currentTime);
```

My use-case for cancel-and-hold:
Assume that long-running gain automations (of the order of seconds) are in effect on many (lets say seven or eight) gain nodes of a moderately complex audio graph. Each gain node requires a separate automation curve. The number of gain nodes changes with respect to time (over the order of tens of seconds). Unpredictable user interactions trigger a change of all existing automations. It is necessary to change the automations in a way that avoids clicks and pops. It is acceptable for there to be a slight lag (of the order of tens of milliseconds) between user interaction and the change of automation.

Solving that use-case is trivial if cancel-and-hold exists, but extremely complex if it doesn't exist.
Right now I'm dreaming up ways of cross fading between old gain nodes with old automations and new gain nodes with new automations... or perhaps a super-wide audio worker (with tens of channels) fed into a channel splitter that will provide the automation values... or... dunno. Not easy. Very complex.

[...]

PR #580, opened 2015-08-27
Clarify cancelScheduledValues behavior for setTargetAtTime

billhofmann:
clarified wording

### DIFF
--- a/index.html
+++ b/index.html
@@ -1767,7 +1767,9 @@ The following rules will apply when calling these methods:
     <dd>
       <p>
         Cancels all scheduled parameter changes with times greater than or
-        equal to <code>startTime</code>.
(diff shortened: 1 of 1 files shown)


mrkishi:
What does "active" mean? What should happen in this case?

```
param.setValueAtTime(440, context.currentTime);
param.setTargetAtTime(880, context.currentTime + 5, 100);
param.cancelScheduledValues(context.currentTime + 10);
```

Related question: what about curve values?

```
param.setValueAtTime(440, context.currentTime);
param.setCurveValueAtTime(curve, context.currentTime + 5, 10);
param.cancelScheduledValues(context.currentTime + 10);
```

rtoy:
@mrkishi I expect your questions are answered (or will be) in #344

Issue #503, opened 2015-03-25
AudioParam setter does not work after the execution of automation.


hoch:
This is something to think about  - I don't think this behavior is clearly specified. What happens if the setter is used after `cancelScheduledValues()` method called?

~~~js
var context = new AudioContext();
var sine = context.createOscillator();

sine.connect(context.destination);
sine.start();

sine.frequency.setValueAtTime(440.0, 0);
sine.frequency.linearRampToValueAtTime(880.0, 3);
sine.frequency.cancelScheduledValues(3.5);

var freq = 660.0;

function setValue() {
  freq = (freq === 660.0) ? 770.0 : 660.0;
  sine.frequency.value = freq;
  console.log('current time: ' + context.currentTime);  
  console.log('settting frequency: ' + freq);
  setTimeout(setValue, 2000);
}

setValue();
~~~

FYI, currently FF and Chromium don't do anything with the setter. Once the automation happens, it locks up the setter for good.

notthetup:
Hmm! I didn't know this. This might explained a few issues I was noticing with cancelling automation. Locking up the setter seems very unintuitive. 

I would expecting a setter call, to internally cancel all automation and then set the value to the argument of the setter. 

Otherwise it might make sense to somehow indicate that the setter is locked up after the automation starts.

rtoy:
Not all is lost, though.  setValueAtTime still works and has the same effect as using the setter, except for the de-zippering that might happen.  If de-zippering is needed you can use setTargetAtTime with an appropriate time constant.

I think the main issue is what happens to setTargetAtTime automations that are essentially infinite in duration.  If it's already started, what does cancelScheduledEvents to with that?  Or with any automation that has started but has not stopped yet?

notthetup:
@rtoy Good point.

I have gotten into a habit of doing this this every time I cancel any automation, because I couldn't wasn't sure what was actually happening. I know it's not perfect, but it works.

```
var currentVal = osc.frequency.value;
osc.frequency.cancelScheduledValues(context.currentTime);
osc.frequency.setValueAtTime(currentVal, context.currentTime);
```

rtoy:
Ah, this is a duplicate of issue #128

hoch:
Just to leave a note about this issue, here's the excerpt from the [section 2.5.3](http://webaudio.github.io/web-audio-api/#h4_computation-of-value):

> If the value attribute is set after any automation events have been scheduled, then these events will be removed.

[...]

PR #583, opened 2015-08-27
Fix #128: clarify effect of setting the value attribute with respect …

### LINKED ISSUE #128: Note that setting AudioParam.value will be ignored when any automation events have been set on the object
> Originally reported on W3C Bugzilla [ISSUE-19765](https://www.w3.org/Bugs/Public/show_bug.cgi?id=19765) Mon, 29 Oct 2012 23:07:00 GMT
> Assigned to 

We should note that if the author sets the value property on an AudioParam which has any automation events scheduled, the new value will be silently ignored without any exception being raised.

joeberkovitz:
…to equivalent parameter automation action.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1556,6 +1556,9 @@ The following rules will apply when calling these methods:
       when there are any automation events scheduled then it will be ignored and
       no exception will be thrown.
     </p>
+    <p>
+      The effect of setting this attribute is equivalent to calling <code>setValueAtTime()</code> with the current <code>AudioContext</code>'s <code>currentTime</code> and the requested value. Subsequent accesses to this attribute's getter will return the same value.
+    </p>
   </dd>
   <dt> readonly attribute float defaultValue </dt>
   <dd>


rtoy:
While we agreed to this a while back, during our uncall today, @cwilso, @hoch, and @rtoy coincidentally chatted (off-line) about this.  I think we need to discuss the implications with everyone.  Does equivalent mean actually doing everything that setValueAtTime would do, including adding an event into the timeline?

joeberkovitz:
Thanks for sharing with the class :-)  How about I reword this to indicate that it is identical to setValueAtTime in terms of value changes, but there is no actual automation event in the timeline (which would drag in a bunch of other baggage including, strangely, disabling subsequent assignments to the setter)?

joeberkovitz:
Oh now I see your comments in #503 which sounds like you're headed the other way: you want it to be equivalent (and in turn to have the setter no longer cancel automations). Yeah, we should discuss. I like that idea too, at least right now.

PR #586, opened 2015-09-03
Fix #5: Add setTargetAtTime curve to example

### LINKED ISSUE #5: AudioParam Automation Example graph missing setTargetAtTime
> Originally reported on W3C Bugzilla [ISSUE-17701](https://www.w3.org/Bugs/Public/show_bug.cgi?id=17701) Thu, 05 Jul 2012 15:03:59 GMT
> Assigned to 

The example at the end of the section on AudioParam Automation is giving examples for setValueAtTime, linearRampToValueAtTime, exponentialRampToValueAtTime and setValueCurveAtTime but not setTargetValueAtTime - which is too bad since that one seems to be the one hardest to comprehend.

rtoy:
The graph of parameter automation is updated to include an example of
setTargetAtTime with the code update appropriately.  The graph also
includes some simple annotation of each part and also gives an example
of how to have another automation follow a setTargetAtTime.  A
setValueAtTime is required to terminate the setTargetAtTime and give a
starting value for the following exponentialRampToValueAtTime.


rtoy:
The updated automation graph:  The updated automation graph

chrislo:
Is that discontinuity at `t=0.2` correct, @rtoy? Shouldn't the linear ramp start at `t=0.2, A=0.3` and end at `t=0.3, A=1`?

rtoy:
@chrislo: Oops.  You're right, there's a setValueAtTime(0.4, 0.3) that was in the code, but no annotation for it.  I'll fix that shortly.

rtoy:
Updated the graph to be:
screenshot from 2015-09-10 14 17 51

Also added a comment on how the graph was generated.

I thought it would have been really cool if the spec could have generated the graph dynamically using WebAudio, just like the same code in the description (with an image fallback for browsers without WebAudio).  However, I failed to get flot graphing to work with ReSpec, and, most importantly, the graph looks very different in Chrome and Firefox.

I think, however, that the current graph (above) is the expected graph for the automations scheduled.

padenot:
As I told Raymond the other day, Firefox currently has issues with `AudioParam`s, we are fixing those.

PR #588, opened 2015-09-10
Clarify the meaning of the operation "acquire the content".

### LINKED ISSUE #567: Should copyToChannel affect AudioBuffers acquired by an AudioBufferSourceNode/ConvolverNode?
Should copyToChannel affect AudioBuffers that have already been acquired by AudioNodes and used for audio processing?  Here's an example:

```javascript
var bufferSource = context.createBufferSource();
bufferSource.buffer = myAudioBuffer;

// After start(), the buffer source node "acquires the contents of the AudioBuffer"
bufferSource.start(); 

// Should the channel data write impact playback of the bufferSource node above?
myAudioBuffer.copyToChannel(myChannelData, /*channelIndex*/ 0);  
```
For this example, in FireFox, copyToChannel does not affect playback of the already started AudioBufferSourceNode node.  Instead, it creates a new copy of the AudioBuffer for JavaScript and modifies it.  I believe this is consistent with getChannelData().  Should we add this behavior to the spec?  I plan to implement the same behavior as FireFox in Edge.

padenot:
This fixes #567

### DIFF
--- a/index.html
+++ b/index.html
@@ -2068,6 +2068,7 @@ interface is an <a><code>AudioNode</code></a> with a single input and single out
   invoker.
 </p>
 
+
 <p>
 When an <dfn id="acquire-the-content">acquire the content</dfn> operation occurs on an
 <a>AudioBuffer</a>, run the following steps:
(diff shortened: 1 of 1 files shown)


rtoy:
lgtm

PR #590, opened 2015-09-10
Clarify that computedPlaybackRate must be implemented by resampling the audio.

### LINKED ISSUE #497: computedPlaybackRate and pitch
A question about Section 2.9.1 and Section 2.9.2:

Per the current wording of the specification, both `playbackRate` and `detune` affect the speed of the audio stream.  In all current implementations, this affects the pitch as well.  To anybody with audio experience, this is expected.

However, based on the current wording, would it be possible for an implementor to use a time stretching algorithm to alter the "effective speed" of the AudioBuffer, keep the pitch the same, and still technically be in compliance with the specification.

Specifically: should there be wording that clarifies how the `computedPlaybackRate` actually affects the pitch of the audio.

padenot:
This fixes #497.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2255,14 +2255,22 @@ invoked in the following cases:
 Both <code>playbackRate</code> and <code>detune</code> are <a>k-rate</a>
 parameters and are used together to determine a <em>computedPlaybackRate</em>
 value:
+</p>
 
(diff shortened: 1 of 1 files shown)


rtoy:
lgtm.

PR #592, opened 2015-09-10
Clarify uninterrupted operations of unconnected nodes

### LINKED ISSUE #8: Behavior of unconnected nodes needs to be specified
> Originally reported on W3C Bugzilla [ISSUE-17794](https://www.w3.org/Bugs/Public/show_bug.cgi?id=17794) Tue, 17 Jul 2012 18:34:19 GMT
> Reported by Chris Wilson
> Assigned to 

This can be described as "What happens when a playing node is temporarily disconnected?" - or, conversely, "If you play a node, and no one is listening (aka connected), does it really play?".

This first came to my attention when I was working with Z Goddard on the Fieldrunners article for HTML5Rocks (http://www.html5rocks.com/en/tutorials/webaudio/fieldrunners/) - particularly, read the section entitled Pausing Sounds.  In short - they'd noticed that if you disconnected an audio connection, it paused the audio "stream".  I thought this seemed pretty wrong - knowing what I knew about how automation on AudioParams works - in discussions with Chris Rogers, he confirmed this wasn't his expected behavior.

My mental model of connections as an API user still really wants to be "they're just like plugging 1/4" audio cables between hardware units," despite knowing that is not the case here; I would expect if a node was playing and I disconnected its graph, then replugged it 0.5 sec later, it would be 0.5 sec further along - i.e., I would expect the behavior to be the same as if I had connected the node to a zero-gain gain node connected to the audiocontext.destination.

joeberkovitz:
Fixes #8

### DIFF
--- a/index.html
+++ b/index.html
@@ -1059,6 +1059,9 @@ and the <a href="#channel-up-mixing-and-down-mixing"></a>
 
 </p>
 
+<p>The processing of inputs and the internal operations of an <a>AudioNode</a> take place continuously with respect to <a>AudioContext</a> time, regardless of whether the node has connected outputs, and regardless of whether these outputs ultimately reach an <a>AudioContext</a>'s <a>AudioDestinationNode</a>.
+</p>
+
 <p>
 For performance reasons, practical implementations will need to use block
 processing, with each <a><code>AudioNode</code></a> processing a fixed number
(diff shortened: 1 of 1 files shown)


joeberkovitz:
@karlt I just pushed an additional commit that adds a note on this point to the informative "node lifetime" section. I thought it was a good idea to get this in along with the spec change.

rtoy:
lgtm

PR #595, opened 2015-09-10
34-change language on audio file formats to include video-element-supported formats

billhofmann:


index.html
```diff
@@ -597,7 +597,7 @@ <h2 id="AudioContext">The AudioContext Interface</h2>
     The ArrayBuffer can, for example, be loaded from an XMLHttpRequest's
     <code>response</code> attribute after setting the <code>responseType</code>
     to "arraybuffer".  Audio file data can be in any of the formats supported by
-    the <code>audio</code> element.
+    the <code>audio</code> or <code>video</code> elements.
```
padenot: I think simply saying `HTMLMediaElement` is better here, so it's really clear what we mean.

Also there was a tentative of layering `HTMLMediaElement` on top of the Web Audio API, in which case this becomes problematic.

billhofmann:
Going to re-implement on updated gh-pages

PR #596, opened 2015-09-10
Add reference(s) to use cases in intro + near end

billhofmann:
Note that section numbering at end is a little wonky


rtoy:
lgtm.

PR #600, opened 2015-09-23
Link to the mimesniff spec in `decodeAudioData`

### LINKED ISSUE #57: Make the usage of content sniffing for decodeAudioData explicit
> Originally reported on W3C Bugzilla [ISSUE-21702](https://www.w3.org/Bugs/Public/show_bug.cgi?id=21702) Mon, 15 Apr 2013 21:46:34 GMT
> Assigned to 

We should probably explicitly say in the spec that the buffer passed to decodeAudioData should be sniffed to get the content type, according to the http://mimesniff.spec.whatwg.org/ spec, and that the implementation must reject buffers that do not pass that content sniffing.

This will help in situations like https://bugzilla.mozilla.org/show_bug.cgi?id=861187, where one implementation (Gecko) cannot sniff the content type properly but another (WebKit) can.

padenot:
This fixes #57

### DIFF
--- a/index.html
+++ b/index.html
@@ -641,7 +641,10 @@ or <a href="#widl-AudioContext-decodeAudioData-Promise-AudioBuffer--ArrayBuffer-
     The ArrayBuffer can, for example, be loaded from an XMLHttpRequest's
     <code>response</code> attribute after setting the <code>responseType</code>
     to "arraybuffer".  Audio file data can be in any of the formats supported by
-    the <code>audio</code> element.
+    the <code>audio</code> element.  The buffer
+    passed to <a
+    href="#widl-AudioContext-decodeAudioData-Promise-AudioBuffer--ArrayBuffer-audioData-DecodeSuccessCallback-successCallback-DecodeErrorCallback-errorCallback">decodeAudioData</a>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -641,7 +641,10 @@ <h2 id="AudioContext">The AudioContext Interface</h2>
     The ArrayBuffer can, for example, be loaded from an XMLHttpRequest's
     <code>response</code> attribute after setting the <code>responseType</code>
     to "arraybuffer".  Audio file data can be in any of the formats supported by
-    the <code>audio</code> element.
+    the <code>audio</code> element.  The buffer
+    passed to <a
+    href="#widl-AudioContext-decodeAudioData-Promise-AudioBuffer--ArrayBuffer-audioData-DecodeSuccessCallback-successCallback-DecodeErrorCallback-errorCallback">decodeAudioData</a>
+  has its content-type determined by sniffin, as described in [[mimesniff]].
```
notthetup: Should it be 'sniffing' instead of 'sniffin'?

padenot: Yep, I just pushed a fix.

rtoy:
lgtm

PR #603, opened 2015-10-01
Fix #40 Add detail of connecting audio node to AudioParam.

### LINKED ISSUE #40: Add detail of connecting audio node to non audio node
> Originally reported on W3C Bugzilla [ISSUE-21538](https://www.w3.org/Bugs/Public/show_bug.cgi?id=21538) Tue, 02 Apr 2013 16:00:50 GMT
> Assigned to 

In section "The connect to AudioParam method", add detail of connecting audio node to non audio node. An explanation on why would you do it (LFO example) could be added to the graph routing introduction.

joeberkovitz:


joeberkovitz:
I agree with your suggestions, thanks! Please see updated diagram.

rtoy:
Looks great!

Issue #68, opened 2013-09-11
Specify the data model for AudioContext.createPeriodicWave


olivierthereaux:
> Originally reported on W3C Bugzilla [ISSUE-22830](https://www.w3.org/Bugs/Public/show_bug.cgi?id=22830) Mon, 29 Jul 2013 18:48:53 GMT
> Reported by Ehsan Akhgari [:ehsan]
> Assigned to 

We need to specify exactly what happens when you call AudioContext.createPeriodicWave.  Roc's proposal was that we should copy the arrays passed into this method when it gets called and use the internal copies from that point on so that we don't race with those arrays being modified on the main thread.  I support that proposal.

Issue #67, opened 2013-09-11
Specify the data model for AudioParam.setValueCurveAtTime


olivierthereaux:
> Originally reported on W3C Bugzilla [ISSUE-22831](https://www.w3.org/Bugs/Public/show_bug.cgi?id=22831) Mon, 29 Jul 2013 18:49:41 GMT
> Reported by Ehsan Akhgari [:ehsan]
> Assigned to 

We need to specify exactly what happens when you call AudioParam.setValueCurveAtTime.  Roc's proposal was that we should copy the arrays passed into this method when it gets called and use the internal copies from that point on so that we don't race with those arrays being modified on the main thread.  I support that proposal.

PR #608, opened 2015-10-08
Fix #65 by clarifying copy-on-write behavior of WaveShaperNode.curve.

### LINKED ISSUE #65: Specify the data model for WaveShaperNode.curve
> Originally reported on W3C Bugzilla [ISSUE-22832](https://www.w3.org/Bugs/Public/show_bug.cgi?id=22832) Mon, 29 Jul 2013 18:50:44 GMT
> Assigned to 

We need to specify exactly what happens when you set WaveShaperNode.curve to a new array value.  Roc's proposal was that we should copy the arrays set to this attribute when it the attribute is set and use the internal copies from that point on so that we don't race with those arrays being modified on the main thread.  I support that proposal.

joeberkovitz:


joeberkovitz:
Also addresses #67 and #68 which were mostly covered by the existing text, but these new changes make the language consistent by referring to "copies".

Issue #326, opened 2014-05-27
Panner distance model with distance < refDistance


rtoy:
The formulas for the distance model in http://webaudio.github.io/web-audio-api/#idl-def-DistanceModelType shows how to compute the distance effect depending on the distance between the source and listener.

However, no information is given if the distance is less than refDistance, which defaults to 1. The formulas probably should say min(distance,refDistance) so that the audio volume doesn't increase arbitrarily as distance approaches 0.

Issue #328, opened 2014-06-02
Panner distance model with distance > maxDistance


rtoy:
The formulas for the distance models don't indicate what happens if distance > maxDistance.  However, the description for maxDistance says that this is the maximum distance after which the volume is not further decreased.

The formulas should probably include this to make it more obvious.

PR #609, opened 2015-10-08
Update formulas for the panner distance models to say the distance is clamped

rtoy:
Update formulas for the panner distance models to say the distance is
clamped between d_ref and d_max (if appropriate).

Fix WebAudio/web-audio-api#326 and WebAudio/web-audio-api#328.


rtoy:
Close this and redo later because it's too hard to rebase this after the big tidying update.

Issue #247, opened 2013-10-03
Filter node pops when frequency parameter is modulated below 0


joshontheweb:
When using an oscillator to modulate the frequency parameter of a filter node you get a popping sound when the modulation dips the frequency value below 0.  A common case for this is when you are modulating the filter cutoff with an LFO.  example code here http://jsfiddle.net/hyf5N/

The desired effect is a clean cutoff to silence when the frequency value is at or below 0

padenot:
This might be solved by #391, that proposes to clamp when an AudioParam is outside is nominal range. The nominal range for the frequency of a biquad is [10; sampleRate/2] (I'm not sure why we put a 10 in there as the lower bound, this seems restrictive).

rtoy:
Why 10? This seems rather arbitrary.  Certainly, people won't be able to hear anything that is output from such a filter, but the filter could be going into an audioparam.

padenot:
Yeah, zero is fine.

joeberkovitz:
@padenot Is this being addressed in #391 by #620? It seems as though it should, but the nominal range for the biquad frequency there is -Nyquist, +Nyquist

padenot:
Yes. The biquad should be [0, Nyquist], that was a bug I just fixed.

rtoy:
I think this can be closed. #620 has specified the appropriate ranges, and I don't hear any pops from the test case (once you change createGainNode to createGain).

PR #620, opened 2015-10-12
Spec nominal ranges, nyquist, and behavior when outside the range.

### LINKED ISSUE #391: Specify nominal ranges for AudioParam
Some of them already have a nominal range, but some are missing. I also propose that we clamp the values to the nominal range if it happens to be outside the range.

The one that are missing, along with a proposal, are:
- GainNode.gain: `[-sampleRate/2; sampleRate/2]` (that is, the max of any other AudioParam, for AudioNode -> GainNode -> AudioParam purposes)
- AudioBufferSourceNode.playbackRate: `[0; 10000]`. The upper range is somewhat arbitrary

padenot:
This fixes #391.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2337,7 +2337,7 @@ function setupRoutingGraph() {
           value changes can be sampled more coarsely. Each individual
           <code>AudioParam</code> will specify that it is either an
           <a>a-rate</a> parameter which means that its values must be taken
-          into account on a per-audio-sample basis, or it is <a>k-rate</a>
+          into account on a per-audio-sample basis, or it is a <a>k-rate</a>
           parameter.
         </p>
         <p>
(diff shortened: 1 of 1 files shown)


padenot:
How about something like this ?

padenot:
Is this waiting for something ?

joeberkovitz:
This also fixes #247

rtoy:
What's the final conclusion here?  Add back the max and min attributes for every node that needs it? Perhaps they should be readonly so that the user can't do arbitrary things?

joeberkovitz:
I propose that we begin by specing these in V1 and defer adding the discoverable stuff since we haven't discussed it in this cycle, and it may not be as easy to add back in as we think. We can't keep adding new features and this is a feature.

rtoy:
It's not really a new feature; it's putting back what we used to have but ripped out. :-)

joeberkovitz:
Well conservatively, putting on my "get this done" hat, that makes it a new feature... but if you and Paul feel you can implement this easily I will beat a retreat on this.

rtoy:
Since Paul is the author, I'll let him decide what to do. I'm ok with either.

joeberkovitz:
Resolution agreed by @rtoy and @padenot Merge this PR and make a new one with min and max discoverability.

padenot:
This is rebased, squashed, and tidied.

index.html
```diff
@@ -6168,29 +6203,30 @@ <h3 id="example-2">
...
+            0. It forms a <a>compound parameter</a> with <code>frequency</code>.
           </dd>
           <dt>
             readonly attribute AudioParam Q
           </dt>
           <dd>
             The <a href="https://en.wikipedia.org/wiki/Q_factor">Q</a> factor
-            has a default value of 1, with a nominal range of 0.0001 to 1000.
+            has a default value of 1. Its <a>nominal range</a> is [0.0001,
+            1000].
```
rtoy: Do we really need to limit the Q values to this range? From the Wikipedia article, I think we only need to limit Q to the range (0, inf).

padenot: Yes, I think I just did a mechanical conversion here.

rtoy:
lgtm.

Issue #612, opened 2015-10-08
Behavior of exponentialRampToValue when the previous event is less than equal to 0


rtoy:
http://webaudio.github.io/web-audio-api/#widl-AudioParam-exponentialRampToValueAtTime-void-float-value-double-endTime says an error must be thrown if the value parameter is not strictly positive.  This is easy to understand and immediately indicates that the event is not inserted.

However, the description for the value parameter also says the exception must be thrown if the previous value (the start value V0) is not positive.  In practice, this error cannot be thrown until the exponentialRamp is being processed.  After the error is thrown, what should happen?  The formula is undefined for negative V0, but the event is already in the timeline.  

Should the timeline just hold the value from the end of the previous event?  That seems like the only reasonable thing to do and behaves as if the exponentialRamp were never inserted into the timeline.

padenot:
Yes I think you're right.

rtoy:
One other thing.  The formula is well-defined if both V0 and V1 are strictly negative.  Should that be allowed?  Then we have a problem if V0 and V1 have opposite signs, but we can throw an error in that case and just hold the previous value.

padenot:
That seems useful to me for some AudioParams.

joeberkovitz:
TPAC 2015: Modify spec to only throw if value is exactly zero. Redefine the formula to distinguish the case where V0 and V1 have opposite signs, in which case the value of the formula is V0 (i.e preserve the previous value). In case where both are strictly negative, arrange for the formula to invert both V0 and V1 and then invert its result.

rtoy:
I think we also said that we only throw for exponentialRampToValue(0, time).  That is, the error is only thrown if the value for the ramp is 0.  This implies that if v0 is 0 (or has opposite sign), no error is thrown because it's only knowable at the time the exponential is being run.  Throwing seems less useful at that point because there's nothing the application can do about it.

joeberkovitz:
Yes, that is my recollection too.

PR #624, opened 2015-10-13
Specify exponential ramp behavior when previous value is 0

rtoy:
Clarify what happens if the previous event value is 0.  We also
explicitly allow v0 and v1 to both be negative (the formula is
well-defined in this case).

If an error is thrown, the previous value is propagated.

Fixes WebAudio/web-audio-api#612.


index.html
```diff
@@ -2396,9 +2397,23 @@ <h2 id="AudioParam">
               </dt>
               <dd>
                 The value the parameter will exponentially ramp to at the given
-                time. A NotSupportedError exception MUST be thrown if this
-                value is less than or equal to 0, or if the value at the time
-                of the previous event is less than or equal to 0.
+                time. A NotSupportedError exception MUST be thrown if any
```
padenot: `<code>` around the exception name.

rtoy: Ok.  Are we going to do that everywhere else in the document?  We don't do that anywhere right now.

padenot: Yeah I think I have a patch that does it lying around in a branch.

index.html
```diff
@@ -2376,9 +2376,10 @@ <h2 id="AudioParam">
               - T_0}{T_1 - T_0} $$
             </p>
             <p>
-              where \(V_0\) is the value at the time \(T_0\) and \(V_1\) is the
-              <code>value</code> parameter passed into this method.  It is an error if either
-              \(V_0\) or \(V_1\) is not strictly positive.
+              where \(V_0\) is the value at the time \(T_0\) and \(V_1\) is
+              the <code>value</code> parameter passed into this method.  It
+              is an error if \(V_0\) and \(V_1\) have opposite signs or if
+              either \(V_0\) or \(V_1\) is zero.
```
rtoy: Actually, I have some second thoughts about this.  I think it's useful if V_1 is zero to signal an error.  But if you're running the automation and when the exponential ramp event becomes current and we then discover that V_0 is 0, throwing an error is kind of useless.  You get notified, but there's nothing you can do about it, and the error can happen long after the exponential ramp has already finished.

index.html
rtoy: Although I did this in this patch, couldn't we do this the same way that "MUST" is magically highlighted (by respec)?

PR #625, opened 2015-10-13
Fix #85 by clarifying AudioContext.currentTime and various events' pl…

### LINKED ISSUE #85: Need description of latency expectation in AudioProcessEvent
> Originally reported on W3C Bugzilla [ISSUE-17350](https://www.w3.org/Bugs/Public/show_bug.cgi?id=17350) Tue, 05 Jun 2012 11:41:13 GMT
> Assigned to 

Audio-ISSUE-59 (AudioProcessingEventScheduling): AudioProcessingEvent scheduling [Web Audio API]

http://www.w3.org/2011/audio/track/issues/59

Raised by: Philip Jägenstedt
On product: Web Audio API

How close to AudioContext.currentTime is AudioProcessingEvent.playbackTime expected to be when the audioprocess event is fired?

May a compliant implementation fire events arbitrarily far ahead of time in an arbitrary order? Presumably Web authors will make assumptions that the event is fired as close as possible to "now", so the spec should require this.

joeberkovitz:
…aybackTime.


index.html
```diff
@@ -7410,6 +7408,16 @@ <h2 id="Performance">
           animation frame-rates do. Depending on the application, a reasonable
           latency can be from as low as 3-6 milliseconds to 25-50 milliseconds.
         </p>
+        <p>
+          Implementations will generally seek to minimize overall latency.
+        </p>
+        <p>
+          Along with minimizing overall latency, implementations will generally seek to minimize
+          the difference between an <a><code>AudioContext</code></a>'s <code>currentTime</code>
+          and a <a><code>AudioProcessingEvent</code></a>'s <code>playbackTime</code>.
```
mark-buer: `and an <a><code>AudioProcessingEvent</code></a>'s ...`
instead of
`and a <a><code>AudioProcessingEvent</code></a>'s ...`
?

index.html
```diff
@@ -705,7 +705,9 @@ <h2 id="AudioContext">
           <dd>
             <p>
               This is a time in seconds which starts at zero when the context
-              is created and increases in real-time. All scheduled times are
+              is created and increases in real-time. At any given point, its value
+              represents the starting time of the next block of audio to be processed
```
padenot: This is hard to implement in practice, because it depends on the event loop load, etc.

index.html
joeberkovitz: Thanks -- this is resolved now.

index.html
joeberkovitz: @padenot @rtoy Could you please suggest alternate wording? The definition above is my best understanding of what is actually implemented so I need your help here.

karlt: Joe Berkovitz writes:

>> @@ -705,7 +705,9 @@ <h2 id="AudioContext">
>>            <dd>
>>              <p>
>>                This is a time in seconds which starts at zero when the context
>> -              is created and increases in real-time. All scheduled times are
>> +              is created and increases in real-time. At any given point, its value
>> +              represents the starting time of the next block of audio to be processed

Perhaps "indicates the time immediately after the last block of
audio processed".

It means almost the same thing, but may differ if there is a
block currently being processed, and avoids any hint of
implication that using the value of currentTime will guarantee
scheduling at precisely that time.

joeberkovitz:
I've adopted wording similar to @karlt's suggestion on the definition of AudioContext.currentTime.

joeberkovitz:
@rtoy @padenot @karlt Let me ask a (hopefully) clarifying question that would help us understand if the two ways of defining `currentTime` (start of next render vs. just after last render) are actually different.

Question: Are there any ways _other than at initialization or after a block render_ that `currentTime` can change? A case I can think of is just after a `resume()`, but before the next block has been rendered. In this case, we have to ask whether `currentTime` reflects the next render that will occur, or whether it still reflects the "frozen" value that has been sitting there since `suspend()` until a render takes place.

rtoy:
On Thu, Nov 12, 2015 at 6:44 AM, Joe Berkovitz <notifications@github.com>
wrote:

> @rtoy <https://github.com/rtoy> @padenot <https://github.com/padenot>
> @karlt <https://github.com/karlt> Let me ask a (hopefully) clarifying
> question that would help us understand if the two ways of defining
> currentTime (start of next render vs. just after last render) are
> actually different.
>
> Question: Are there any ways *other than at initialization or after a
> block render* that currentTime can change? A case I can think of is just
> after a resume(), but before the next block has been rendered. In this
> case, we have to ask whether currentTime reflects the next render that
> will occur, or whether it still reflects the "frozen" value that has been
> sitting there since suspend() until a render takes place.
>

​This is a good question. I'm not 100% sure, but I think Chrome reflects
the "frozen" value from suspend because the time is updated *after* the
block is rendered.

> —
> <https://github.com/WebAudio/web-audio-api/pull/625#issuecomment-156121619>
> .
>



-- 
Ray

joeberkovitz:
That's what I expected. And that is part of why I think that defining with respect to the last block rendered makes sense, rather than in terms of the start of the next block to be rendered.

rtoy:
Yeah, but how do reconcile that desire with the fact that the is no "last
block rendered" for the very first block?

If this is obvious consistent to every one, I'm not going to block this PR
for this reason.

On Thu, Nov 12, 2015 at 12:18 PM, Joe Berkovitz <notifications@github.com>
wrote:

> That's what I expected. And that is part of why I think that defining with
> respect to the last block rendered makes sense, rather than in terms of the
> start of the next block to be rendered.
>
> —
> <https://github.com/WebAudio/web-audio-api/pull/625#issuecomment-156222492>
> .
>



-- 
Ray

joeberkovitz:
I think the answer is that we state (as in this PR) that for the special case where no blocks have been rendered yet, `currentTime` is zero by definition.

Note that defining in terms of the "next block" start time has another subtle problem, sort of the mirror of the `resume()` problem I noted before. After you call `suspend()`, there really is no "next block", and may never be one.

To try to make this friendlier to developer I will clarify in the text that for a running context, it is always the case that currentTime is the same as the expected start time of the next block.

rtoy:
On Thu, Nov 12, 2015 at 1:56 PM, Joe Berkovitz <notifications@github.com>
wrote:

> I think the answer is that we state (as in this PR) that for the special
> case where no blocks have been rendered yet, currentTime is zero by
> definition.
>
> Note that defining in terms of the "next block" start time has another
> subtle problem, sort of the mirror of the resume() problem I noted
> before. After you call suspend(), there really is no "next block", and
> may never be one.
>

​Well, there could be if you resume.  Which is not like the previous block
at time 0 where there can never be one, no matter what you do. :-)​


> To try to make this friendlier to developer I will clarify in the text
> that for a running context, it is always the case that currentTime is the
> same as the expected start time of the next block.
>

​This works for me.​


> —
> <https://github.com/WebAudio/web-audio-api/pull/625#issuecomment-156248281>
> .
>



-- 
Ray

joeberkovitz:
I hope I have this in fairly good shape now.

joeberkovitz:
Yeah, now that I re-read it (I didn't write it!) the statement on non-uniformity seems really alarming and kind of wrong to me, especially since we're talking about a running context. I reworded it, see if you like this better. I said it is increased uniformly by the rendering thread. Of course, this might not look uniform to a main-thread observer, but that's a side point.

rtoy:
Thanks for your patience. It looks good and I've merged it, as you probably
already know.

On Thu, Nov 12, 2015 at 3:44 PM, Joe Berkovitz <notifications@github.com>
wrote:

> Yeah, now that I re-read it (I didn't write it!) the statement on
> non-uniformity seems really alarming and kind of wrong to me, especially
> since we're talking about a running context. I reworded it, see if you like
> this better. I said it is increased uniformly by the rendering thread. Of
> course, this might not look uniform to a main-thread observer, but that's a
> side point.
>
> —
> <https://github.com/WebAudio/web-audio-api/pull/625#issuecomment-156272749>
> .
>



-- 
Ray

karlt:
Joe Berkovitz writes:

> @rtoy @padenot @karlt Let me ask a (hopefully) clarifying question that would
> help us understand if the two ways of defining `currentTime` (start of next
> render vs. just after last render) are actually different.

FWIW as long as "next" render is the next after the last, and not
the next after a block currently rendering but partially complete,
these are the same.  Any usage of "next" would need to be careful
about exactly which it means.

> Question: Are there any ways _other than at initialization or after a block
> render_ that `currentTime` can change?

I would expect that currentTime can change only after a block
render.  It is zero at initialization.

> A case I can think of is just after a
> resume()`, but before the next block has been rendered. In this case, we have
> to ask whether `currentTime` reflects the next render that will occur, or
> whether it still reflects the "frozen" value that has been sitting there since
> suspend()` until a render takes place.

Even here, I would expect the frozen value from the time the
suspend promise resolved to match the value just after resume() but
before the next block has been rendered (assuming resume was called
after the suspend promise resolved).

However, I would not assume that the value is necessarily frozen
at the time of the suspend() call.  If there is a block of
rendering that has started but is not complete at the time of the
suspend() call, then I would expect currentTime to change after
the suspend() call but before its promise resolves.

PR #630, opened 2015-10-15
Fix #34 one more time, hopefully getting the language right.

### LINKED ISSUE #34: decodeAudioData Prose: avoid video containers that have an audio track
> Originally reported on W3C Bugzilla [ISSUE-21520](https://www.w3.org/Bugs/Public/show_bug.cgi?id=21520) Tue, 02 Apr 2013 14:41:45 GMT
> Assigned to 

Change:

"Audio file data can be in any of the formats supported by the audio element"

For:

"...can be accepted in formats containing only audio data (w/o video)"

This is to avoid the overhead of dealing with video containers that have an audio track.

billhofmann:
Question: we're still pulling to gh-pages, right?


billhofmann:
Agh.  Killing this one, it was done pre-tidy.

PR #635, opened 2015-10-16
Fix #110 by adding explicit section on Audio Signal Values.

### LINKED ISSUE #110: Sample format: what do +/-Infinity and NaN mean?
> Originally reported on W3C Bugzilla [ISSUE-17394](https://www.w3.org/Bugs/Public/show_bug.cgi?id=17394) Tue, 05 Jun 2012 12:14:04 GMT
> Reported by Michael[tm] Smith
> Assigned to 

Audio-ISSUE-19 (SamplingFormat): Define sampling format [Web Audio API]

http://www.w3.org/2011/audio/track/issues/19

Raised by: Marcus Geelnard
On product: Web Audio API

In the internal sample format the interpretation of Float32 NaN, +-Inf etc is not defined. More generally, it seems to not be defined what the sampling format of everything is (AudioBuffer, WaveShaperNode, etc).

joeberkovitz:
After reviewing the thread in #110 I feel that it's best to add a new and extremely brief section that states the nominal range of legal signal values at a destination node (note that there is more than one type of destination node, so specing this in `AudioDestinationNode` seems like an unclear option). I have opted to say that the audio rendition of values outside [-1,+1] or of NaN/+Inf/-Inf are not defined by this specification as this feels like a perfectly reasonable V1 approach.

Open to suggestions that some sort of clipping for [-1, +1] should be forced, but adding more cost to the impl doesn't seem like a great idea to me.

### DIFF
--- a/index.html
+++ b/index.html
@@ -7566,8 +7566,10 @@ Quad up-mix:
         Audio Signal Values
       </h2>
       <p>
-        The nominal range of all audio signals at a destination node of any audio graph is from -1 to +1. The audio rendition of
-        signal values outside this range, or of the values <code>NaN</code>, positive infinity or negative infinities, is undefined by this specification.
+        The nominal range of all audio signals at a destination node of any
+        audio graph is [-1, 1]. The audio rendition of signal values outside
+        this range, or of the values <code>NaN</code>, positive infinity or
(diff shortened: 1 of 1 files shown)


joeberkovitz:
Thanks Ray -- should be OK now.

PR #637, opened 2015-10-16
Fix #260 by using JavaScript for all examples.

### LINKED ISSUE #260: Use consistent language and approach to document algorithms
The following issue was raised by the W3C TAG as part of their [review of the Web Audio API](https://github.com/w3ctag/spec-reviews/blob/master/2013/07/WebAudio.md)

----------------

What language are the examples in Section 11 written in? Can it be executed?

joeberkovitz:


index.html
```diff
@@ -7472,10 +7472,14 @@ <h3 id="Spatialization-distance-effects">
           based on the panner and listener positions according to:
         </p>
         <pre class="highlight">
-  v = panner.position - listener.position
-  </pre>
-        <pre class="highlight">
-  distance = sqrt(dot(v, v))
+  function dotProduct(v1, v2) {
+    var d = 0;
+    for (var i = 0; i &lt; Math.min(v1.length, v2.length); i++)
```
padenot: Remove this HTML entity.

rtoy:
Can you fix the title?  I think it would be nice if formulas were given instead of code.  (I'm not asking to convert this code to formulas now.)

joeberkovitz:
This PR fixes #260

joeberkovitz:
This should now be good to go.

index.html
```diff
@@ -8028,25 +8029,39 @@ <h3 id="Spatialization-sound-cones">
           <a><code>PannerNode</code></a>) and the listener:
         </p>
         <pre class="highlight">
-  if (source.orientation.isZero() || ((source.coneInnerAngle == 0) &amp;& (source.coneOuterAngle == 0)))
+  function dotProduct(v1, v2) {
```
rtoy: Seems unfortunate to have to specify dotProduct twice.

rtoy:
Except for the duplicated dotProduct, looks good.  Not sure what to do about the duplication.

joeberkovitz:
I don't love the duplication either but I don't have a solution to it either so I propose we live with it for now. We could later refactor a new section with common utility methods if we feel like it but it doesn't rise to that level of issue for me.

rtoy:
Works for me.

PR #644, opened 2015-10-23
Spec the processing model for the Web Audio API

padenot:
This lays down some infrastructure we can build upon.

The cycle detection and ordering is not finished, though.

PR #645, opened 2015-10-23
Add <pre> around mathjax

padenot:
This is based on #644 and adds more `<pre>` around mathjax equations so we can format them as we want.


padenot:
(This is now being worked on in #644).

PR #646, opened 2015-10-23
Annotate synchronous sections

padenot:
This annotates all the synchronous sections for when exceptions are thrown. It's mostly mechanical changes: grep on "throw", surround with the `<span class="synchronous">`.

Again, this is based on patches in other PRs.

### DIFF
--- a/index.html
+++ b/index.html
@@ -100,6 +100,18 @@
       }
     });
     </script>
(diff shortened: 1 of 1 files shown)


padenot:
(This is now being worked on in #644).

PR #648, opened 2015-10-25
Fix #560 clarify loop behavior and time units, also playback start of…

### LINKED ISSUE #560: Clarify computedPlaybackRate with other attributes
I always have to resort to trial and error to remember how `computedPlaybackRate` interacts with the following:

1) The `loopStart` and `loopEnd` attributes of `AudioBufferSourceNode`
2) The `when`, `offset`, and `duration` parameters of `AudioBufferSourceNode#start`
3) The `when` parameter of `AudioBufferSourceNode#stop`

Assume that I'm using a `playbackRate` of `10.0` on a buffer containing a 100Hz sine wave.

From my testing, I need to multiply `duration` by `playbackRate` in order to have the audio sample last for the specified amount of time (and not end prematurely).  Whereas `when` is always specified relative to the time of the AudioContext.  `loopStart` and `loopEnd` are relative to the buffer, before playbackRate.

It would be nice to clarify all of these in the spec.

joeberkovitz:
…fset, as independent of ABSN playback rate.

### DIFF
--- a/index.html
+++ b/index.html
@@ -3090,18 +3090,30 @@ param.setValueCurveAtTime(curve, t7, t8 - t7);
           </dt>
           <dd>
             An optional value in seconds where looping should begin if the
-            <code>loop</code> attribute is true. Its default <code>value</code>
-            is 0, and it may usefully be set to any value between 0 and the
-            duration of the buffer.
+            <code>loop</code> attribute is true. Its default value is 0, and it
+            may usefully be set to any value between 0 and the duration of the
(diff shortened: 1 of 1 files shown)


joeberkovitz:
Sorry @rtoy -- which is "the one issue that you refer to which is holding this up"? I can't seem to find it in the trail anywhere.

joeberkovitz:
Thanks @rtoy, I believe we are good to go now.

PR #649, opened 2015-10-25
Fix #443: clarify garbage collection lifecycle for decodeAudioData

### LINKED ISSUE #443: OfflineAudioContext and AudioContext decoding should complete before garbage collection
The following code snippet illustrates this scenario:

function playOfflineAudio() {
    var context = new OfflineAudioContext(...);
    ...    
    context.oncomplete = finishAudioTest;
    context.startRendering(); 
}

If garbage collection runs before the complete event fires, the OfflineAudioContext will get garbage collected and never fire the event.  It would be consistent with the rest of our media event handling to require the OfflineAudioContext to complete before allowing this.

A similar issue can be encountered calling decodeAudioData() in AudioContext.

Proposed Solution: The OfflineAudioContext should remain alive until it finishes processing and fires the complete event.  Similarly, AudioContext should remain alive until it finishes decoding and fires the decoding complete event.

billhofmann:
and OfflineAudioContext startRendering

### DIFF
--- a/index.html
+++ b/index.html
@@ -1430,8 +1430,8 @@ function setupRoutingGraph() {
               rendering audio. The system shall ensure that the
               <code>OfflineAudioContext</code> is not garbage collected until
               either the promise is resolved and any callback function is 
-              called and completes, or until the <code>suspend</code>
-              function is called.
+              called and completes, or until the <code>suspend</code> function
+              is called.
             </p>
             <p>
               Although the primary method of getting the rendered audio data is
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -919,7 +919,10 @@ <h2 id="AudioContext">
             <p>
               Although the primary method of interfacing with this function is
               via its promise return value, the callback parameters are
-              provided for legacy reasons.
+              provided for legacy reasons. The system shall ensure that the
+              <code>AudioContext</code> is not garbage collected before the
```
padenot: You can just surround keywords by an empty `<a>` tag, this makes a cross-reference.

billhofmann:
in markup or in display?  I ran tidy -config tidyconf.txt index.html to produce the output.

billhofmann:
on 12/3 good to go, can merge once conflicts are fixed

PR #651, opened 2015-10-30
Fix #348 add AudioContext constructor dictionary with buffer hint

### LINKED ISSUE #348: Need to provide hints to increase buffering for power consumption reasons
It's been requested that we enable a batching mechanism to let users of the API tell us to process a batch of frames at a time (rather than the lowest latency we can support on a given system), trading off higher latency for applications that would rather minimize power (e.g. a simple background-music-playback app).

billhofmann:
add optional constructor argument for AudioContext that provides a dictionary that for now is used for buffer hint.  This is maximally forward compatible, but might be too forward compatible :)

### DIFF
--- a/index.html
+++ b/index.html
@@ -510,8 +510,8 @@ function setupRoutingGraph() {
           MediaStream sent to a remote peer.
           </li>
           <li>An <a><code>AudioWorker</code></a> interface representing a
-          factory for creating custom nodes that can process audio directly 
-          in JavaScript.
+          factory for creating custom nodes that can process audio directly in
+          JavaScript.
           </li>
           <li>An <a><code>AudioWorkerNode</code></a> interface, an
           <a><code>AudioNode</code></a> representing a node processed in an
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -695,8 +695,32 @@ <h2 id="AudioContext">
...
+          <dd>
+            Provide the lowest latency possible without glitching. This is the
+            default.
+          </dd>
+          <dt>
+            playback
+          </dt>
+          <dd>
+            Latency not important, sustained playback without interruption is
+            priority.Lowest power consumption.
```
padenot: Space after ".".

index.html
```diff
@@ -1353,6 +1377,15 @@ <h2 id="AudioContext">
             The error that occurred while decoding.
           </dd>
         </dl>
+        <dl title="dictionary AudioContextInit" class="idl">
+          <dt>
+            AudioContextPlaybackCategory bufferingHint
+          </dt>
+          <dd>
+            Provide the hint for how the user agent should trade off between
+            latency and power consumption.
```
padenot: s/latency/audio output latency/ I think, there are a lot of other latencies everywhere, best to be specific.

billhofmann:
Well, they are initialization options.  @padenot?  Tell me what makes the most sense.  We're not using a constructor argument like this elsewhere.

joeberkovitz:
You will may want to wait to merge #662 since I cleaned up a lot of things to do with constructor declarations as part of the whole de-sugaring business, including AudioContext.

billhofmann:
OK.  There’s still discussions on what this ought be called, so I’ll hold.  Shall we discuss next Thursday?

billhofmann:
Need to review post-Joe's desugaring changes

billhofmann:
only important thing is name of property, should be playbackCategory

billhofmann:
will merge after fixing conflicts

billhofmann:
Loving that merging.

PR #654, opened 2015-11-02
642 conflict on resuming context

hoch:


hoch:
padenot@

Travis CI is failing with cmake issue and I am not sure how to solve that. Could you take a look?

PR #656, opened 2015-11-02
OfflineAudioContext suspend/resume behavior

hoch:


index.html
```diff
@@ -1458,11 +1462,26 @@ <h2 id="OfflineAudioContext">
                   start rendering <code>length</code> sample-frames of audio
                   into <var>buffer</var>.
                   </li>
-                  <li>For every render quantum, check and suspend the rendering
-                  if necessary.
+                  <li>At the beginning of every render quantum, check scheduled
```
padenot: This would be better put inside the section "processing model" when my branch merges.

hoch: You mean this whole section? Or just this line?

rtoy:
I don't think we should block this on the processing model.  I think the description here should be here.  No one wants to read the entire processing model section to figure out exactly how suspend/resume works for offline contexts. But I don't oppose putting appropriate text also in the processing model section.

rtoy:
@padenot Can you clarify what you really want?  Should this wait for your processing PR to land first? I'd rather not block this PR on that.

index.html
padenot: So you can rework the prose by linking the right part of the processing algorithm, now, since it has merged.

Maybe beween 2. and 3. ?

padenot:
Can we update this so that it fits in well with the processing model prose ?

index.html
```diff
@@ -1391,10 +1391,14 @@ <h2 id="OfflineAudioContext">
...
           <code>AudioBuffer</code>.
-        </p>The OfflineAudioContext is constructed with the same arguments as
-        AudioContext.createBuffer. A NotSupportedException exception MUST be
-        thrown if any of the arguments is negative, zero, or outside its
-        nominal range.
+        </p>
+        <p>
+          The <a><code>OfflineAudioContext</code></a> is constructed with the
+          same arguments as <a href=
+          "#widl-AudioContext-createBuffer-AudioBuffer-unsigned-long-numberOfChannels-unsigned-long-length-float-sampleRate">
```
joeberkovitz: Should refer to `BaseAudioContext` here.

index.html
```diff
@@ -1458,11 +1462,26 @@ <h2 id="OfflineAudioContext">
...
                   <li>If a suspended context is resumed, continue to render the
-                  buffer.
+                  buffer. When the context is resumed,
+                    <ol>
+                      <li>Resolve the promise associated with the resume.
+                      </li>
+                      <li>Queue a task to fire an event named
+                      <code>statechange</code> at this instance.
+                      </li>
+                    </ol>
```
joeberkovitz: This is already documented in `BaseAudioContext` specs for suspension/resumption (I believe that did not exist at the time this PR was created).

joeberkovitz:
@hoch @padenot I believe the outstanding issue here is not so much the processing model, but is a conflict with already existing language for the `BaseAudioContext` methods `suspend()` and `resume()`. 

This PR's definition suggests that the rendering thread checks the suspension state of the AudioContext at the start of every render quantum, and that the suspension Promise resolves and statechange event gets dispatched only when the suspension is acted on in a rendering sense. However the definitions in BaseAudioContext do not involve the rendering thread, only the control thread. Thus they imply that a `statechange` event could be dispatched even as the audio thread is still working on the end of the final quantum prior to suspension.

I've filed a new issue #919 to capture this question.

joeberkovitz:
I think this is now good to go given that #994 was merged

hoch:
I am currently working on this in conjunction with #1060.

Issue #650, opened 2015-10-29
Chaining AudioNode.connect() and AudioParam's automation methods


hoch:
This is related to issue #258 and PR 44ebce6044882da08c41f30a835b3564c922a407.

Currently what we have in the spec is:
```js
AudioNode connect (AudioNode destination, ...);
AudioParam connect (AudioParam destination, ...);
```

I believe we've discussed the first method in the meeting, but have we agreed on the second one? Actually second one looks a bit weird.

```js
osc.connect(filter.frequency)
  .linearRampToValue(880, context.currentTime + 2.0);
```

Since we cannot chain the automation method, the chaining fun ends here. Also it is a bit weird to see the change of flow from `connect() -> linearRampToValue()`. Personally, I would like to have connecting AudioNode to AudioParam not return anything.

In addition to that, chaining AudioParam methods would be a nice addition.

```js
var now = context.currentTime;
lowpass.frequency.setValueAtTime(220, now)
  .exponentialRampToValueAtTime(880, now + 1.0)
  .linearRampToValueAtTime(440, now + 2.0);
```

It saves a bit of typing and improves the flow of code. This is kind of what I have been doing with [my library](https://github.com/hoch/spiral/blob/master/examples/audioparam-automation.html) and I think it is quite convenient.

padenot:
Agreed. Similarly, it's very easy to do and looks nice.

hoch:
So here's my proposal:

```webidl
partial interface AudioNode : EventTarget {
    AudioNode  connect (AudioNode destination, optional unsigned long output = 0, 
                        optional unsigned long input = 0);
    void connect (AudioParam destination, optional unsigned long output = 0);
}
```
Connecting to an `AudioNode` returns the destination node whereas connecting to an `AudioParam` returns nothing.

```webidl
partial interface AudioParam {
    AudioParam setValueAtTime (float value, double startTime);
    AudioParam linearRampToValueAtTime (float value, double endTime);
    AudioParam exponentialRampToValueAtTime (float value, double endTime);
    AudioParam setTargetAtTime (float target, double startTime, float timeConstant);
    AudioParam setValueCurveAtTime (Float32Array values, double startTime, 
                                    double duration);
    AudioParam cancelScheduledValues (double startTime);
};
```
Unlike `AudioNode::connect`, these methods return the `AudioParam` object itself.

mrkishi:
[...]

PR #660, opened 2015-11-10
Support chaining on AudioNode.connect() and AudioParam automation methods

hoch:
- Revert: AudioNode.connect(AudioParam) does not return anything. (see #650)
- Also changed the weird location of 'parameters' in the method description.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1745,20 +1745,21 @@ function setupRoutingGraph() {
 </pre>
           </dd>
           <dt>
-            AudioParam connect()
+            void connect()
           </dt>
           <dd>
             Connects the <a><code>AudioNode</code></a> to an
             <a><code>AudioParam</code></a>, controlling the parameter value
-            with an audio-rate signal. The <code>destination</code> is also
-            returned by this method in order to enable easy chaining.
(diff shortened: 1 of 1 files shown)


hoch:
I reverted back to the current HEAD and re-applied the changes without redundant edit. (Sorry that we lost our comments in there.)

index.html
```diff
@@ -2388,9 +2389,13 @@ <h2 id="AudioParam">
               "#widl-AudioParam-exponentialRampToValueAtTime-void-float-value-double-endTime">
               exponentialRampToValueAtTime</a></code>, respectively.
             </p>
+            <p>
+              This method returns the current <a>AudioParam</a> object for easy
```
rtoy: I didn't find anything about AudioNode connect() allowing easy chaining in the spec (did I miss it?), so I don't see why we need to say anything about easy chaining.  I think it's fairly obvious you can chain them together now.

hoch: Then should I remove all the instances of this paragraph as whole?

rtoy: No, I think you need to say what the actual returned object is.  I would just remove the easy chaining part.

PR #677, opened 2015-12-09
Fix #12 by adding attributes for performance timestamps and output la…

### LINKED ISSUE #12: Need a way to determine AudioContext time of currently audible signal
> Originally reported on W3C Bugzilla [ISSUE-20698](https://www.w3.org/Bugs/Public/show_bug.cgi?id=20698) Thu, 17 Jan 2013 14:15:09 GMT
> Reported by Joe Berkovitz / NF
> Assigned to 

Use case:

If one needs to display a visual cursor in relationship to some onscreen representation of an audio timeline (e.g. a cursor on top of music notation or DAW clips) then knowing the real time coordinates for what is coming out of the speakers is essential.

However on any given implementation an AudioContext's currentTime may report a time that is somewhat ahead of the time of the actual audio signal emerging from the device, by a fixed amount.  If a sound is scheduled (even very far in advance) to be played at time T, the sound will actually be played when AudioContext.currentTime = T + L where L is a fixed number.

On Jan 16, 2013, at 2:05 PM cwilso@google.com wrote:

It's problematic to incorporate scheduling other real-time events (even knowing precisely "what time it is" from the drawing function) without a better understanding of the latency.

The idea we reached (I think Chris proposed it, but I can't honestly remember) was to have a performance.now()-reference clock time on AudioContext that would tell you when the AudioContext.currentTime was taken (or when that time will occur, if it's in the future); that would allow you to synchronize the two clocks.  The more I've thought about it, the more I quite like this approach - having something like AudioContext.currentSystemTime in window.performance.now()-reference.

On Jan 16, 2013, at 3:18 PM, Chris Rogers <crogers@google.com> wrote:

the general idea is that the underlying different platforms/OSs can have very different latency characteristics, so I think you're looking for a way to query the system to know what it is.  I think that something like AudioContext.presentationLatency is what we're looking for.  Presentation latency is the time difference between when you tell an event to happen and the actual time when you hear it.  So, for example, with source.start(0), you would hope to hear the sound right now, but in reality will hear it with some (hopefully) small delay.  One example where this could be useful is if you're trying to synchronize a visual "playhead" to the actual audio being scheduled...

I believe the goal for any implementation should be to achieve as low a latency as possible, one which is on-par with desktop/native audio software on the same OS/hardware that the browser is run on.  That said, as with other aspects of the web platform (page rendering speed, cache behavior, etc.) performance is something which is tuned (and hopefully improved) over time for each browser implementation and OS.

pozdnyakov:
…tency

This patch adds BaseAudioContext.currentPerformanceTime and AudioDestinationNode.latency
attributes so that the audible time of the frame corresponding to 'BaseAudioContext.currentTime'
can be estimated as 'time = context.performanceTimeStamp + destNode.latency'.

### DIFF
--- a/index.html
+++ b/index.html
@@ -799,6 +799,58 @@ function setupRoutingGraph() {
             </p>
           </dd>
           <dt>
+            readonly attribute DOMHighResTimeStamp currentPerformanceTime
+          </dt>
+          <dd>
+            <p>
+              The <code>DOMHighResTimeStamp</code> representation of <a href=
+              "#widl-BaseAudioContext-currentTime"><code>currentTime</code></a>.
+            </p>
+            <p>
+              The <a href=
+              "#widl-BaseAudioContext-currentPerformanceTime"><code>currentPerformanceTime</code></a>
+              property contains the <code>DOMHighResTimeStamp</code> (described
+              in [[!hr-time-2]]) representing the time when the <a href=
+              "#widl-BaseAudioContext-currentTime"><code>currentTime</code></a>
+              was last time updated and incremented internally by the duration
+              in milliseconds of the audio block most recently processed by the
+              context's rendering graph, so that <code>currentTime</code> and
+              <code>currentPerformanceTime</code> values always correspond to
+              each other. Naturally, the <code>currentPerformanceTime</code>
+              attribute value is updated synchronously with the
+              <code>currentTime</code> attribute value increments. If the
+              context's rendering graph has not yet processed a block of audio
+              (i.e. <code>currentTime</code> has never been updated), then
+              <a href=
+              "#widl-BaseAudioContext-currentPerformanceTime"><code>currentPerformanceTime</code></a>
+              has a value of zero.
+            </p>
+            <p class="note">
+              The audio block duration (which forms the
+              <code>currentPerformanceTime</code> shift) depends on the system:
+              on repeated calls of the Web Audio API the reference is always
+              the first call in a row. For example if the system processes
+              audio blocks of 512 sample-frames (calling the Web Audio API 4
+              times in a row) the <code>currentPerformanceTime</code> value
+              will be shifted by: <em>duration = 1000 * 4 * 128 / 44100Hz =
+              11.6ms</em>. If <code>currentTime</code> is updated after each
+              call in a row, <code>currentPerformanceTime</code> should still
+              read the time from the reference call, and add the duration of
+              the blocks processed so far in the row, so as to preserve the
+              logical linear relationship between these two attributes. The
+              implementation should adjust the value of the
+              <code>currentPerformanceTime</code> shift according to the
+              current system block duration, if it varies over time.
+            </p>
+            <p>
+              The value of <a href=
+              "#widl-BaseAudioContext-currentPerformanceTime"><code>currentPerformanceTime</code></a>
+              property is in same units and starting from the same origin as
+              values returned by <code>performance.now()</code> (described in
+              [[!hr-time-2]]).
+            </p>
+          </dd>
+          <dt>
             readonly attribute AudioListener listener
           </dt>
           <dd>
@@ -2303,6 +2355,23 @@ function setupRoutingGraph() {
               "#widl-AudioNode-channelCount"><code>channelCount</code></a> is
               determined when the offline context is created and this value may
               not be changed.
+            </p>
+          </dd>
+          <dt>
+            readonly attribute double latency
+          </dt>
+          <dd>
+            <p>
+              The user agent's best estimation in seconds of audio output
+              latency, i.e. the delay between the moment the implementation
+              sends a buffer to the system and the audible sound from the audio
+              output (speaker or connector).
+            </p>
+            <p>
+              If user agent is not able to estimate the audio output latency
+              then <a href=
+              "#widl-AudioDestinationNode-latency"><code>latency</code></a> has
+              a value of zero.
             </p>
           </dd>
         </dl>


pozdnyakov:
@joeberkovitz @jipodine could you pls have a look?

jipodine:
About `currentPerformanceTime`: as `currentTime` represents the time of the next audio block, and `currentPerformanceTime` the last time-stamp `currentTime` was updated, there is a difference of one audio block. Even if we do not exactly know the time of the next update, it would be more straightforward to already shift `currentPerformanceTime` by the audio block duration, so as it matches `currentTime` more closely. I am also concerned by the fact that a lot of systems do not regularly advance `currentTime` by 128 samples (as required by the standard but depending on the underlying system), and that the actual block duration is not exposed to the user.

About the latency: it should be defined somehow. Maybe as the delay between the processing of the audio when `currentTime` reaches the scheduled time, and the audible sound from the audio output (speaker or connector). There is already some definitions in [section 8.1 Latency](https://webaudio.github.io/web-audio-api/#latency) that could clarify the different latencies.

rtoy:
On Tue, Dec 15, 2015 at 10:50 AM, Jean-Philippe Lambert <
notifications@github.com> wrote:

> About currentPerformanceTime: as currentTime represents the time of the
> next audio block, and currentPerformanceTime the last time-stamp
> currentTime was updated, there is a difference of one audio block. Even
> if we do not exactly know the time of the next update, it would be more
> straightforward to already shift currentPerformanceTime by the audio
> block duration, so as it matches currentTime more closely. I am also
> concerned by the fact that a lot of systems do not regularly advance
> currentTime by 128 samples (as required by the standard but depending on
> the underlying system), and that the actual block duration is not exposed
> to the user.
>

​What do you mean that currentTime does not advance by 128 samples?  What
block duration are you talking about here?​


> About the latency: it should be defined somehow. Maybe as the delay
> between the processing of the audio when currentTime reaches the
> scheduled time, and the audible sound from the audio output (speaker or
> connector). There is already some definitions in section 8.1 Latency
> <https://webaudio.github.io/web-audio-api/#latency> that could clarify
> the different latencies.
>
> —
> <https://github.com/WebAudio/web-audio-api/pull/677#issuecomment-164853289>
> .
>



-- 
Ray

jipodine:
On some devices, when repeatedly reading the `currentTime`, the difference between the values is never 128 samples, but a bigger multiple of it. It can be 256, 512, or even 4096 samples. (You can try for yourself [here](http://jipodine.github.io/web-audio-clock/).) My guess is that the Web Audio advance follows the one of the underlying audio system: if the system audio buffer is 1024 samples, then it processes 1024 samples at once, possibly calling the Web Audio 8 times in a row.

Then is will be hard to make a correspondence between `currentPerformanceTime` and `currentTime`, because they do not refer to the same time. My proposal was about shifting `currentPerformanceTime` by the actual advance of `currentTime`, but I think it would be cleaner to extend the time-stamp with 2 values, when the Web Audio time advances:
- `previousTime` as the last `currentTime` 
- `previousPerformanceTime`as the last `performance.now`

This way it is clear that we refer to a previous time, and that the 2 value match. Then it is easy to estimate a delay and a skew.

rtoy:
On Wed, Dec 16, 2015 at 2:42 AM, Jean-Philippe Lambert <
notifications@github.com> wrote:

> On some devices, when repeatedly reading the currentTime, the difference
> between the values is never 128 samples, but a bigger multiple of it. It
> can be 256, 512, or even 4096 samples. (You can try for yourself here
> <http://jipodine.github.io/web-audio-clock/>.) My guess is that the Web
> Audio advance follows the one of the underlying audio system: if the system
> audio buffer is 1024 samples, then it processes 1024 samples at once,
> possibly calling the Web Audio 8 times in a row.
>

​Yes,  with Chrome, this is basically what happens.  However, currentTime
always advances by 128 samples, but it may run 8 of them back-to-back very
quickly. (Depends on the OS.  512 for linux, 128 for OSX, 10ms for
windows.) This is an result of some audio backends not being able to
reliably call back every 128 samples on a consistent basis.  Without this,
some processing blocks don't have enough time to finish before the next
callback, glitching the audio.  OSX is the only system where Chrome can do
128 frames at a time.​


> Then is will be hard to make a correspondence between
> currentPerformanceTime and currentTime, because they do not refer to the
> same time. My proposal was about shifting currentPerformanceTime by the
> actual advance of currentTime, but I think it would be cleaner to extend
> the time-stamp with 2 values, when the Web Audio time advances:
>
>    - previousTime as the last currentTime
>    - previousPerformanceTimeas the last performance.now
>
> This way it is clear that we refer to a previous time, and that the 2
> value match. Then it is easy to estimate a delay and a skew.
>
> —
> <https://github.com/WebAudio/web-audio-api/pull/677#issuecomment-165062817>
> .
>



-- 
Ray

pozdnyakov:
>Then it will be hard to make a correspondence between currentPerformanceTime and currentTime, because they do not refer to the same time. My proposal was about shifting currentPerformanceTime by the actual advance of currentTime, but I think it would be cleaner to extend the time-stamp with 2 values, when the Web Audio time advances:

  >* previousTime as the last currentTime
  >* previousPerformanceTime as the last performance.now

@jipodine Thanks for your comments! Your first proposal looks more preferable to me. I'm a bit concerned about exposing multiple timestamp properties. Wouldn't it be confusing for the API users (what is `previousTime` vs `currentTime` and considering that the `currentTime` is not actually "current" but something in advance)?.. 
On the other hand a single newly-added property bound to a "well-known" `currentTime` would be easier to understand. Besides, on browser side we know exactly how much frames the `currentTime` is shifted, so there should not be technical problems to adjust `currentPerformanceTime` to `currentTime`. WDYT?

jipodine:
@pozdnyakov Yes, this was the idea of my first proposal. I also think is is better integrated. Any way `currentPerformanceTime` is an estimation, so better to stay with a simple API (with `currentPerformanceTime` internally shifted by the browser block duration to match `currentTime`).

By the way, some definitions should be clarified:
- block duration: block size * sample rate
- browser block size: number of samples processed together by the browser

pozdnyakov:
@jipodine  I've updated the PR based on your feedback. Could you have a look?

jipodine:
It looks good. Maybe an example would be good for clarification (regarding @rtoy comments about the audio block), or a sketch.
> If the system processes audio blocks of 512 samples at 4410 Hz, (calling the Web Audio API 4 times in a row) the system audio block duration is 1000 * 1024 / 44100 = 11.609977324263038 ms

pozdnyakov:
> It looks good. Maybe an example would be good for clarification (regarding @rtoy comments about the audio block), or a sketch.

Added a dedicated paragraph, thanks.

jipodine:
One last remark: On repeated calls of the Web Audio API, it should be clear that the reference is the *first* call in a row. In the example, the shift is 4 * 128 samples from there.

pozdnyakov:
> One last remark: On repeated calls of the Web Audio API, it should be clear that the reference is the first call in a row. In the example, the shift is 4 * 128 samples from there.

Updated, thanks

rtoy:
It's, unfortunately, significantly more complicated than that.  Chrome on Windows with WASAPI uses 10 ms buffers (441 frames at 44.1 kHz).  There's a fifo between webaudio and the driver so that most of the time 4*128 frames are processed, but occasionally only 3*128 frames are processed because there are some samples left over from previous calls.

There's going to be some jitter, even if you make it the first of a set of calls.  It should, however, all average out.

pozdnyakov:
@rtoy but on the implementation side we are aware of the exact quantity of the frames to process each time, so `currentPerformanceTime` can be shifted considering that, right? Do you think it's enough to say that it's the implementation duty to provide the most accurate `currentPerformanceTime` values on a given platform and adjustments are possible?

jipodine:
@rtoy As long as the correspondence between `currentTime` and `currentPerformanceTime` is good (as close as possible to the reality), we are fine: we do not need regular sampling, but a carefully adjusted advance for `currentPerformanceTime`. As you said, it might not be trivial, so this example should be included in the document.

pozdnyakov:
@rtoy @jipodine I've just added a paragraph saying that the implementation is allowed to adjust the `currentPerformanceTime` shift if needed. Do you think it is ok?

jipodine:
> The implementation is allowed to adjust the value of the currentPerformanceTime

@pozdnyakov I would say that the implementation "shall" adjust (instead of "is allowed to" adjust) to the current processing block duration.

And maybe (maybe not if you think it is too much) provide @rtoy example of a variable block size over time, or simply add something like "according to the current system block duration, if it varies over time".

pozdnyakov:
> I would say that the implementation "shall" adjust (instead of "is allowed to" adjust) to the current processing block duration.

Done.
> And maybe (maybe not if you think it is too much) provide @rtoy example of a variable block size over time, or simply add something like "according to the current system block duration, if it varies over time".

I think the @rtoy example is great indeed but I would keep the description generic, so I used the quote that you've proposed. Thanks!

pozdnyakov:
@joeberkovitz @padenot @cwilso could you provide your feedback to the PR? If it is accepted I would be interested in implementing the proposed attributes in Chromium.

index.html
```diff
@@ -2303,6 +2355,23 @@ <h2 id="AudioDestinationNode">
               not be changed.
             </p>
           </dd>
+          <dt>
+            readonly attribute double latency
+          </dt>
+          <dd>
+            <p>
+              The UA's best estimation in seconds of audio output latency, i.e.
```
padenot: Please spell UA in its entirety.

index.html
```diff
@@ -797,6 +797,58 @@ <h2 id="BaseAudioContext">
...
+            <p>
+              For blocks of 128 sample-frames audio block duration is equal to
+              roughly 3ms at a sample-rate of 44.1KHz, however the used audio
+              block duration depends on the system: on repeated calls of the
+              Web Audio API the reference is always the first call in a row.
+              For example if the system processes audio blocks of 512
+              sample-frames (calling the Web Audio API 4 times in a row) the
+              <code>currentPerformanceTime</code> value will be shifted by:
+            </p>
+            <pre>duration = 1000 * 4 * 128 / 44100Hz = 11.6ms</pre>
```
padenot: This should be in an informative note or example.

index.html
```diff
@@ -797,6 +797,58 @@ <h2 id="BaseAudioContext">
...
+            <p>
+              The <code>DOMHighResTimeStamp</code> representation of <a href=
+              "#widl-BaseAudioContext-currentTime"><code>currentTime</code></a>.
+            </p>
+            <p>
+              The <a href=
+              "#widl-BaseAudioContext-currentPerformanceTime"><code>currentPerformanceTime</code></a>
+              property contains the <code>DOMHighResTimeStamp</code>
+              representing the time in milliseconds from the <code>time
+              origin</code> (described in [[!hr-time-2]]) to the moment when
```
padenot: `DOMHighResTimeStamp` is a time in microsecond, with a granularity of 5 microseconds.

index.html
```diff
@@ -797,6 +797,58 @@ <h2 id="BaseAudioContext">
...
+              the <code>currentPerformanceTime</code> shift according to the
+              current system block duration, if it varies over time.
+            </p>
+            <p>
+              The value of <a href=
+              "#widl-BaseAudioContext-currentPerformanceTime"><code>currentPerformanceTime</code></a>
+              property is in same units and starting from the same origin as
+              values returned by <code>performance.now()</code> (described in
+              [[!hr-time-2]]) and it is guaranteed to increase monotonically
+              and not to be a subject to system clock skew or adjustments.
```
padenot: Again, no need to repeat the properties of `performance.now()`.

index.html
```diff
@@ -797,6 +797,58 @@ <h2 id="BaseAudioContext">
...
+              The <a href=
+              "#widl-BaseAudioContext-currentPerformanceTime"><code>currentPerformanceTime</code></a>
+              property contains the <code>DOMHighResTimeStamp</code>
+              representing the time in milliseconds from the <code>time
+              origin</code> (described in [[!hr-time-2]]) to the moment when
+              the <a href=
+              "#widl-BaseAudioContext-currentTime"><code>currentTime</code></a>
+              was last time updated and incremented internally by the duration
+              in milliseconds of the audio block processed by the system, so
+              that <code>currentTime</code> and
```
padenot: Right now, we don't know how `currentTime` is incremented, so it's a bit of a problem. More over, UAs don't agree on this yet. Gecko updates `currentTime` during stable state, other implementations fetch it every time.

index.html
```diff
@@ -2303,6 +2355,23 @@ <h2 id="AudioDestinationNode">
...
           </dd>
+          <dt>
+            readonly attribute double latency
+          </dt>
+          <dd>
+            <p>
+              The UA's best estimation in seconds of audio output latency, i.e.
+              the delay between the moment the implementation sends a buffer to
+              the system and the audible sound from the audio output (speaker
+              or connector).
```
padenot: I think we have some language in the "Processing Model" section of the spec to refer to the rendering of audio, but I can't remember.

index.html
pozdnyakov: Wouldn't it be enough to just say that `currentTime` and `currentPerformanceTime` should be always updated simultaneously (whatever moment `currentTime` is incremented)?

padenot: That would work indeed.

joeberkovitz: I think that the currentTime update mechanics were in fact fully specified in https://github.com/WebAudio/web-audio-api/pull/625

index.html
```diff
@@ -797,6 +797,58 @@ <h2 id="BaseAudioContext">
...
+              sample-frames (calling the Web Audio API 4 times in a row) the
+              <code>currentPerformanceTime</code> value will be shifted by:
+            </p>
+            <pre>duration = 1000 * 4 * 128 / 44100Hz = 11.6ms</pre>
+            <p>
+              In order to keep the correspondence between
+              <code>currentTime</code> and <code>currentPerformanceTime</code>
+              attributes values the implementation should adjust the value of
+              the <code>currentPerformanceTime</code> shift according to the
+              current system block duration, if it varies over time.
```
joeberkovitz: I don't understand what is being said here in "adjust the value of the... shift... if it varies over time". Why do we not simply define the way that `currentPerformanceTime` behaves in terms of the increments that are applied during the processing of each block? The behavior that I _think_ you are describing is just a natural consequence of the UA happening to process a bunch of blocks "in a row", but there is no need to describe any special behavior that the implementation should take like "adjusting a shift" -- if the UA processes 4 blocks (or any number of blocks) of 128 frames, then it should behave as though `currentTime` and `currentPerformanceTime` were incremented separately for each block.

pozdnyakov:
@joeberkovitz @padenot I have updated the PR according to your comments, thanks.

jipodine:
@joeberkovitz No: the net effect is not the same. This is what this discussion is about. While the `currentTime` increments by the audio block duration on each call, the `currentPerformanceTime` must read from `performance.now` and anticipate what the next time will be.

The following example assumes these simplifications: the browser processes 4 blocks of 128 samples in a row, and it needs 64 samples of real time (seen as `performance.now`) to process each block. `currentTime` is drawn in samples, as `performance.now`. 
currentperformancetime

pozdnyakov:
@jipodine thanks for the nice diagram, I've put back the dedicated paragraph as a note.

joeberkovitz:
@jipodine I agree with the diagram 100%, I see that I was a bit behind on the meaning of  `currentPerformanceTime` as it evolved during this PR. I see that it's defined as a prediction of when the next block render will occur.

jipodine:
@joeberkovitz Yes, it is getting clearer as the discussion goes into the details. On the other hand, I also agree that the description should be generic enough.

I did not follow the discussions about the moment `currentTime` updates, as raised by @padenot, but I guess the implementers will find a way to update `currentPerformanceTime` according to their implementations, whenever `currentTime` updates (so as `currentTime` and `currentPerformanceTime` always correspond to each other, as specified).

Contrary to the previous example, if `currentTime` is updated after each call in a row, `currentPerformanceTime` should still read the time from the first call, and add the duration of the blocks processed so far, so as to preserve the logical linear relationship between the two times:
currentperformancetimeupdate

pozdnyakov:
> if currentTime is updated after each call in a row, currentPerformanceTime should still read the time from the first call, and add the duration of the blocks processed so far, so as to preserve the logical linear relationship between the two times

Think it is an important detail that needs to be included to the attribute description

pozdnyakov:
>> if currentTime is updated after each call in a row, currentPerformanceTime should still read the time from the first call, and add the duration of the blocks processed so far, so as to preserve the logical linear relationship between the two times

>Think it is an important detail that needs to be included to the attribute description

Done.  @jipodine do you think the attributes description is comprehensive enough? thanks

jipodine:
@pozdnyakov I looks precise enough to me: we will see what the implementers say.

pozdnyakov:
@joeberkovitz @padenot does it look OK to you?

joeberkovitz:
@pozdnyakov I think the current description is a bit confusing and doesn't quite convey the sense of the discussion so far.

Much of the complexity in this discussion has emerged because you and @jipodine decided not to go with an `outputPerformanceTime` attribute that factored in latency. You've diverged a lot from the solution that the WG blessed in #12. That approach would have indicated the performance time when the next-rendered sample would emerge from the audio hardware. Instead you have been trying to work from the performance time at which a block is rendered, and add a `destination.latency` to it which presumably is a fairly stable number. But that doesn't really work very well when blocks are rendered at irregular intervals (e.g. when blocks are batched as per the example in the graph above).

The current text of the description, as I read it, uses the time at which the _previous_ block was rendered, not the time at which the _next_ block will be rendered. That seems to be at odds with what @jipodine was asking for.

I will ask the WG to take this pull request up for discussion on tomorrow's call so that we can try to settle this question.

pozdnyakov:
> Much of the complexity in this discussion has emerged because you and @jipodine decided not to go with an outputPerformanceTime attribute that factored in latency. You've diverged a lot from the solution that the WG blessed in #12. That approach would have indicated the performance time when the next-rendered sample would emerge from the audio hardware. Instead you have been trying to work from the performance time at which a block is rendered, and add a destination.latency to it which presumably is a fairly stable number. 

I thought this approach had been agreed right before the PR was created ( https://github.com/WebAudio/web-audio-api/issues/12#issuecomment-161003078 , https://github.com/WebAudio/web-audio-api/issues/12#issuecomment-161986268  and https://github.com/WebAudio/web-audio-api/issues/12#issuecomment-161988584) , or maybe I got it wrong?
The `currentTime` attribute does not consider the latency so wouldn't it be natural for `currentPerformanceTime` (which is `currentTime` representation in terms of `Performance.now()` ) to do the same?

 > But that doesn't really work very well when blocks are rendered at irregular intervals (e.g. when blocks are batched as per the example in the graph above).

Why the `outputPerformanceTime` (with factored in output latency) would be a better solution in this case?

> The current text of the description, as I read it, uses the time at which the previous block was rendered, not the time at which the next block will be rendered. That seems to be at odds with what @jipodine was asking for.

The *next* block is meant there actually (and apparently the description should be improved in order to make it more explicit) and the time is calculated as `timestamp from most recent update + shift`, so that `currentTime` and  `currentPerformanceTime` always correspond to each other and refer to the same position.

pozdnyakov:
> I will ask the WG to take this pull request up for discussion on tomorrow's call so that we can try to settle this question.

thanks!

joeberkovitz:
Let me try to explain the problem as I see it, to set the stage for further discussion. I think the goal is to allow the user to employ both some kind of performance time attribute and a latency number to determine when audio will be heard. I have concerns that the scheme in this PR may not be able to do that reliably.

Let's consider the situation @jipodine described in which blocks are not rendered at regular intervals -- for instance, assume that 4 128-sample blocks are rendered consecutively in a batch with a very short time interval in between.

Now let's say that we're about to start rendering a batch, and that at the moment, `currentPerformanceTime + latency` does indeed accurately describe the time at which the next block's output will be heard. So things are good right now, in terms of the goal of predicting an actual output time.

Now process the first block in the batch. `currentTime` will advance by the duration of 128 samples. What should `currentPerformanceTime` advance by?

If you advance it by any amount other than the exactly same increment (i.e. the duration of 128 samples, although in milliseconds) you will destroy the ability to use `currentPerformanceTime + latency` to predict the actual output time of the start of the next block, assuming that latency stays roughly stable in any short time interval.

However, if you *do* advance it by the same increment (which as @jipodine showed in the graph is a smaller increment than 128 samples), then `currentPerformanceTime` does not accurately describe the time at which the next render will take place.

This is why I find `currentPerformanceTime` ultimately not very useful as an ingredient in latency predictor by itself. It's fine to have it, but I don't think it releases us from the need for `outputPerformanceTime`. This is what I think @srikumars meant in his comment https://github.com/WebAudio/web-audio-api/issues/12#issuecomment-166947246

jipodine:
@joeberkovitz As far as I understand, `outputPerformanceTime` =  `currentPerformanceTime` + `destination.latency` so the complexity is equivalent. The complexity comes from the prediction of the `performance` time for the next block, in order to match `currentTime`.  (And also from the details of the implementation, that are important for the precision, but may be confusing.) But to the user, it should always be that `currentTime` always matches `currentPerformanceTime`. And, while they advance by steps, they logically represent a linear time.

Any way, each time `currentTime` updates, `currentPerformanceTime` or `outputPerformanceTime` must update, in order to keep matching each other.
- if `currentTime` updates only once after every block in a row are processed, this is the first graph: `currentPerformanceTime` is the `performance` time of the first block, shifted by the duration of the row.
- if `currentTime` updates after each block in a row, then it is the second graph: `currentPerformanceTime` is the `performance` time of the first block, shifted by the duration of the block processed so far. 
This second case is generic enough to be applied also for the first case. (It should be enough in the description.)

> This is why I find currentPerformanceTime ultimately not very useful as an ingredient in latency predictor by itself. It's fine to have it, but I don't think it releases us from the need for outputPerformanceTime. This is what I think @srikumars meant in his comment #12 (comment)

Decoupling the latency and the mapping to `performance.now` is way to allow for latency compensation using only `currentTime`, which is the only sample-accurate time available, yet. Like `currentTime`, it also allows to schedule events regardless of the actual latency. (I.e. without taking care of anticipating the time more than the latency, to stay in the future.) It depends on the use-case, and it is anyway easy to work the other way. If you think that `outputPerformanceTime` is easier to understand or matches more use-cases, it is perfectly equivalent, as long as there is a simple way to get the latency.

joeberkovitz:
@jipodine: Are you saying that `currentPerformanceTime` always gives the output performance time for the sample at `currentTime`, once latency is added to it?  If so, I think it may be easiest to actually define it that way in the spec (whether we actually have an `outputPerformanceTime` attribute or not).

jipodine:
@joeberkovitz Yes, it is a clear definition. And true as well in the `AudioContext` time reference, as `currentTime` gives the output time once latency is added to it. This could be added to the definition of the latency also.

(Would it be better to define the latency with the `currentTime`, instead of `currentPerformanceTime`? On devices with a high-latency, there will be a measurable difference: around 4 samples per second of latency due to the drift of `performance.now` and `currentTime`. But the latency is not exact any way, and `currentPerformanceTime` is an estimation, so the approximation that the latency is the same is fine.)

joeberkovitz:
One more question then: what is the value of knowing `currentPerformanceTime` *without* the latency added to it? What useful decisions or actions can a developer make with `currentPerformanceTime` alone, when they're not interested in the future emergence of the sound signal from some audio hardware? I think that is really the issue that should decide whether we go with `currentPerformanceTime` or `outputPerformanceTime`.

It's already been successfully argued (in my opinion anyway) that adding latency to `currentTime` doesn't work because `currentTime` advances in a jumpy and unpredictable fashion (whenever an actual block render occurs), with potential jitter that could well exceed 4 samples per second in some high-latency scenarios. Also, due to the batching scenario you have described so well, the time derivative of `currentTime` is anything but constant.

joeberkovitz:
I wanted to also point out that the expression `currentTime - latency` gives us a good estimate of the `AudioContext` time for the signal currently being heard by the listener. So `latency` is innately useful -- I'm not debating that point.

pozdnyakov:
> One more question then: what is the value of knowing currentPerformanceTime without the latency added to it?

It could be used as an estimation of the earliest possible time when any change scheduled in the current state might take effect, so, for example, at this time the app UI could be updated accordingly to audio graph modifications. Maybe @jipodine can provide better examples..
However I can also confirm that if people think `outputPerformanceTime` +  `latency` is more convenient API than `currentPerformanceTime` + `latency` I'm totally fine with that.

joeberkovitz:
> It could be used as an estimation of the earliest possible time when any change scheduled in the current state might take effect, so, for example, at this time the app UI could be updated accordingly to audio graph modifications.

That's useful, thanks! However, this use case assumes the older, simpler definition of `currentPerformanceTime` as the performance time when the next block render takes place. This definition, as we've agreed, cannot be used to predict the performance time of the output of the next rendered sample -- but as you point out, this definition does allow us to determine the performance time at which changes to the graph will be applied.

Perhaps this leads to another useful option, have _both_ attributes. `current...` tells us when the next block render takes place, and `output...` tells us when the next block will be heard.

> However I can also confirm that if people think outputPerformanceTime + latency is more convenient API than currentPerformanceTime + latency I'm totally fine with that.

Just to clarify, I think the choice for predicting the time at which the next-rendered sample will be heard is between `outputPerformanceTime` (which already accounts for latency) and `currentPerformanceTime + latency`. I am trying to come up with a definition of `currentPerformanceTime` that is helpful, to see how we can defend that option.

joeberkovitz:
From the WG call today, I have an action item to write up a revised proposal for how we fix this issue based on clear use cases. The group wants to revisit the approach for #12 and feels that we should review that proposal in hand before working further on this PR. Neither the current PR (`currentPerformanceTime`, `latency`) nor the previously recorded approach to #12 (`outputPerformanceTime`) feel adequate to the group.

I'll be adding my further thoughts to the thread on #12 for WG review and general comment.

jipodine:
> It's already been successfully argued (in my opinion anyway) that adding latency to currentTime doesn't work because currentTime advances in a jumpy and unpredictable fashion (whenever an actual block render occurs), with potential jitter that could well exceed 4 samples per second in some high-latency scenarios. Also, due to the batching scenario you have described so well, the time derivative of currentTime is anything but constant.

This is the exact same issue with `outputPerformanceTime`. 

See the second graph: `currentPerformanceTime` (with the arrows), is in linear relationship with `currentTime` if the system processes blocks in a row and then waits until it needs more. It is a simple heuristic, but to improve the accuracy, we then need a linear regression.

> what is the value of knowing `currentPerformanceTime` without the latency added to it?

I would say that its main value lies in its correspondence to `currentTime`, so we can use both to estimate the mapping of `performance.now` to or from  `currentTime`. And, as @pozdnyakov said, it is a way to schedule events as fast as possible (without taking the latency into account, like `currentTime`).

I totally agree that `outputPerformanceTime` might be more general (and compatible with the time of the audio element, as far as I understand it), and one subtraction away from `currentPerformanceTime`.

joeberkovitz:
@jipodine to help with the forthcoming revised proposal can you please give an example -- a real-world use case -- of how one might use a function that as you put it, "maps from `performance.now()` to or from `currentTime`". I think the WG is having trouble understanding how `currentTime` (which only indicates what the audio engine is about to render and progresses in a jumpy, quantized fashion) relates to real time.

It may help to note that _context time_ is the term the WG uses to describe the timeline along which `currentTime` moves, but they are not the same thing. We definitely see a relationship between context time and real time, as regards the context time of the signal that is currently being played. However, that is not the same as `currentTime`.

jipodine:
Yes, I am talking about `currentTime` that updates at irregular intervals. We chose to base our distributed rendering on it because:
- `performance.now` is not available on all platforms
- we ultimately need to use it for audio rendering, so we avoid intermediate estimations, and we keep the audio time as the reference time
- with enough samples, its accumulated precision is fine (not so far from context time)

Here is the code that we use for synchronisation, that maps `currentTime` to an other (high-resolution) time. It is simple, and could be even simpler without the ping-pong. https://github.com/collective-soundworks/sync/blob/master/client/index.es6.js
`serverTime` (T) is usually a time from a node.js server (`process.hrtime`) or a time from an other browser (`performance.now` via WebRTC). It could also be a local `performance.now` in our discussion. While `clientTime` is `currentTime` (t).
We generate ping-pong and we keep the quickest travel times, to exchange the time. With these samples over time (t, T), we compute a linear regression. Thanks to the linear regression, we are able to estimate the relationship between the times, and we are also smoothing out the variations of `currentTime`, with t(T) = t0 + (T - T0) / R. (Note that we are not reading t here, we are using t0 and T0, which are the mean values used for the regression.)

Then, we schedule everything using T, a logical synchronised time, when we do not need sample -accurate time. And there are short-term lookahead schedulers that convert to the local `currentTime`
(We use short-term free-running `currentTime`-based processes for sample-accurate processing.)

Here is a simple example of calibration, that uses this synchronisation. Because we do not now neither the latency, nor the gain, we run it before experiences when people come with their own devices. Once the compensation is set, the devices should keep together. (The database is not up-to-date, as we are waiting for the exposition of the latency.)
https://github.com/collective-soundworks/soundworks-calibration.git
(npm install && gulp)

Based on this, we developed a distributed reverberation, a collective loop-sequencer, etc. It works pretty well.

joeberkovitz:
Thanks @jipodine, I remember I saw a paper on this at WAC in Paris. I believe that this sort of approach should be mostly unnecessary in the future once we have a first-class method for determining latency.

However I would like to point out that there is a subtle issue with the description you offered that I still want to drill down on further. You said [emphasis mine]:

> Yes, I am talking about currentTime that updates at irregular intervals. We chose to base our distributed rendering on it because [...] with enough samples, its accumulated precision is fine (_not so far from context time_)

You speak of context time as a definite value that is not far off from `currentTime`. However, "context time" is a system of time measurement with an origin -- an axis, if you like. It is not any particular time value on that axis. So my question to you is: what is your definition of the specific "context time" value in the above sentence? That's the thing that I can't understand from the discussion so far.

jipodine:
@joeberkovitz I will present the paper at the next WAC. I would love that the facilities leave in the standard, it would be easier and more precise: use any time as a reference, re-sample to correct the drift with sample-accuracy... even across devices. :)

To me, the context time is the audio time, that starts at 0 when the context starts, and that allows us the schedule sample-accurate events with this absolute time. When reading this time, via `currentTime`, the precision is limited by the system block duration, but this refers to the same time. Is it correct? (I think I used both terms interchangeably; the same with `performance.now` for reading the performance time. Sorry for the confusion.)

The idea is to get a high-resolution estimation of the context time. As we use an other (high-resolution) clock to to this, we need to jointly estimate the relative drift.

To simplify, t = context time, T = performance time. When we read the context time, via `currentTime`, we get a value for t, with a precision of a system audio block ∆t. In the previous example ∆t = 4 * 128 / 44100 ≈ 12 ms. After getting samples (ti, Ti), we can compute a linear regression. Then, we got an estimation of t, the context time, out of T, the performance time:

t(T) = t0 + (T - T0) / R

It is true that our reference (t0, T0) is arbitrary, but it does not matter. What matters is that there are enough samples, so as t0 refines. It also matters that the sampling duration is long enough, so as the estimation of the drift R is good enough. Its relative error is around ∆t/t: if we want to cope with a 100PPM drift (2 clocks with 50 PPM), we need 120 seconds of observation with our example. This is what we use. (Before that, we correct only the delay between the times, to quickly start.)

joeberkovitz:
Thanks for the detailed answer, but I don't think you answered the question I was asking. Please let me try again, I appreciate your patience.

"Context time" is not a single time value at some moment: it is a linear system of time measurement in which samples and audio events are scheduled. It does not suffice to say "_the_ context time": there is no privileged context time. we need to say _which_ context time within this system of measurement we are referring to.

Let's define "smoothly increasing" as saying that the derivative of some function with respect to `performance.now()` time is almost exactly 1, up to clock drift.

There is no smoothly increasing context time that can be based on block rendering (which is what `currentTime` represents). `currentTime` simply tells you what is the context time associated with the start of the next block, and it's inherently non-smooth because its purpose is to tell you what the renderer is doing, not what the user is hearing.

However, there _is_ a smoothly increasing context time *with respect to audio output*. This is the context time value associated with the _sample that the user is currently hearing_. 

I think you are trying to estimate this latter time in your work (the context time associated with audio output). Is this understanding correct?

jipodine:
> Thanks for the detailed answer, but I don't think you answered the question I was asking. 

Well, I tried to explain you my very practical use-case of a mapping.

> we need to say which context time within this system of measurement we are referring to.

I refer a to a single audio context. Of course, not an off-line one. Except for those, I never use more than one. Indeed, I am trying to get a single reference time, or a mapping between the times.

> However, there is a smoothly increasing context time *with respect to audio output*. This is the context time value associated with the *sample that the user is currently hearing*.
> I think you are trying to estimate this latter time in your work (the context time associated with audio output). Is this understanding correct?

You understood very well, yes. Within a context, I try to estimate this "smoothly increasing context time *with respect to audio output*" according to the only observable value, which is `currentTime`. I certainly know how it works. As it comes from the audio hardware, it is forced to advance with it, at least in the mid-term, as the output buffer is filled.

To advance, I proposed one solution that suffers from its user-space implementation, and that only needs the latency to work today. What I would love to use is a unified high-resolution time, with an easy mapping or slaving between contexts. Whatever the implementation, really.

joeberkovitz:
@jipodine Thanks, now that we're on the same page I think we're going to be able to design a solution that will work for you and save us all a lot of trouble.

joeberkovitz:
Please read https://github.com/WebAudio/web-audio-api/issues/12#issuecomment-178587849 for the updated WG proposal on this topic.

padenot:
We can close this, the the work is happening in #754.

PR #683, opened 2015-12-11
Specify omni-directional source for PannerNode

### LINKED ISSUE #682: Inconsistency on cone inner and outer angles for PannerNode
The [spec](http://webaudio.github.io/web-audio-api/#widl-PannerNode-coneInnerAngle) currently says the cone angles are taken modulo 360.  The spec also says the default angle is 360.

If we follow the spec, the default angles are then treated as 0 deg. In particular, the outer angle is 0, which means the sound is reduced to coneOuterGain (defaulting to 0), except if you're directly pointed at the listener.

With Chrome (and Safari), if the inner and outer angles are both 360, no cone effects are applied. That is, the source is treated as omni-directional.  Firefox appears to do the same thing.

Proposal:  Add text to say the if the inner and outer angles are both 360, the source is treated as omnidirectional.

rtoy:
Fix #682 by specifying that if the cone inner and outer angles are
both 360 deg, the source is omni-directional and no cone effects are
applied.

### DIFF
--- a/index.html
+++ b/index.html
@@ -8119,7 +8119,7 @@ if (pan &lt;= 0) {
     return v;
   }
 
-  if (dotProduct(source.orientation, source.orientation) == 0 || ((source.coneInnerAngle == 0) &amp;& (source.coneOuterAngle == 0)))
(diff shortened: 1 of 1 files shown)


rtoy:
Issue #695 makes this invalid.  Closing.

PR #691, opened 2016-01-07
Add length attribute to OfflineAudioContext

### LINKED ISSUE #690: Add OfflineAudioContext length attribute
It would be rather nice if an OfflineAudioContext included (read-only) attributes for the number of channels and the length that was used to create the context.

Yes, I know I can go look through the code to find out how it was constructed, but it would certainly be nice to be to do a little introspection and just ask the context what the values were.  Yes, I can also render the context and look at the resulting buffer, but I want to do this before rendering is done.

rtoy:
Fixes #690

### DIFF
--- a/index.html
+++ b/index.html
@@ -1646,8 +1646,8 @@ function setupRoutingGraph() {
           </dt>
           <dd>
             <p>
-              The size of the buffer in sample-frames.  This is the same as the value of the
-              <code>length</code> parameter for the constructor.
+              The size of the buffer in sample-frames. This is the same as the
+              value of the <code>length</code> parameter for the constructor.
             </p>
           </dd>
           <dt>
--- a/index.html
+++ b/index.html
@@ -1642,6 +1642,15 @@ function setupRoutingGraph() {
             </dl>
           </dd>
           <dt>
+            readonly attribute unsigned long length
+          </dt>
+          <dd>
+            <p>
+              The size of the buffer in sample-frames.  This is the same as the value of the
+              <code>length</code> parameter for the constructor.
+            </p>
+          </dd>
+          <dt>
             attribute EventHandler oncomplete
           </dt>
           <dd>


index.html
```diff
@@ -1642,6 +1642,15 @@ <h2 id="OfflineAudioContext">
             </dl>
           </dd>
           <dt>
+            readonly attribute unsigned long length
```
rtoy: Since the parameter is called numberOfFrames, we should probably call this numberOfFrames too instead of length.

rtoy: Oops. It's numberOfFrames in chrome's idl.  But numberOfFrames is probably more intuitive than just length.

hoch: I find `length` is more consistent with the current constructor pattern. We
already use the term `duration` to specify the number is in 'seconds'.

padenot: Length is always in frames in the API, yeah.

rtoy: Nothing for me to do then.  Ready to merge?

domenic:
I would really suggest the group consider an alternative name (`numberOfFrames` is nice and descriptive; `frameLength` is reasonable too). `length` on the platform always indicates an indexed collection.

hoch:
I really do not have a strong opinion on this, but `length` here is literally the length of the array buffer. So I think it nicely falls into the category of 'indexed collection'. No?

domenic:
No. OfflineAudioContext itself is not an indexed collection; that is, you cannot do `oac[0]` through `oac[oac.length - 1]`.

hoch:
Hmm. `AudioBuffer` also is not an array and you can't access with the index. We're simply following the convention to `AudioBuffer.length`, because that's what OAC produces.

And we do not have any instance of `numberOfFrames` nor `frameLength` in our spec, so it will be a new property name.

@padenot WDYT?

domenic:
Oh, I see. In that case maybe consistency is better. The correct time to fix this would have been when designing AudioBuffer. It's a real shame that AudioBuffer uses length for this purpose.

padenot:
I agree that consistency is better.

index.html
foolip: The type `unsigned long` matches the constructor, but it doesn't match `AudioBuffer.prototype.length`, which is just `long`. Filed https://github.com/WebAudio/web-audio-api/issues/747 for a similar issue.

foolip:
Is there anything still blocking this from being merged? The [Intent to Implement and Ship: WebAudio OfflineAudioContext.length](https://groups.google.com/a/chromium.org/d/msg/blink-dev/qcBNCz8A5eY/KX4y6GtPAQAJ) for Blink would benefit from having this in the spec.

padenot:
No, sorry, just an oversight on my part.

Issue #708, opened 2016-01-26
No mention of what AudioNode.connect() returns.


jan-ivar:
AudioNode.connect() supposedly returns the destination object. At least I think so because:

1. I infer it from the *other* `connect` method (there are two) explicitly saying it does *not* return it.
2. It's what the Firefox implementation does today.
3. It's quite [handy for chaining](https://jsfiddle.net/jib1/7wd2z8rz/): `nodes.reduce((a, b) => a && a.connect(b));`

However there's no explicit mention of what it returns, which led [MDN to get it wrong](https://developer.mozilla.org/en-US/docs/Web/API/AudioNode/connect%28AudioNode%29$revision/988065), so there probably should be.

hoch:
I believe this is explicit enough on the spec. Take a look at the IDL for AudioNode.

```webidl
AudioNode connect (AudioNode destination, optional unsigned long output = 0
              , optional unsigned long input = 0
              );
```

mdjp:
After working group discussion agreed with comment from @hoch this is already specified.

jan-ivar:
Sorry, I missed your comment 4 days ago, where exactly. How is that explicit in any way? It says it returns a value of type `AudioNode` but there's no mention of which one, `this` or `destination`. Am I missing something?

jan-ivar:
Apparently I'm not allowed to reopen. Paging @hoch and @mdjp.

jan-ivar:
As I mentioned in https://github.com/WebAudio/web-audio-api/issues/708#issue-128686551 MDN had it returning `this` which is wrong.

hoch:
Thanks for pointing that out. In the review process of the PR(https://github.com/hoch/web-audio-api/commit/bfa4fa044a009d719423bf857fa4fb07fb52f489), somehow we decided to remove the description of method chaining.

I am opening the issue and will review your PR.

By the way, if you found something wrong with MDN, probably this is not the right place to report.

skratchdot:
Agreed. Spec is ambiguous. Could return `source/this` or `destination`.  Implementors involved with the spec might understand the meaning, but it's not clear from: http://webaudio.github.io/web-audio-api/#widl-AudioNode-connect-AudioNode-AudioNode-destination-unsigned-long-output-unsigned-long-input

The function returns `void` when `input` is not passed in.  It returns `AudioNode` when `input` is passed in.  A simple description next the to `Return type: AudioNode` would fix the ambiguity.

change:
```html
<div>
[...]

PR #709, opened 2016-01-26
Mention that AudioNode.connect() returns destination object.

jan-ivar:
Fix for https://github.com/WebAudio/web-audio-api/issues/708.


rtoy:
This was fixed in PR #710 and #711.

PR #716, opened 2016-02-10
Add spatialListener attribute to AudioContext.

### LINKED ISSUE #714: Need to add a spatialListener on AudioContext
Although we define the SpatialListener interface, we don't have an instance of it on AudioContext to mirror the "listener" property.

rtoy:
Fix #714 by adding spatialListener attribute.

### DIFF
--- a/index.html
+++ b/index.html
@@ -808,6 +808,16 @@ function setupRoutingGraph() {
             </p>
           </dd>
           <dt>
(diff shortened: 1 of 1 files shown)


rtoy:
I think the only thing really needing review is the actual name for the attribute.  I think spatialListener is not controversial.

padenot:
LGTM.

Issue #718, opened 2016-02-12
5.1 to mono downmixing


rtoy:
The downmixing formula in http://webaudio.github.io/web-audio-api/#down-mix for 5.1 to mono says

    output = 0.7071 * (input.L + input.R) + input.C + 0.5 * (input.SL + input.SR)

This explicitly says 0.7071, but presumably it really meant sqrt(1/2).  I think the formula should say sqrt(1/2) instead of 0.7071.  If you're testing downmixing, 0.7071 vs sqrt(1/2) makes a difference.

hoch:
+1

rtoy:
Fixed by PR #720

PR #720, opened 2016-02-12
Replace 0.7071 with sqrt(0.5)

rtoy:
Fix issue #718.  I think the intent is use sqrt(0.5) and not 0.7071.


padenot:
Yes, thanks.

PR #727, opened 2016-02-18
Truncate float to integer for Analyser.getByte.*Data

### LINKED ISSUE #722: Clarify getByteTimeDomainData() computation
The spec for [getByteTimeDomainData](https://webaudio.github.io/web-audio-api/#widl-AnalyserNode-getByteTimeDomainData-void-Uint8Array-array) says how to compute the byte value from the float, but it isn't explicit about how to round the float value to a byte value.

Chrome currently uses the floor function to truncate the float to a byte.  We should either specify this or say explicitly say how to compute the value, perhaps by rounding to the nearest integer.

rtoy:
Explicitly say that the byte data is computed from the floating point
data by truncating the float to an integer.

Fix #722

### DIFF
--- a/index.html
+++ b/index.html
@@ -5358,11 +5358,12 @@ function calculateNormalizationScale(buffer)
             </p>
             <pre>
                   $$
-                    b[k] = \frac{255}{\mbox{dB}_{max} - \mbox{dB}_{min}}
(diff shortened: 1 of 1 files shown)


padenot:
Looks good, the indent is off tough. Can you fix it and merge ?

PR #728, opened 2016-02-19
Scale the FFT by 1/N in the AnalyserNode

### LINKED ISSUE #725: Analyser frequency computations missing a scaling factor
The equations describing how the AnalyserNode computes the frequency result is missing a scaling factor.  In Chrome, the magnitude is scaled by 1/fftSize so that the 0 dB sine wave is 0 dB. I think this scaling makes sense.

The formula needs to be updated to reflect this scaling.

rtoy:
Apply a scale factor to the FFT so that a 0 dbFS sine wave is 0 dBfs.

Chrome does this; I didn't check anywhere else, but it makes sense
that the FFT output should have the same amplitude as the input,
instead of leaving the FFT unscaled.

Fix #725.

### DIFF
--- a/index.html
+++ b/index.html
@@ -5548,10 +5548,9 @@ function calculateNormalizationScale(buffer)
           </p>
           <pre>
             $$
-              X[k] = \sum_{n = 0}^{N - 1} \hat{x}[n] e^{\frac{-2\pi i k n}{N}}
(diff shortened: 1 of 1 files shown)


padenot:
Gecko does this as well.

PR #738, opened 2016-02-25
Fix #729: Specify how multiple calls to get{Float,Byte}FrequencyData works

### LINKED ISSUE #729: Multiple calls to getFloatFrequencyData
Consider the following bit of code:

```
var context = new OfflineAudioContext(1, 44100, 44100);

osc.frequency.setValueAtTime(261.6, 0.0);

var a = context.createAnalyser();
a.fftSize = 32;

osc.connect(a);
a.connect(context.destination);

osc.start();

context.suspend(128/context.sampleRate).then(function () {
 var f1 = new Float32Array(a.frequencyBinCount);
 var f2 = new Float32Array(a.frequencyBinCount);

 a.getFloatFrequencyData(f1);
 a.getFloatFrequencyData(f2);

 console.log(f1);
 console.log(f2);

 context.resume();
});
```

What should the console log contain?  Currently on Chrome, Edge, and Firefox, the values of f1 and f2 are different.  Why?  Because each call to getFloatFrequencyData computes (the same) FFT again and applies the smoothing again.

I find this behavior odd.  I think f1 and f2 should return exactly the same values because I'm asking for the FFT at exactly the same time.  There shouldn't be additional smoothing applied for each call.

I concede that the spec doesn't explicitly allow or forbid either of the behaviors.

One use case (from @hoch) where it might really matter is if you want to do getByteFrequencyData and getFloatFrequencyData at the same time.  The byte value is used for an animation and the float value is used for further detailed analysis and processing.  Ideally, the values should be consistent, but they're not; one has an extra smoothing applied.

rtoy:
If a call to getFloatFrequencyData occurs in the same rendering
quantum as a previous call, return the existing data.  This is useful
in getting consistent data if getByteFrequencyData and
getFloatFrequencyData are called in the same rendering quantum.

Fix #729


index.html
```diff
@@ -5581,6 +5581,13 @@ <h3 id="Convolution-channel-configurations">
               be ignored.
             </p>
             <p>
+              If another call to <code>getFloatFreqencyData</code> or
```
padenot: Typo here (missing U).

padenot:
It's a bit weird to reference the concept of a "rendering quantum" from the main thread, because we haven't speced yet when `currentTime` is updated from the main thread point of view.

rtoy:
Do you have a suggestion on how to rephrase this?

padenot:
I don't really know, but I imagine that if we'd spec the way `currentTime` is updated on the main thread, than we could tie this new concept to the update of the main thread buffer of `get{Float,Byte}FrequencyData`.

joeberkovitz:
I think the problems here have less to do with `currentTime` and more to do with the Processing Model failing to address the notion of AudioNodes undertaking rendering-thread actions _other than computing a block of audio_.

In https://webaudio.github.io/web-audio-api/#rendering-loop, step 6.3 currently reads: "If this AudioNode is a source node, compute a block of audio, and make it available for reading."

I think it would be useful to leverage to broaden the notion of _compute a block of audio_ in the definition of `AnalyserNode` and state that the computation of current frequency data occurs as part of computing a block of audio, and that this is communicated to the control thread via queuing or other specifiable mechanism. Additionally we might choose to expand the definition in step 6.3 of the Processing Model to say something like "cause the AudioNode to compute a block of audio and perform any other other block-based computations" to cover this and other such cases in the future.

rtoy:
https://webaudio.github.io/web-audio-api/#rendering-loop specifies how `currentTime` is updated.  Is that enough?

I think the issue about the processing model in https://github.com/WebAudio/web-audio-api/pull/738#issuecomment-236437051 is a different issue from this.  At least in chrome, the processing of the analyzer block data does not happen on the rendering thread.  It grabs the current set of data (which does get updated by the rendering thread) to compute the FFT.

rtoy:
@padenot PTAL. Updated the text to link to render quantum which is now defined precisely.

PR #743, opened 2016-03-01
Spell out the algorithm to set AudioBufferSourceNode.buffer.

### LINKED ISSUE #666: The AudioBufferSourceNode.prototype.buffer setter doesn't handle null correctly
https://webaudio.github.io/web-audio-api/#AudioBufferSourceNode

"This attribute can only be set once, or a InvalidStateError must be thrown."

This needs clarification. Setting `buffer` to null when it already null should arguably do nothing, while trying to set `buffer` to anything when it is not null should throw InvalidStateError. The way it's written, this should throw:
```JavaScript
var source = context.createBufferSource();
source.buffer = null;
source.buffer = null;
```

And that would make no sense.

padenot:
This fixes #666.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1070,6 +1070,8 @@ function setupRoutingGraph() {
           </dt>
           <dd>
             Creates an <a><code>AudioBufferSourceNode</code></a>.
+            <a>AudioBufferSourceNode</a> are created with an internal flag
+            <code>buffer set</code>, initially set to false.
           </dd>
           <dt>
             Promise&lt;AudioWorker&gt; createAudioWorker()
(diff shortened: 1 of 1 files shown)


padenot:
You're right, setting `null` multiple times should be legal.

rtoy:
lgtm.

foolip:
No *buffer set* internal flag ought to be needed for this, and there is none in Blink's implementation. Instead it simply throws if in the setter if the buffer is already non-null. The spec now says "If new buffer is not null and buffer set is true, throw an InvalidStateError" which would allow setting the buffer back to null, which doesn't seem right. Can you reopen #666?

foolip:
I'd suggest "If *new buffer* is not null and `buffer` is not null, throw an InvalidStateError and abort these steps." That would avoid throwing when setting to null when `buffer` is already null, but otherwise matches Blink, which does the same but without the "*new buffer* is not null" condition.

rtoy:
Blink isn't compliant with what this issue proposes because Blink doesn't allow you set the buffer to null.

The intent is that the following is legal:

    s = context.createBufferSource();
    s.buffer = null;
    s.buffer = context.createBuffer(...)
    s.buffer = null;

But if we now try to set the buffer to another non-null buffer, an error is thrown.

foolip:
Sure, but why should it be allowed to set `buffer` back to null? It would leave the `AudioBufferSourceNode` in a state where it can never do anything useful again, so why not just disconnect it and let it be GC'd?

rtoy:
I have no recollection of how we came to this point. Let me look....

rtoy:
Ok.  I think this comes from the fact that `AudioBufferSourceNode.buffer` is of type `AudioBuffer?` because the initial value is `null`.  Because of this, it seemed that we should be able to set buffer back to null after setting it to some other non-null AudioBuffer.  But to preserve the one-shot fire-and-forget aspect, you cannot set buffer to another non-null value once you've already done that.

I think the preferred method is to let it finish or stop() it and drop the reference; this should allow the node to be collected once it has stopped playing.

foolip:
Right, the only point of setting `buffer` back to null would be to allow the `AudioBuffer` to be GC'd, but the `AudioBufferSourceNode` itself is useless as that point so one might as well allow that to be GC'd instead.

PR #754, opened 2016-03-08
Fix #12 by adding outputTimeStamp attribute to AudioDestinationNode

### LINKED ISSUE #12: Need a way to determine AudioContext time of currently audible signal

pozdnyakov:
This patch introduces 'AudioDestinationNode.outputTimeStamp'
and 'AudioDestinationNode.outpuLatency' attributes, so that it's possible
to estimate when the given context's stream position is audible.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2547,6 +2547,31 @@ function setupRoutingGraph() {
         </section>
       </section>
       <section>
+        <h2>
+          AudioTimeStamp
+        </h2>The <code>AudioTimeStamp</code> dictionary is used to bind a
+        context's time value to a corresponding
+        <code>DOMHighResTimeStamp</code> value (described in [[!hr-time-2]]).
+        <dl title="dictionary AudioTimeStamp" class="idl">
+          <dt>
+            double contextTime
+          </dt>
+          <dd>
+            Represents a point in the time coordinate system of context's
+            <a href=
+            "#widl-BaseAudioContext-currentTime"><code>currentTime</code></a>.
+          </dd>
+          <dt>
+            DOMHighResTimeStamp performanceTime
+          </dt>
+          <dd>
+            Represents a point in the time coordinate system of a
+            <code>Performance</code> interface implementation (described in
+            [[!hr-time-2]]).
+          </dd>
+        </dl>
+      </section>
+      <section>
         <h2 id="AudioDestinationNode">
           The AudioDestinationNode Interface
         </h2>
@@ -2615,6 +2640,100 @@ function setupRoutingGraph() {
               "#widl-AudioNode-channelCount"><code>channelCount</code></a> is
               determined when the offline context is created and this value may
               not be changed.
+            </p>
+          </dd>
+          <dt>
+            readonly attribute double outputLatency
+          </dt>
+          <dd>
+            <p>
+              The estimation in seconds of audio output latency, i.e., the
+              interval between the time the UA requests the host system to play
+              a buffer and the time at which the first sample in the buffer is
+              actually processed by the audio output device. For devices such
+              as speakers or headphones that produce an acoustic signal, this
+              latter time refers to the time when a sample's sound is produced.
+            </p>
+            <p>
+              The <a><code>outputLatency</code></a> attribute value depends on
+              the platform and the connected hardware audio output device. The
+              <a><code>outputLatency</code></a> attribute value does not change
+              for the context's lifetime as long as the connected audio output
+              device remains the same. If the audio output device is changed
+              the <a><code>outputLatency</code></a> attribute value will be
+              updated accordingly.
+            </p>
+          </dd>
+          <dt>
+            readonly attribute AudioTimeStamp outputTimeStamp
+          </dt>
+          <dd>
+            <p>
+              Holds an <a><code>AudioTimeStamp</code></a> instance containing
+              two correlated context's audio stream position values: the
+              <a href=
+              "#widl-AudioTimeStamp-contextTime"><code>contextTime</code></a>
+              member contains the time of the sample frame which is currently
+              being rendered by the audio output device (i.e., output audio
+              stream position), in the same units and origin as context's
+              <a href=
+              "#widl-BaseAudioContext-currentTime"><code>currentTime</code></a>;
+              the <a href=
+              "#widl-AudioTimeStamp-performanceTime"><code>performanceTime</code></a>
+              member contains the time estimating the moment when the sample
+              frame corresponding to the stored <code>contextTime</code> value
+              was rendered by the audio output device, in the same units and
+              origin as <code>performance.now()</code> (described in
+              [[!hr-time-2]]).
+            </p>
+            <p>
+              If the context's rendering graph has not yet processed a block of
+              audio, then <a><code>outputTimeStamp</code></a> holds an
+              <code>AudioTimeStamp</code> instance with both members containing
+              zero.
+            </p>
+            <p>
+              After the context's rendering graph has started processing of
+              blocks of audio, its <a href=
+              "#widl-BaseAudioContext-currentTime"><code>currentTime</code></a>
+              attribute value always exceeds the <a href=
+              "#widl-AudioTimeStamp-contextTime"><code>contextTime</code></a>
+              value obtained from <a href=
+              "#widl-AudioDestinationNode-outputTimeStamp"><code>outputTimeStamp</code></a>
+              attribute.
+            </p>
+            <p>
+              The value of <a><code>outputTimeStamp</code></a> attribute can be
+              used to get performance time estimation for the slightly later
+              context's time value:
+            </p>
+            <pre class="highlight example">
+            function outputPerformanceTime(contextTime) {
+                var timestamp = destinationNode.outputTimeStamp;
+                var elapsedTime = contextTime - timestamp.contextTime;
+                return timestamp.performanceTime + elapsedTime * 1000;
+            }
+            </pre>
+            <p>
+              In the above example the accuracy of the estimation depends on
+              how close the argument value is to the current output audio
+              stream position: the closer the given <code>contextTime</code> is
+              to <code>timestamp.contextTime</code>, the better the accuracy of
+              the obtained estimation.
+            </p>
+            <p class="note">
+              The difference between the values of context's <a href=
+              "#widl-BaseAudioContext-currentTime"><code>currentTime</code></a>
+              and the <a href=
+              "#widl-AudioTimeStamp-contextTime"><code>contextTime</code></a>
+              obtained from <a href=
+              "#widl-AudioDestinationNode-outputTimeStamp"><code>outputTimeStamp</code></a>
+              attribute value cannot be considered as a reliable output latency
+              estimation because <a href=
+              "#widl-BaseAudioContext-currentTime"><code>currentTime</code></a>
+              may be incremented at non-uniform time intervals, so <a href=
+              "#widl-AudioNode-outputLatency"><code>outputLatency</code></a>
+              attribute should be used instead.
             </p>
           </dd>
         </dl>


index.html
```diff
@@ -2309,6 +2309,97 @@ <h2 id="AudioDestinationNode">
...
+          <dd>
+            <p>
+              The estimation in seconds of audio output latency, i.e. the delay
+              between the moment the implementation sends a buffer to the
+              system and the audible sound from the audio output (speaker or
+              connector).
+            </p>
+          </dd>
+          <dt>
+            AudioTimestamp&lt;void&gt; getPosition()
```
rtoy: getPosition seems rather vague. Is there a better name?

index.html
```diff
@@ -2309,6 +2309,97 @@ <h2 id="AudioDestinationNode">
...
+            </p>
+            <pre class="highlight example">
+            function outputPerformanceTime(contextTime) {
+                var timestamp = destNode.getPosition();
+                var elapsedTime = contextTime - timestamp.contextTime;
+                return timestamp.performanceTime + elapsedTime * 1000;
+            }
+            </pre>
+            <p>
+              The closer input value to <code>timestamp.contextTime</code>, the
```
rtoy: Please clarify. I can't figure out what "closer input value" really means.

index.html
```diff
@@ -2309,6 +2309,97 @@ <h2 id="AudioDestinationNode">
...
+              The difference between the values of context's <a href=
+              "#widl-BaseAudioContext-currentTime"><code>currentTime</code></a>
+              and the <a href=
+              "#widl-AudioTimestamp-contextTime"><code>contextTime</code></a>
+              returned from <a href=
+              "#widl-AudioNode-getPosition"><code>getPosition</code></a> call
+              cannot be considered as a reliable output latency estimation
+              because <a href=
+              "#widl-BaseAudioContext-currentTime"><code>currentTime</code></a>
+              is incremented in jumpy way and at unpredictable moments, so
```
rtoy: "is incremented in jumpy way and at unpredictable moments" -> "may be incremented at non-uniform time intervals"

index.html
pozdnyakov: Would `getStreamPosition` be better?

index.html
pozdnyakov: yeah, thanks!

index.html
```diff
@@ -2309,6 +2309,108 @@ <h2 id="AudioDestinationNode">
...
+          <dd>
+            <p>
+              The estimation in seconds of audio output latency, i.e. the delay
+              between the moment the implementation sends a buffer to the
+              system and the audible sound from the audio output (speaker or
+              connector).
+            </p>
+            <p>
+              The <a><code>latency</code></a> attribute value depends on the
+              platform and it does not change for the context's lifetime.
```
padenot: What happens if you unplug your usb device ? This changes the sound card and probably changes the latency as well. Same think if you disconnect an bluetooth device that is using a2ps, you go from 400+ms to low latency.

pozdnyakov: The attribute value should change in this case indeed, I'll update the description accordingly.. Thanks for pointing it out.

pozdnyakov:
@padenot does it look OK now?

index.html
```diff
@@ -2309,6 +2309,112 @@ <h2 id="AudioDestinationNode">
               not be changed.
             </p>
           </dd>
+          <dt>
+            readonly attribute double latency
```
rtoy: Since we also now have baseLatency (for the playback category stuff), latency is, perhaps, too generic.  Maybe outputLatency, or deviceOutputLatency, or something else?

index.html
```diff
@@ -2309,6 +2309,112 @@ <h2 id="AudioDestinationNode">
...
+          </dt>
+          <dd>
+            <p>
+              The estimation in seconds of audio output latency, i.e. the delay
+              between the moment the implementation sends a buffer to the
+              system and the audible sound from the audio output (speaker or
+              connector).
```
rtoy: What does connector mean here?

index.html
```diff
@@ -2309,6 +2309,112 @@ <h2 id="AudioDestinationNode">
...
+              member contains performance time estimation of the moment when
+              the stored audio stream position is rendered, in the same units
+              and origin as <code>performance.now()</code> (described in
+              [[!hr-time-2]]).
+            </p>
+            <p>
+              If the context's rendering graph has not yet processed a block of
+              audio, then <a><code>devicePosition</code></a> holds an
+              <code>AudioTimestamp</code> instance with both members containing
+              zero.
```
rtoy: So, after the first rendered data is sent, the devicePosition suddenly changes from zero to performance.now() time?

index.html
```diff
@@ -2309,6 +2309,112 @@ <h2 id="AudioDestinationNode">
...
+              The <a><code>latency</code></a> attribute value depends on the
+              platform and the hardware audio output device, the attribute
+              value does not change for the context's lifetime as long as audio
+              output device remains the same. If the audio output device is
+              changed the <a><code>latency</code></a> attribute value will be
+              updated accordingly.
+            </p>
+          </dd>
+          <dt>
+            readonly attribute AudioTimestamp devicePosition
```
rtoy: I keep reading "devicePosition" as the physical position of the device.  Can we choose a different name that's more indicative of the time nature of the attribute?

index.html
```diff
@@ -2309,6 +2309,112 @@ <h2 id="AudioDestinationNode">
...
+                var elapsedTime = contextTime - timestamp.contextTime;
+                return timestamp.performanceTime + elapsedTime * 1000;
+            }
+            </pre>
+            <p>
+              In the above example the preciseness of the estimation depends on
+              how close the argument value is to the current output audio
+              stream position: the closer given <code>contextTime</code> to
+              <code>timestamp.contextTime</code>, the better preciseness of the
+              obtained estimation.
```
rtoy: Grammar:  "the closer given..., the better preciseness...." ->
"the closer the given contextTime is to timestamp.contextTime, the better the accuracy of the obtained estimation."

index.html
```diff
@@ -2309,6 +2309,112 @@ <h2 id="AudioDestinationNode">
...
+            </p>
+            <pre class="highlight example">
+            function outputPerformanceTime(contextTime) {
+                var timestamp = destNode.devicePosition;
+                var elapsedTime = contextTime - timestamp.contextTime;
+                return timestamp.performanceTime + elapsedTime * 1000;
+            }
+            </pre>
+            <p>
+              In the above example the preciseness of the estimation depends on
```
rtoy: I think we want "accuracy", not "preciseness" here.

index.html
```diff
@@ -2309,6 +2309,112 @@ <h2 id="AudioDestinationNode">
...
+              may be incremented at non-uniform time intervals, so <a href=
+              "#widl-AudioNode-latency"><code>latency</code></a> attribute
+              should to be used instead.
+            </p>
+          </dd>
+        </dl>
+      </section>
+      <section>
+        <h2>
+          AudioTimestamp
```
rtoy: Not sure what respec does, but should the definition of the AudioTimestamp dictionary come before the description of the attributes?

index.html
joeberkovitz: I think this language is a little vague though the intent is clear. My suggestion would be:

The estimation in seconds of audio output latency, i.e. the interval between the time
the UA requests the host system to play a buffer and the time at which the first sample in the buffer
is actually processed by the audio output device. For audible devices such as speakers or headphones
that produce an acoustic signal, this latter time refers to the time when a sample's sound is produced.

index.html
hoch: We already have the `baseLatency` in the context, this needs to be moved to there. Getting two different latency values from different entities is weird. Also this should be renamed - I prefer `deviceOutputLatency`.

index.html
hoch: I second @rtoy's feedback - `devicePosition` needs to be renamed. Here are my thoughts:
1. This should be queried through the audio context, not the destination.
2. Why is this a getter? Wouldn't it be better to use a method since it returns an object, not a number?
3. How about `EstimatedTimeStamp` for the dictionary name?

index.html
```diff
@@ -2309,6 +2309,112 @@ <h2 id="AudioDestinationNode">
...
+          </dd>
+          <dt>
+            readonly attribute AudioTimestamp devicePosition
+          </dt>
+          <dd>
+            <p>
+              Holds an <a><code>AudioTimestamp</code></a> instance containing
+              two correlated context's audio stream position values: the
+              <a href=
+              "#widl-AudioTimestamp-contextTime"><code>contextTime</code></a>
```
hoch: I know you have the working prototype of this implementation; can you point me where the calculation of `contextTime` is happening? How do you estimate this number? The algorithm might be written in the spec as well. Otherwise, you'll see different estimations across browsers. We can do our best, but our best might not be what you want.

index.html
```diff
@@ -2309,6 +2309,112 @@ <h2 id="AudioDestinationNode">
...
+              zero.
+            </p>
+            <p>
+              The value of <a><code>devicePosition</code></a> attribute can be
+              used to get performance time estimation for the slightly later
+              context's time value:
+            </p>
+            <pre class="highlight example">
+            function outputPerformanceTime(contextTime) {
+                var timestamp = destNode.devicePosition;
```
hoch: `destNode` to `destinationNode`

index.html
hoch: I suggest `AudioTimeStamp` for the dictionary name. (based on `DOMHighResTimeStamp`)

index.html
```diff
@@ -2309,6 +2309,112 @@ <h2 id="AudioDestinationNode">
...
+      </section>
+      <section>
+        <h2>
+          AudioTimestamp
+        </h2>The <code>AudioTimestamp</code> dictionary is used to bind a
+        context's time value to a corresponding
+        <code>DOMHighResTimeStamp</code> value (described in [[!hr-time-2]]).
+        <dl title="dictionary AudioTimestamp" class="idl">
+          <dt>
+            double contextTime
```
hoch: It's still confusing to me that now we have `contextTime` and `currentTime`. I strongly suggest to rename this.

index.html
pozdnyakov: It is just taken from the platform API (for Windows` IAudioClock::GetPosition`), no specific algorithms are required. Pls. see https://github.com/crosswalk-project/chromium-crosswalk/blob/96132b6d03991f931b48f138b83c8f06de9a4578/media/audio/win/audio_low_latency_output_win.cc#L544

index.html
pozdnyakov: what would you propose?  I can imagine `framePosition` or `audioTime` but both do not look better to me..

index.html
pozdnyakov: @rtoy would `outputTimeStamp` look OK for you?
@hoch why would you move it to context? This is an output device property.
Attributes are used much more often within the spec and many of them hold objects.. As I see methods usually do smth more complex than just returning a value, so I would keep it as an attribute. wdyt?

index.html
pozdnyakov: it's not exactly performance.now(), it's a performance time stamp of the moment when the audible output stream position was equal to `devicePosition.contextTime`. If nothing has been played the output stream position is equal to zero and both `devicePosition` field contain zero as well.

index.html
pozdnyakov: @joeberkovitz Thanks for you suggestion!

index.html
pozdnyakov: I'd prefer `outputLatency` as it matches to `outputTimeStamp` in the latest patch. IMO `outputLatency`  should remain at `AudioDestinationNode` as this interface is representing an output device, so it's a natural place to contain the attribute representing output device latency. wdyt?

index.html
pozdnyakov: indeed, thanks for pointing out

pozdnyakov:
@rtoy @joeberkovitz @hoch thanks a lot for your comments. They are applied in the latest patch.

svgeesus:
Discussed at 21 April telcon, looks good.

pozdnyakov:
> Discussed at 21 April telcon, looks good.

@padenot @rtoy can the PR be merged?

index.html
```diff
@@ -2614,6 +2639,90 @@ <h2 id="AudioDestinationNode">
               not be changed.
             </p>
           </dd>
+          <dt>
+            readonly attribute double outputLatency
+          </dt>
+          <dd>
+            <p>
+              The estimation in seconds of audio output latency, i.e. the
```
rtoy: "i.e." -> "i.e.," (Add comma)

index.html
```diff
@@ -2614,6 +2639,90 @@ <h2 id="AudioDestinationNode">
...
+              interval between the time the UA requests the host system to play
+              a buffer and the time at which the first sample in the buffer is
+              actually processed by the audio output device. For audible
```
rtoy: "audible"?  Probably best to just leave this out?

index.html
```diff
@@ -2614,6 +2639,90 @@ <h2 id="AudioDestinationNode">
...
+              devices such as speakers or headphones that produce an acoustic
+              signal, this latter time refers to the time when a sample's sound
+              is produced.
+            </p>
+            <p>
+              The <a><code>outputLatency</code></a> attribute value depends on
+              the platform and the hardware audio output device, the attribute
```
rtoy: "device, the attribute"  This is a run-on sentence, and I'm not sure exactly what you're saying.

index.html
```diff
@@ -2614,6 +2639,90 @@ <h2 id="AudioDestinationNode">
...
+          </dd>
+          <dt>
+            readonly attribute AudioTimeStamp outputTimeStamp
+          </dt>
+          <dd>
+            <p>
+              Holds an <a><code>AudioTimeStamp</code></a> instance containing
+              two correlated context's audio stream position values: the
+              <a href=
+              "#widl-AudioTimeStamp-contextTime"><code>contextTime</code></a>
```
rtoy: I'm confused on the difference between currentTime and contextTime.  You say below that contextTime is the estimate of the output currently being rendered by the AudioDestinationNode.  How is this value different from the currentTime of the AudioDestinationNode?

index.html
pozdnyakov: Done.

index.html
pozdnyakov: Fixed.

index.html
pozdnyakov: Done.

index.html
pozdnyakov: I think you can treat it as `currentTime` of the `AudioDestinationNode` (i.e.,  `currentTime` of the output currently being rendered). Would you like me to emphasize it in the description or rename `AudioTimeStamp.contextTime` into `AudioTimeStamp.currentTime`?

rtoy: Hmm.  This now brings up the question of how this is different from just grabbing context.currentTime and calling performanceNow() right after.

pozdnyakov: @rtoy In practice the difference can be seen at https://github.com/WebAudio/web-audio-api/issues/12#issuecomment-200521873 the `AudioTimeStamp` values are always behind `context.currentTime` and `performance.now()`. The difference depends on the system buffer size and output latency value.

`context.currentTime` represents the sample frame immediately following the last sample-frame in the block of audio most recently processed by the context's rendering graph.

`AudioTimeStamp.contextTime` represents the sample frame which is currently being rendered by the speaker. It can be considered as an analogue of [IAudioClock::GetPosition](https://msdn.microsoft.com/en-us/library/windows/desktop/dd370889(v=vs.85).aspx) from WASAPI.

rtoy: Ah, I get it now.  I think you should include some variant of this explanation in the description.  It makes it really clear what's happening.

pozdnyakov: I've rephrased the attribute description, PTAL.

rtoy: Looks good.  I think I'd also try to work in your comment:

> context.currentTime represents the sample frame immediately following the last sample-frame in the block of audio most recently processed by the context's rendering graph.

somewhere. It makes the distinction really clear.

Also, contextTime is, I think, to similar to currentTime.  Maybe outputContextTime to kind of match outputTimeStamp?

pozdnyakov: > I think I'd also try to work in your comment:

>>context.currentTime represents the sample frame immediately following the last sample-frame in the block of audio most recently processed by the context's rendering graph.

This is actually a quote from the `currentTime` attribute description, maybe there is a way to add a reference to this particular paragraph in order not to duplicate it?

> Also, contextTime is, I think, to similar to currentTime. Maybe outputContextTime to kind of match outputTimeStamp?

If possible I'd keep it generic, so that `AudioTimeStamp` is not bound to the particular attribute and can be reused (for example for input audio stream synchronization)

Thanks for your comments!

rtoy: On Fri, Apr 29, 2016 at 1:26 PM, Mikhail Pozdnyakov <
notifications@github.com> wrote:

> In index.html
> <https://github.com/WebAudio/web-audio-api/pull/754#discussion_r61639671>:
>
> > +              the platform and the hardware audio output device, the attribute
> > +              value does not change for the context's lifetime as long as audio
> > +              output device remains the same. If the audio output device is
> > +              changed the <a><code>outputLatency</code></a> attribute value
> > +              will be updated accordingly.
> > +            </p>
> > +          </dd>
> > +          <dt>
> > +            readonly attribute AudioTimeStamp outputTimeStamp
> > +          </dt>
> > +          <dd>
> > +            <p>
> > +              Holds an <a><code>AudioTimeStamp</code></a> instance containing
> > +              two correlated context's audio stream position values: the
> > +              <a href=
> > +              "#widl-AudioTimeStamp-contextTime"><code>contextTime</code></a>
>
> I think I'd also try to work in your comment:
>
> context.currentTime represents the sample frame immediately following the
> last sample-frame in the block of audio most recently processed by the
> context's rendering graph.
>
> This is actually a quote from the currentTime attribute description,
> maybe there is a way to add a reference to this particular paragraph in
> order not to duplicate it?
>

​Oh.  I'll leave it up to you to link it or not.​


> Also, contextTime is, I think, to similar to currentTime. Maybe
> outputContextTime to kind of match outputTimeStamp?
>
> If possible I'd keep it generic, so that AudioTimeStamp is not bound to
> the particular attribute and can be reused (for example for input audio
> stream synchronization)
>

​Ok, that's sensible.

So, I'm pretty much fine with the text as it is.

I invite others to have a final look at this.​


> Thanks for your comments!
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/WebAudio/web-audio-api/pull/754/files/d51bbd05700eb0ea0682d3fd008bcc7b383ff5d0#r61639671>
>



-- 
Ray

pozdnyakov:
>  I invite others to have a final look at this.​

@padenot, @joeberkovitz, others, are you OK with the PR?

pozdnyakov:
Judging by https://travis-ci.org/WebAudio/web-audio-api/builds/128246251 CI failure is not related to the PR (locally tidy passes without errors in the patch).

rtoy:
I suspect that this is caused by someone checking in a new revision that
use CR/LF (windows end-of-line) instead of LF (Unix).

I will fix this shortly.

On Mon, May 9, 2016 at 3:58 AM, Mikhail Pozdnyakov <notifications@github.com
> wrote:

> Judging by https://travis-ci.org/WebAudio/web-audio-api/builds/128246251
> CI failure is not related to the PR (locally tidy passes without errors in
> the patch).
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/WebAudio/web-audio-api/pull/754#issuecomment-217835082>
>



-- 
Ray

pozdnyakov:
All checks have passed successfully, @rtoy could it be merged or we're still waiting for some input?

rtoy:
I would still like someone else to review it too.  Still lgtm from me of
course.

On Wed, May 11, 2016 at 1:25 AM, Mikhail Pozdnyakov <
notifications@github.com> wrote:

> All checks have passed successfully, @rtoy <https://github.com/rtoy>
> could it be merged or we're still waiting for some input?
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/WebAudio/web-audio-api/pull/754#issuecomment-218394252>
>



-- 
Ray

padenot:
I just re-read everything, I think the text is good now.

PR #755, opened 2016-03-08
Fix #673: Add min/max attribute to AudioParam

### LINKED ISSUE #673: Add min/max read-only attributes for parameters in each node
As discussed in https://github.com/WebAudio/web-audio-api/pull/620#issuecomment-161724683, we need to add read-only attributes to make it easier to discover the nominal limits for the attributes of each node where we have describe limits.

This makes it easier to do some introspection for nominal ranges instead of having to look them up in the spec.

rtoy:
Puts back the readonly min and max attributes for an AudioParam to
make introspection of the actual nominal range of the AudioParam
possible.

This is a follow on of the discussion in #620.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2355,8 +2355,8 @@ function setupRoutingGraph() {
         <p>
           Each <code>AudioParam</code> includes <code>min</code> and
           <code>max</code> attributes that together form the <a>nominal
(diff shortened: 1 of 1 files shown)


padenot:
LGTM, just merge it after cleaning it up.

rtoy:
Thanks for the review

PR #770, opened 2016-03-23
Fix #769 by correcting the formulas

### LINKED ISSUE #769: Formulas for lowpass and highpass are incorrect
In https://webaudio.github.io/web-audio-api/#filters-characteristics, the formulas for the lowpass and highpass filters depend on the term `alpha_b`.  But the formula for `alpha_b` depends on `G`, the gain AudioParam of the filter.  This is wrong.  The lowpass and highpass filters depend on `Q`, not `G`.

The relevant code from Chrome (and similarly for other browsers because we all agree) is:

        // Compute biquad coefficients for lowpass filter
        resonance = std::max(0.0, resonance); // can't go negative
        double g = pow(10.0, 0.05 * resonance);
        double d = sqrt((4 - sqrt(16 - 16 / (g * g))) / 2);

The `resonance` term is related to `Q`.  I think a simple replacement of the `G` with the correct expression in the `alpha_b` equation will fix this.

rtoy:
The formulas for the lowpass and highpass filters are wrong; they
depend on Q, not G. Update the formulas to match what browsers
actually do.

Also updated the description of Q to say that Q is in dB for lowpass
and highpass filters and linear for all others.  Because that's what's
really happening.

### DIFF
--- a/index.html
+++ b/index.html
@@ -6578,7 +6578,8 @@ function calculateNormalizationScale(buffer)
           <dd>
             The <a href="https://en.wikipedia.org/wiki/Q_factor">Q</a> factor
             has a default value of 1. Its <a>nominal range</a> is (0,
-            \(\infty\)).
+            \(\infty\)). For the lowpass and highpass filters, <a>Q</a> is in
+            dB. For all other filter types, <a>Q</a> is a linear value.
           </dd>
           <dt>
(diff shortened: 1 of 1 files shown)


rtoy:
This is no longer in light of the resolution of #769 as wontfix

PR #775, opened 2016-03-28
Fix #348: Add latencyHint, replacing playbackCategory

### LINKED ISSUE #348: Need to provide hints to increase buffering for power consumption reasons

rtoy:
This updates and replaces the current `playbackCategory` to a
`latencyHint` that accepts an `AudioContextLatencyCategory` (a renaming of
`AudioContextPlaybackCategory` to reflect better the actual intent) or a
double for finer control of the latency.

An additional attribute, `processingLatency`, is added to the
`BaseAudioContext` to indicate what latency was actually used by the
context.


index.html
```diff
@@ -839,6 +839,25 @@ <h2 id="BaseAudioContext">
             <a>control thread</a>.
           </dd>
           <dt>
+            readonly attribute double processingLatency
```
rtoy: As discussed at the F2F, `processingLatency` is renamed to `baseLatency`

index.html
```diff
@@ -1563,11 +1583,23 @@ <h2 id="BaseAudioContext">
...
+            </p>
+            <p>
+              The preferred value of the <code>latencyHint</code> is a value
+              from <a>AudioContextLatencyCategory</a>. However, a double can
+              also be specified for the number of seconds of latency for finer
+              control to balance latency and power consumption. It is at the
+              browser's discretion to interpret the number appropriately. The
+              actual latency used is given by AudioContext's <a href=
+              "#widl-BaseAudioContext-baseLatency">baseLatency</a>
+              attribute.
```
padenot: The base latency does not include the system +hardware latency, and I think `latencyHint` can influence it, I think the last sentence is a bit confusing.

rtoy: Is the description of baseLatency in the AudioContext section enough?  Or should I add more text here as well?

PR #792, opened 2016-04-11
Fix #760 by widening the range to (-inf,inf)

### LINKED ISSUE #760: Nominal range for biquad gain parameter is too low
The nominal range for the biquad gain parameter is [-40,40].  This seems rather low.  We don't limit the possible values for the gain of a gain node, so I don't see why this should be restricted in any way.  Well, except that the biquad gain is in dB (see #759), so there is an actual upper limit to prevent overflow.  This limit is closer to 300 than 40.

rtoy:
Change the nominal range of the Biquad gain parameter to be any real
value.

### DIFF
--- a/index.html
+++ b/index.html
@@ -6590,7 +6590,7 @@ function calculateNormalizationScale(buffer)
           </dt>
           <dd>
             The gain has a default value of 0. Its <a>nominal range</a> is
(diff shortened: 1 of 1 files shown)


rtoy:
I'll run tidy on this after review.

PR #819, opened 2016-05-20
Fix #815: setValueCurve should throw on non-finite elements.

### LINKED ISSUE #815: setValueCurveAtTime should throw if the curve has non-finite values
We don't explicitly say it but invalid values such as NaN or Infinity in the curve for setValueCurveAtTime should throw an exception.  We don't allow such values in the AudioParam value setter or any of the other automation methods.

(Is there a WebIDL way to constrain the values of the elements of an array?  That would be ideal.)

rtoy:
If the curve contains non-finite values, setValueCurveAtTime must
signal a TypeError.  These values are illegal.

### DIFF
--- a/index.html
+++ b/index.html
@@ -3212,6 +3212,11 @@ function setupRoutingGraph() {
                   attribute is with a Float32Array that has a length less than
                   2.
                 </p>
+                <p>
+                  All elements of the value curve must be finite floating-point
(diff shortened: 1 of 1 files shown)


padenot:
We agreed to fix this by changing the WebIDL, closing without merging.

PR #825, opened 2016-05-25
Fix #817 by moving outputTimeStamp and outputLatency to context

### LINKED ISSUE #817: Property ownership of outputTimeStamp and outputLatency
Background:
Recently @pozdnyakov submitted the [intent to blink-dev](https://groups.google.com/a/chromium.org/forum/#!topic/blink-dev/-d2ewB7k3Zg) for the implementation.

First, I would like to note that I wholeheatedly welcome the idea, discussion, PR from outside of working group. In [the original PR](https://github.com/WebAudio/web-audio-api/pull/754), I expressed my concern against the location of these new properties being AudioDestinatioNode but the PR made its way into the spec.

The following is what's planned in the intent:
AudioDestinationNode.outputTimeStamp.contextTime
AudioDestinationNode.outputTimeStamp.performanceTime
AudioDestinationNode.outputLatency

Currently we have (or firmly decided):
AudioContext.currentTime
AudioContext.baseLatency

In short, we will have two different places to query the timing information. The argument for this design was 'AudioDestinatioNode should be treated as the device output'. However, I believe this is an interpretation based on Chromium's implementation - and the implementation should not dictate the API design. The audio destination node is the output of graph, and it does not represent the actual hardware device. Conceptually the destination node is a passive element just like any other node. It is not responsible for advancing time or calculating the expected latency.

On the other hand, we already placed the time information in the context, and this we cannot change. (it has been there since day one of Web Audio API). I believe the context is the right place to query the time/latency information.

By all means, I really appreciate @pozdnyakov's effort on this, so I hope we could solve this issue as soon as possible.

pozdnyakov:


pozdnyakov:
@hoch @rtoy @joeberkovitz PTAL

rtoy:
lgtm, assuming you did just move the sections around. I didn't read every single line to compare.

padenot:
This needs to be rebased, but looks good.

pozdnyakov:
Rebased and moved the attributes from `BaseAudioContext` to `AudioContext` (as was recommended at https://github.com/WebAudio/web-audio-api/issues/817#issuecomment-221655397 and https://github.com/WebAudio/web-audio-api/issues/817#issuecomment-221673914)

Issue #826, opened 2016-05-26
`orientationX` description is broken


padenot:
It reads:

> The orientationX, orientationY, orientationZ parameters represent a direction vector in 3D space. 

and does not say anything about k-rate vs. a-rate, default value, nominal ranges, etc.

rtoy:
Fixed in pr #827.

PR #827, opened 2016-05-26
Issue #826, fix markup so that the right text for `orientationX` appears

padenot:


index.html
```diff
@@ -5165,10 +5165,6 @@ <h2 id="AudioWorkerParamDescriptor">
             readonly attribute AudioParam orientationX
           </dt>
           <dd>
-            The <code>orientationX, orientationY, orientationZ</code>
```
rtoy: We should probably move this into a different section, just to make it clear what these are.  But that's a different PR.

PR #835, opened 2016-06-02
Fix #830 by converting 'outputTimeStamp' attribute to method

### LINKED ISSUE #830: Does outputTimeStamp return a new object every time?
https://webaudio.github.io/web-audio-api/#widl-AudioContext-outputTimeStamp

From the prose it's not clear if a new object is returned every time, or if the same attribute is returned every time. There's a similar situation with media elements where the HTML spec goes to some length to ensure that the attribute does not change while a script is running:
https://html.spec.whatwg.org/multipage/embedded-content.html#official-playback-position

Something similar would probably make outputTimeStamp less useful, so a method that returns the most recent data every time is probably better.

This problem seems to frequently come up on the web platform, see https://w3ctag.github.io/design-principles/#attributes-like-data

@domenic @travisleithead

pozdnyakov:
Whereas 'AudioContext.outputTimeStamp' attribute is returning a new object every time it should be converted to a method.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1728,21 +1728,21 @@ function setupRoutingGraph() {
             </p>
           </dd>
           <dt>
-            readonly attribute AudioTimeStamp outputTimeStamp
+            AudioTimestamp getOutputTimestamp()
           </dt>
           <dd>
             <p>
-              Holds an <a><code>AudioTimeStamp</code></a> instance containing
-              two correlated context's audio stream position values: the
-              <a href=
-              "#widl-AudioTimeStamp-contextTime"><code>contextTime</code></a>
+              Returns a new <a><code>AudioTimestamp</code></a> instance
+              containing two correlated context's audio stream position values:
+              the <a href=
+              "#widl-AudioTimestamp-contextTime"><code>contextTime</code></a>
               member contains the time of the sample frame which is currently
               being rendered by the audio output device (i.e., output audio
               stream position), in the same units and origin as context's
               <a href=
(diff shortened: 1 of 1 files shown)


pozdnyakov:
This patches proposes `getOutputTimestamp` name for the method (`get` prefix is added to be consistent with the existing getters https://webaudio.github.io/web-audio-api/#widl-AudioBuffer-getChannelData-Float32Array-unsigned-long-channel). @rtoy @foolip PTAL

index.html
```diff
@@ -1728,11 +1728,11 @@ <h2 id="AudioContext">
             </p>
           </dd>
           <dt>
-            readonly attribute AudioTimeStamp outputTimeStamp
+            AudioTimeStamp getOutputTimestamp()
           </dt>
           <dd>
             <p>
-              Holds an <a><code>AudioTimeStamp</code></a> instance containing
+              Returns an <a><code>AudioTimeStamp</code></a> instance containing
```
foolip: "Return a new `AudioTimeStamp` object ..." for clarity? Also might want to rename to `AudioTimestamp` for consistency.

foolip:
Looks good.

index.html
rtoy: I agree.

Other than these small nits, this looks good to me.

pozdnyakov: Done, thanks for looking!

rtoy:
lgtm.

pozdnyakov:
@padenot PTAL

PR #846, opened 2016-06-21
Define the term "render quantum" and link back to the definition.

### LINKED ISSUE #737: "Render quantum" is not defined
"Render quantum" is used in a few places like http://localhost:7100/#widl-OfflineAudioContext-startRendering-Promise-AudioBuffer, but it is not defined anywhere.

We probably want to define this term.

padenot:
This fixes #737.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1972,8 +1972,8 @@ function setupRoutingGraph() {
               rendering <code>length</code> sample-frames of audio into
               <var>buffer</var>.
               </li>
-              <li>For every render quantum, check and suspend the rendering if
-              necessary.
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -7781,7 +7781,8 @@ <h2 id="MediaStreamAudioSourceNode">
           Rendering an audio graph
         </h3>
         <p>
-          Rendering an audio graph is done in blocks of 128 samples-frames.
+          Rendering an audio graph is done in blocks of 128 samples-frames. A
```
rtoy: Shouldn't "samples-frames" be "sample-frames"?

PR #848, opened 2016-06-21
Spec how currentTime is updated.

### LINKED ISSUE #847: Spec how currentTime is updated
It's updated when a render quantum has been processed, and the change is immediately made visible to the main thread.

padenot:
This fixes #847.

### DIFF
--- a/index.html
+++ b/index.html
@@ -829,6 +829,11 @@ function setupRoutingGraph() {
               the earliest possible time when any change scheduled in the
               current state might take effect.
             </p>
+            <p>
+              <code>currentTime</code> MUST be read <a data-lt=
+              "atomic">atomically</a> on the control thread before being
+              returned.
+            </p>
           </dd>
           <dt>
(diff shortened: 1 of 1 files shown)


padenot:
This is a first cut, up for review. In particular, do we want to store the number of frames rendered, and divide in the `currentTime` getter, rather than adding `128 / sampleRate` each time?

rtoy:
Good question.  Chrome used to update currentTime by 128/sampleRate, but changed to counting time by render quantums.  This greatly reduced the number of issues caused by rounding.

I kind of think the spec can assume that currentTime + 128/sampleRate is done without floating-point rounding effects.

rtoy:
lgtm, if you want to land this as is.

rtoy:
Minor correction:  Chrome counts by frames, not rendering quanta.  But quanta has a nice side effect that counters can go 128 times bigger without overflowing.  (There's a bug on this on Windows with a 32-bit chrome.)

PR #851, opened 2016-06-22
Fix #330: analyser uses most recent samples

### LINKED ISSUE #330: Analyser FFT size vs processing block size
Audio processing required to process on blocks of 128 frames at a time.  The FFT size in the analyser node can be as small as 32. What does this actually mean and how does this work together?

Of the 128 samples in the current processing block do we take just the first 32 samples from the block?  Some other random set of 32?  

How does averaging work then?  Do we do all 4 sets of 32-point FFTs per processing block for averaging? Do we just do only 1 set of 32-point FFTs per processing block? This would have a strange effect of skipping over 3*32 samples of the signal.

For simplicity, I propose that the smallest FFT size be 128.

rtoy:
Specify that the Analyser node uses the most recent fftSize frames when computing the output.

### DIFF
--- a/index.html
+++ b/index.html
@@ -5789,10 +5789,11 @@ function calculateNormalizationScale(buffer)
             <p>
               Copies the <a>current frequency data</a> into the passed
               floating-point array. If the array has fewer elements than the
-              <a><code>frequencyBinCount</code></a>, the excess elements will
-              be dropped. If the array has more elements than the
-              <a><code>frequencyBinCount</code></a>, the excess elements will
-              be ignored.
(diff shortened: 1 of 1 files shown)


joeberkovitz:
LGTM. Is it possible that any of this language (e.g. "most recent `fftSize` frames" may change in light of what happens with #738)?

rtoy:
I think "most recent" still applies.  What the actual "most recent" samples are will depend on exactly when the calls are done, but back-to-back calls actually happen just before and just after the currentTime is updated will get different sets of data.  I think that's still correct for "most recent".

PR #855, opened 2016-06-23
Fix #30: allow disabling resampling for decodeAudioData

### LINKED ISSUE #30: decodeAudioData: option to disable automatic resampling to context rate
> Originally reported on W3C Bugzilla [ISSUE-21519](https://www.w3.org/Bugs/Public/show_bug.cgi?id=21519) Tue, 02 Apr 2013 14:41:41 GMT
> Assigned to 

Add  an optional 4th argument for decodeAudioData that disables automatic sample rate conversion.

rtoy:
Add yet another optional dictionary parameter to decodeAudioData to
allow disabling the resampling for decodeAudioData.

Defines a new dictionary for these options.

Links and processing model updated appropriately.

### DIFF
--- a/index.html
+++ b/index.html
@@ -733,7 +733,7 @@ function setupRoutingGraph() {
             still be created, through <a href=
             "#widl-BaseAudioContext-createBuffer-AudioBuffer-unsigned-long-numberOfChannels-unsigned-long-length-float-sampleRate">
             createBuffer</a> or <a href=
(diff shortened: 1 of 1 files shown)


joeberkovitz:
@rtoy I assume we'd put this off given the recent WG resolution to table temporary fixes to `decodeAudioData`. Perhaps the best sort of band-aid solution is to tack on yet another optional argument to the current method?

rtoy:
Yes, after last week's teleconf, I assumed this would be postponed to v.next.  If we do want to continue with this (and the progress callback), another optional argument would be ok.  This argument would be a dictionary.  

Perhaps this needs to be discussed further by the WG?

rtoy:
Closing this pull request; we'll revisit this later in a future design of `decodeAudioData`

PR #856, opened 2016-06-23
Fix #824: add OfflineAudioDestinationNode

### LINKED ISSUE #824: OfflineAudioContext destination channelCount and channelCountMode changes?
The channelCount and channelCountMode are normally writable.  But what about the AudioDestinationNode for an OfflineAudioContext? 

I don't know how to make sense to create an offline context of, say, 7 channels, and then change the channelCount to 3.  We've already created a 7-channel buffer to hold the result, so what does it mean to have a channelCount of 3?

Also, the default channelCountMode is "explicit" for the destination.  What if I change to "max"?  Using the above example, if I now feed in a 10-channel source, the "max" setting kind of implies that output will have 10 channels.  But the buffer has already been created.

For simplicity, I propose that channelCount and channelCountMode cannot be changed for an offline context.  I think it's fine to allow the user to change the channelnterpretation to get the desired mixing effects (or not).

rtoy:
Add OfflineAudioDestinationNode so we can specify that the channel
count and channel mode cannot be changed after construction.  Also,
explicitly state that the channel count is the number of channels
specified by the constructor.


rtoy:
PTAL.

This basically defines the offline destination node. 

I'm not sure, but think to do this right, we need to change BaseAudioContext so that the destination attribute is a BaseAudioDestinationNode and then define that AudioDestinationNode and OfflineAudioDestinationNode are subclasses.  We can't have OfflineAudioDestinationNode be a subclass of AudioDestinationNode because the latter has a maxChannels attribute that doesn't make sense for an OfflineAudioDestinationNode.

rtoy:
Well, maybe for simplicity (?) we could have maxChannelCount be equal to channelCount.  Then OfflineAudioDestinationNode could be a subclass of AudioDestinationNode.

joeberkovitz:
@rtoy I do not think we even need a new subclass here. We can instead specify that 1) maxChannelCount is always channelCount, and 2) the AudioDestinationNode for an OfflineAudioContext throws NotSupportedException on any attempt to modify its channelCount.

rtoy:
I'm ok with either approach.

It was suggested by @hoch in https://github.com/WebAudio/web-audio-api/issues/824#issuecomment-221414938 and agreed upon by @padenot to create an `OfflineAudioDestinationNode`. I have a slight preference to make `OfflineAudioDestinationNode` a subclass of `AudioDestinationNode` and making `maxChannels` be set to number of channels used to create the offline context.

joeberkovitz:
Your simplified approach to an `OfflineAudioDestinationNode` subclass is likewise fine with me.

joeberkovitz:
This looks good to me. I'm not sure if it's ready for merge...?

rtoy:
No, it's not yet ready to merge because the current text doesn't define `OfflineAudioDestinationNode` as a subclass of `AudioDestinationNode`.  Plus, need to clarify `maxChannels` for `OfflineAudioDestinationNode`.

rtoy:
If #968 is implemented, then there isn't really a need for a separate OfflineAudioDestination.

rtoy:
Closing this.  The restriction on the channel count and mode will be handled when #968 is fixed.

PR #863, opened 2016-06-28
Fix #697: AudioNode channel options dictionary

### LINKED ISSUE #697: Property bag for channelCount
Should all constructors allow channelCount, channelCountMode, and channelIntepretation properties in the constructors?

Probably.  Presumably they're ignored (like AudioBufferSource) or throw if you set them to invalid values.

#696

rtoy:
Add the AudioNodeChannelOptions dictionary that can be inherited by
most dictionaries for the AudioNode constructors.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2733,6 +2733,29 @@ function setupRoutingGraph() {
           </dd>
         </dl>
         <section>
+          <h2>Dictionaries</h2>
+          <dl title="dictionary AudioNodeChannelOptions" class="idl">
+            <dt>
+              unsigned long channelCount
+            </dt>
+            <dd>
+              Desired number of channels for the <a>channelCount</a> attribute.
+            </dd>
+            <dt>
+              ChannelCountMode channelCountMode
+            </dt>
+            <dd>
+              Desired mode for the <a>channelCountMode</a> attribute.
+            </dd>
+            <dt>
+              ChannelIntepretation channelInterpretation
+            </dt>
+            <dd>
+              Desired mode for the <a>channelCountMode</a> attribute.
+            </dd>
+          </dl>
+        </section>
+        <section>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2733,6 +2733,29 @@ <h2 id="OfflineAudioCompletionEvent">
           </dd>
         </dl>
         <section>
+          <h2>Dictionaries</h2>
+          <dl title="dictionary AudioNodeChannelOptions" class="idl">
```
rtoy: Maybe the dictionary name should just be `AudioNodeOptions` instead of `AudioNodeChannelOptions`?

index.html
```diff
@@ -2733,6 +2733,31 @@ <h2 id="OfflineAudioCompletionEvent">
           </dd>
         </dl>
         <section>
+          <h2>
+            Dictionaries
+          </h2>
+          <dl title="dictionary AudioNodeChannelOptions" class="idl">
```
rtoy: I think this should be called `AudioNodeOptions`. There was no need to define an additional dictionary for AudioNodes for the number of inputs and outputs because these are never settable for any node except for the merger, splitter, and script processor.  I thought it was better just to add the appropriate input/output number separately for each of these nodes.

joeberkovitz:
@rtoy with the change to `AudioNodeOptions`, LGTM. However shouldn't this be the base for the proposed `AudioWorkletNodeOptions` in #869?

joeberkovitz:
After reviewing a number of the constructor PRs I feel strongly that `AudioNodeOptions` is the right name in order to futureproof against changes that introduce non-channel-related options for every `AudioNode`.

rtoy:
Yes, I think AudioNodeOptions should probably be the base dictionary for AudioWorkletNodeOptions.

PR #865, opened 2016-06-28
Fix #700: Add constructor for ChannelMergerNode and ChannelSplitterNode

### LINKED ISSUE #700: Property bag for ChannelSplitter/ChannelMerger
These nodes have no attributes, but the constructors need to specify the number of inputs/outputs.  So the property bag for the constructor needs numberOfOutputs(?) for splitter and numberOfInputs for merger.

#696

rtoy:
Fix #700.

Add constructors and define the options dictionary for each
constructor.

### DIFF
--- a/index.html
+++ b/index.html
@@ -6397,7 +6397,20 @@ function calculateNormalizationScale(buffer)
           "matrix mixing" where individual gain control of each channel is
           desired.
         </p>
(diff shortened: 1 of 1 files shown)


joeberkovitz:
See #863 re question on name of options superclass.

joeberkovitz:
Otherwise LGTM

PR #866, opened 2016-06-28
Fix #701: Add constructor for IIRFilterNode

### LINKED ISSUE #701: Property bag for IIRFilter
The IIRFilter node has no attributes.  To create one, the feedforward and feedback coefficients need to be specified, so these should be the names of the properties.  (These are the names of the parameters for createIIRFilter.)

#696

rtoy:
The appropriate dictionary for the constructor is also defined.


joeberkovitz:
LGTM, except see #863 re question on name of options superclass.

PR #867, opened 2016-06-28
Fix #702: Add constructor for OscillatorNode.

### LINKED ISSUE #702: Property bag for Oscillator
Should the Oscillator node allow a periodicWave property to set the wave on construction?  What if the user specifies a type and periodicWave?  What should happen?

#696

rtoy:
The appropriate dictionary for the constructor is also defined.

### DIFF
--- a/index.html
+++ b/index.html
@@ -7556,7 +7556,9 @@ $$
             A custom periodic wave
           </dd>
         </dl>
-        <dl title="[Constructor(BaseAudioContext context, optional OscillatorOptions options)] interface OscillatorNode : AudioNode" class="idl">
+        <dl title=
+        "[Constructor(BaseAudioContext context, optional OscillatorOptions options)] interface OscillatorNode : AudioNode"
+        class="idl">
           <dt>
             attribute OscillatorType type
           </dt>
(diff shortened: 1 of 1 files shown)


joeberkovitz:
pending checks success, LGTM, except see #863 re question on name of options superclass.

index.html
```diff
@@ -7643,6 +7643,51 @@ <h2 id="WaveShaperNode">
...
           <h2>
+            OscillatorOptions
+          </h2>
+          <p>
+            This specifies the options to be used when constructing an
+            <a><code>OscillatorNode</code></a>. All of the members are optional;
+            if not specified, the normal default values are used for
+            constructing the oscillator.
+          </p>
+          <dl title="dictionary OscillatorOptions : AudioNodeChannelOptions"
```
rtoy: Does it make sense for `OscillatorOptions` to derive from `AudioNodeOptions`?  The channel stuff really only applies to inputs and this node has no inputs at all. Setting them will change the values for the node, as they do today, but they have no effect on the node itself.

Leave it for a kind of backward compatibility?

PR #868, opened 2016-06-28
Fix #703: Add constructor for PeriodicWave.

### LINKED ISSUE #703: Property bag for PeriodicWave
A PeriodicWave has no attributes but createPeriodicWave requires two arrays and an optional constraint.  So perhaps the properties are real, imag (from the parameter names for createPeriodicWave), and disableNormalization (to match the optional constraint).

#696

rtoy:
Define the dictionary too.


joeberkovitz:
LGTM

PR #884, opened 2016-06-30
Fix a couple of WebIDL bugs

dontcallmedom:
I'm not sure if the "channelCount array" is really supposed to be an array of positive integers (see 2nd commit).

### DIFF
--- a/index.html
+++ b/index.html
@@ -5079,7 +5079,7 @@ the event handler.
             </p>
           </dd>
           <dt>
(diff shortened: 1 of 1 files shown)


rtoy:
Thank you very much for your contribution.  However, `AudioWorker` is being replaced by `AudioWorkerlet` (see pr #869), at which point the AudioWorker section will be removed.

Issue #543, opened 2015-06-09
Expose arrays in a PeriodicWave object that were used to create it?


rtoy:
Currently, when you call createPeriodicWave(real, imag), the returned object is opaque.  Wouldn't it be nice to be able to at least inspect the real and imag arrays that were used to create the object?  This would be readonly of course, and immutable.

PR #890, opened 2016-07-06
Fix #875: Construct DelayNode

### LINKED ISSUE #875: Property bag for Delay
An addition to #696.

Constructor:

```
Constructor(BaseAudioContext context, optional DelayOptions options)
```

Dictionary:
```
dictionary DelayOptions : AudioNodeChannelOptions {
  double maxDelayTime;
  double delayTime;
}
```

rtoy:
Add constructor and define the options dictionary.

### DIFF
--- a/index.html
+++ b/index.html
@@ -3646,7 +3646,9 @@ param.setValueCurveAtTime(curve, t7, t8 - t7);
           delay-line mixing takes place using the single prevailing channel
           layout.
         </p>
-        <dl title="[Construct(BaseAudioContext context, optional DelayOptions options)]interface DelayNode : AudioNode" class="idl">
+        <dl title=
+        "[Construct(BaseAudioContext context, optional DelayOptions options)]interface DelayNode : AudioNode"
+        class="idl">
           <dt>
             readonly attribute AudioParam delayTime
           </dt>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -3676,6 +3678,31 @@ <h2 id="DelayNode">
...
+          <a><code>DelayNode</code></a>. All members are optional; if not
+          given, the node is constructed using the normal defaults.
+        </p>
+        <dl title="dictionary DelayOptions : AudioNodeChannelOptions" class=
+        "idl">
+          <dt>
+            double maxDelayTime
+          </dt>
+          <dd>
+            The maximum delay time for the node.
```
joeberkovitz: This is an optional parameter with a default of 1.0 in the existing `createDelay()` method in AudioContext. The default should be specified here.

joeberkovitz:
@rtoy This PR leads to a more question: when we add options for `createXXX()` parameters that are _not_ exposed by a node attribute, should they be? For instance, why is `maxDelayTime` not at least a readonly attribute of `DelayNode`?

rtoy:
As you know, I'm all for introspection so I would like `maxDelayTime` to be available in a `DelayNode`. I can add that in this PR, but should we discuss this in the WG?

rtoy:
Other nodes that have parameters that aren't exposed by the node:

- `PeriodicWave` has real and imaginary parts in the constructor but are not exposed by the node. See issue #543.
- `IIRFilter` has the feedforward and feedback coefficients in the constructor, but these are not exposed by the node.  This is useful for introspection.
- `Oscillator` can take a `PeriodicWave` object but this is not exposed in any way.  Not exactly sure how to expose that except to return a copy of the `PeriodicWave` object used to create the oscillator.

I think these are the only nodes that don't expose things used in the construction of the node.

joeberkovitz:
I'd say in the interest of expediency let's just merge this and deal with improved introspection across the API as a separate issue.

rtoy:
Works for me.

PR #896, opened 2016-07-07
Fix #880: Construct WaveShaper

### LINKED ISSUE #880: Property bag for WaveShaper
An addition to #696.

Constructor:
```
Constructor(BaseAudioContext context, optional WaveShaperOptions options)
```
Dictionary:
```
dictionary WaveShaperOptions : AudioNodeChannelOptions {
  Float32Array curve;
  OverSampleType oversample;
}
```

While we're at it can we change `curve` to be `sequence<float>` instead of `Float32Array`?

rtoy:
Add constructor and define dictionary of options.

### DIFF
--- a/index.html
+++ b/index.html
@@ -7474,8 +7474,8 @@ $$
             <a><code>WaveShaperNode</code></a>. All members are optional; if
             not specified, the normal default is used in constructing the node.
           </p>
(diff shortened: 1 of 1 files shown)


joeberkovitz:
LGTM modulo naming of `AudioNodeOptions` superclass.

PR #902, opened 2016-07-07
Fix #367: Define ConstantSourceNode

### LINKED ISSUE #367: Constructible and connect()able AudioParams
Binding AudioParam of another AudioNode to an AudioWorker

As far as i understand, it is impossible to change a native AudioNodes' parameter at real-time. Some use cases i can think of:
1. Changing parameters of AudioNodes with respect to audio input/output signals (e.g. side-chaining)
2. Changing parameters of AudioNodes with respect to some internal real-time algorithm.
3. Changing parameters of AudioNodes with respect to MIDI messages at minimum latency possible.
  (also resolving following issue https://github.com/WebAudio/web-midi-api/issues/99)

A solution can be achieved by making AudioParams bindable to AudioWorkers.

Possible WebIDL can be something like:
```
interface AudioWorkerNode : AudioNode {
    ...
    void  bindParameter(DOMString name, AudioParam parameter);
    ...
};
```

Example main file javascript:
```
envelope = context.createGain();
worker.bindParameter('envelopeGain', envelope.gain);
```

Example worker code:
```
onaudioprocess= function (e) {
  e.parameters.envelopeGain.value = calculateValue();
}
```

Since AudioWorkers are processed sequentially in audio processing graph, i think changing AudioParam in a AudioWorker's context won't be a problem in terms of concurrency.

rtoy:
Defines a constant source node with one audio param defining the
output value of the source node.

This node can be used in place of a constructible AudioParam.


index.html
```diff
@@ -1321,6 +1321,14 @@ <h2 id="BaseAudioContext">
             </p>
           </dd>
           <dt>
+            ConstantSourceNode createConstantSource()
```
joeberkovitz: Are we going to add more `createXXX()` methods for future nodes? I thought these factory methods were essentially deprecated, although no harm in adding them - it depends on whether we want to present a complete set of factory methods along with constructors in V1.

index.html
```diff
@@ -4208,6 +4216,122 @@ <h3 id="looping-AudioBufferSourceNode">
...
+          </dd>
+          <dt>
+            void start(optional double when = 0)
+          </dt>
+          <dd>
+            <p>
+              Schedules the constant source to playback at an exact time. This
+              is defined exactly the same as for the <a href=
+              "#widl-AudioBufferSourceNode-start-void-double-when-double-offset-double-duration">
+              <code>start()</code></a> method for an
```
joeberkovitz: Is it time to refactor the definitions of `start()`, `stop()` and `onended` into something like a `SchedulableSourceNode`? Having ABSN carry water in the spec for these other nodes feels a little odd.

I also noticed that `OscillatorNode` lacks the `onended` event handler. If this is a mistake, it argues for the refactoring even more strongly.

rtoy: Heh. Chrome internally has an AudioScheduledSourceNode class that implements most of the ABSN and OscillatorNode.  I'm fine with adding a `ScheduledSourceNode` (or whatever name we choose),  Just have to make sure that it's not constructible.

Also, is your spec out of date?  I see [OscillatorNode.onended](https://webaudio.github.io/web-audio-api/#widl-OscillatorNode-onended).

index.html
rtoy: Yeah, I did that for consistency since we have, of course, factory methods for all other nodes. I'm ok with leaving this out.

index.html
rtoy: I don't want to do the refactoring in this PR.  Do you want a new class before or after this PR?

joeberkovitz: After is fine!

rtoy:
Anything else to do? Assuming we keep the factory method.

index.html
```diff
@@ -1342,6 +1342,35 @@ <h2 id="BaseAudioContext">
...
           <dt>
+            ConstantSourceNode createConstantSource()
+          </dt>
+          <dd>
+            <p>
+              Creates a <a><code>ConstantSourceNode</code></a>.
+            </p>
+          </dd>
+          <dt>
+            Promise&lt;AudioWorker&gt; createAudioWorker()
```
padenot: This should be removed from this PR.

index.html
```diff
@@ -4409,6 +4438,122 @@ <h3 id="looping-AudioBufferSourceNode">
         </section>
       </section>
       <section>
+        <h2 id="ConstantSourceNode">
+          The ConstantSourceNode Interface
```
padenot: This should be added in the list summary of nodes at the top of the document.

index.html
```diff
@@ -4409,6 +4438,122 @@ <h3 id="looping-AudioBufferSourceNode">
...
+        </p>
+        <pre>
+  numberOfInputs  : 0
+  numberOfOutputs : 1
+        </pre>
+        <dl title=
+        "[Constructor(BaseAudioContext context, optional ConstantSourceOptions options)]interface ConstantSourceNode : AudioNode"
+        class="idl">
+          <dt>
+            readonly attribute AudioParam sourceValue
```
padenot: Why not just `value` ?

index.html
```diff
@@ -4409,6 +4438,122 @@ <h3 id="looping-AudioBufferSourceNode">
...
+          </dt>
+          <dd>
+            <p>
+              The constant value of the source. Its default value is 0. This
```
padenot: I think 0 makes sense, but have we considered 1 ?

index.html
```diff
@@ -4409,6 +4438,122 @@ <h3 id="looping-AudioBufferSourceNode">
...
+          </h2>
+          <p>
+            This specifies options for constructing a
+            <a><code>ConstantSourceNode</code></a>. All members are optional;
+            if not specified, the normal defaults are used for constructing the
+            node.
+          </p>
+          <dl title="dictionary ConstantSourceOptions" class="idl">
+            <dt>
+              float sourceValue
```
padenot: `required` here, so that we don't forget it when fixing the other occurrence.

index.html
rtoy: I think that was suggested in a teleconf, but I'm not sure. I prefer `value`; `sourceValue` doesn't really add much value to me. ;-)

index.html
rtoy: I think both are good defaults; 1 is better if you think the node is a constant source; 0 is slightly better if you think it's an audioparam that might be connected to some source.

index.html
joeberkovitz: I believe it was me who said that referencing `node.value.value` felt weird and potentially confusing. I think I suggested `source` originally as the name of the AudioParam.

index.html
rtoy: As noted in #367, yes, we're keeping the factory method.

index.html
rtoy: Per #367, this is now named `offset`

index.html
rtoy: Per #367, the default is 1.

index.html
rtoy: Do we really want to require this? The default is 1, so this wouldn't be required.

padenot:
LGTM after tidy up.

PR #911, opened 2016-07-26
Fix #909: Clarify setPosition/setOrientation text

### LINKED ISSUE #909: Clarify setPosition/setOrientation text
In https://webaudio.github.io/web-audio-api/#widl-PannerNode-setOrientation-void-float-x-float-y-float-z, we say that `orientationX` is set.  But `orientationX` is a read-only AudioParam.  We really want to say that `orientationX.value` is set.  Same hold for `setPosition`.

rtoy:
Clarify the text for setPosition/setOrientation to say that we're
setting the values of the corresponding AudioParam's, not the
AudioParam itself.

### DIFF
--- a/index.html
+++ b/index.html
@@ -5723,8 +5723,10 @@ the event handler.
           <dd>
             <p>
               This method is DEPRECATED. It is equivalent to setting
-              <code>positionX</code>, <code>positionY</code>, and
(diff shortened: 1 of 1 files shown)


jpmedley:
Woot!

PR #917, opened 2016-07-28
Fix #901: Remove incorrect value setter clause.

### LINKED ISSUE #901: Value setter description in Computation of Value
In https://webaudio.github.io/web-audio-api/#computation-of-value, bullet item 1 for the `ComputedValue` says

> If the value attribute is set after any automation events have been scheduled, then these events will be removed. 

I don't think this is what anyone does nor is it what we really want.  The value setter has never removed events from the time line.  This is also inconsistent with our desire to make the value setter equivalent to a call to `setValueAtTime` which doesn't remove any events.

I think this sentence should be removed.

rtoy:
Setting the value attribute shouldn't remove events from the timeline;
nobody does that and this conflicts with the idea that the setter is
the same as calling setValueAtTime().

### DIFF
--- a/index.html
+++ b/index.html
@@ -3473,13 +3473,13 @@ function setupRoutingGraph() {
             each time, which is either the value set directly to the
             <code>value</code> attribute, or, if there are any scheduled
             parameter changes (automation events) with times before or at this
-            time, the value as calculated from these events.  When read,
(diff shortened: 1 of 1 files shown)


joeberkovitz:
LGTM.

PR #921, opened 2016-08-02
Fix #870: Construct AudioBuffer

### LINKED ISSUE #870: Property bag for AudioBuffer
An addition for #696.  This isn't an AudioNode, but I think we wanted a constructor for this.

The constructor for AudioBuffer is something like

    Constructor(BaseAudioContext context, AudioBufferOptions options)

and the dictionary should be something like

    dictionary AudioBufferOptions {
      unsigned long numberOfChannels;
      unsigned long length;
      float sampleRate;
    };

`numberOfChannels` and `length` are the names of the `AudioBuffer` attributes and are required to be specified.  It would be convenient if `sampleRate` were not required and have it default to the sample rate of the context.

Or maybe now is the time to allow creating an AudioBuffer that doesn't need a context object.  In that case, `sampleRate` would be required.

rtoy:
Add constructor for AudioBuffer and define an appropriate dictionary.  The dictionary is required and so are all the members.

### DIFF
--- a/index.html
+++ b/index.html
@@ -3739,7 +3739,9 @@ param.setValueCurveAtTime(curve, t7, t8 - t7);
           <a><code>OfflineAudioContext</code></a> and an
           <a><code>AudioContext</code></a>.
         </p>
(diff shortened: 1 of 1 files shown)


joeberkovitz:
LGTM

PR #927, opened 2016-08-04
Fix #926: Link to nominal range, and rephrase other uses

### LINKED ISSUE #926: Usage of nominal range in AudioBuffer is somewhat confusing.
In https://webaudio.github.io/web-audio-api/#AudioBuffer, there is the sentence

>  Its format is non-interleaved IEEE 32-bit linear PCM with a nominal range of -1 -> +1.

This makes it look kind of like "nominal range" as applied to `AudioParam`s and `AudioNode` attributes.  This is clearly not case here.  

We should probably rephrase this not to use the term "nominal range", and probably add markup throughout the document to highlight (either with a style or, better, a link) to the definition of nominal range.

Oh, and "IEEE" should be replaced with "IEEE-754" to make it clear that it's floating point. Or maybe just even replace "IEEE" with "floating-point".  We don't really care what format it is in except that it is floating point. Or say it's of type float (webidl). Or something.

rtoy:
Some parts where missing links to nominal range, so add them.  In some
places the use of nominal range could be confused with the defined
term, so rephrase it to make it less confusing.


index.html
```diff
@@ -3755,11 +3755,12 @@ <h2 id="AudioBuffer">
         <p>
           This interface represents a memory-resident audio asset (for one-shot
           sounds and other short audio clips). Its format is non-interleaved
-          IEEE 32-bit linear PCM with a nominal range of -1 -&gt; +1. It can
-          contain one or more channels. Typically, it would be expected that
-          the length of the PCM data would be fairly short (usually somewhat
-          less than a minute). For longer sounds, such as music soundtracks,
-          streaming should be used with the <code>audio</code> element and
+          32-bit linear floating-point PCM values with a normal range of \([-1,
+          1]\), but values are not limited to this range.. It can contain one
```
joeberkovitz: double periods here

joeberkovitz:
apart from typos, lgtm

index.html
rtoy: Fixed.

PR #928, opened 2016-08-04
Fix #925: Add graph for nominal range clipping

### LINKED ISSUE #925: Nominal range example should have a graph
The [example](https://webaudio.github.io/web-audio-api/#computation-of-value) for the behavior of the nominal range really should have a graph to illustrate the desired result.  It's a bit hard to tell from the text exactly what is supposed to happen.

rtoy:
Add a graph illustrating how clipping to nominal range works for
AudioParams.


joeberkovitz:
LGTM

PR #935, opened 2016-08-15
Fix #414: Add tail-time note for each node.

### LINKED ISSUE #414: Tail-time reference should be noted for Convolver, Delay, AudioWorker
The *Lifetime* section of the spec http://webaudio.github.io/web-audio-api/#lifetime-1 explains a **tail-time reference** that pins a node in the graph while it has pending, audible internal state. ConvolverNode is mentioned explicitly as an example of a node with a tail time, and it's stated that descriptions of nodes elsewhere in the document will mention a tail-time reference if it's applicable to that node.

1. The ConvolverNode interface makes no mention of tail-time reference. Presumably it should, since this is the one node that has it for certain.

2. The AudioWorkerNode should probably documented as having a tail-time reference as long as it has an active onaudioprocess handler, since it can act as a source and has an internal state unknown to the audio engine. It could be emitting a signal at any given time.

3. Shouldn't DelayNode have a tail-time as well?

rtoy:
For each node, make a note if the node has a tail-time reference if
the node produces non-zero output for silent input.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2788,7 +2788,7 @@ function setupRoutingGraph() {
             <li>A <em>connection</em> reference which occurs if another
             <a><code>AudioNode</code></a> is connected to it.
             </li>
-            <li>A <em>tail-time</em> reference which an
+            <li>A <em><dfn>tail-time</dfn></em> reference which an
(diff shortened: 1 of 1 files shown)


joeberkovitz:
LGTM

PR #941, opened 2016-08-23
chore: performance improvements

marcoscaceres:
compressing images with imageOptim: "Saved 522.6KB out of 964.6KB. 56.8% per file on average (up to 70.7%)"


marcoscaceres:
Sent at PR to HTML5 tidy to add support for `as` https://github.com/htacg/tidy-html5/pull/450

rtoy:
Thanks for these optimizations.

One question: did you manually run imageOptim on each file?  Did you specify any options? (I haven't tried imageOptim yet). It would be ideal if this could be automated.  Someone (me) is going to forget if an image needs to be updated.  (The automation image needs to be updated someday because it's missing some examples.)

marcoscaceres:
On August 24, 2016 at 1:08:42 AM, rtoy (notifications@github.com) wrote:
> Thanks for these optimizations.
>
> One question: did you manually run imageOptim on each file?

Yes. I used the GUI-based version of the app.

>  Did you specify any options? (I haven't tried imageOptim yet).

yes, I enabled a few things (too many to list here, unfortunately).

> It would be ideal if this could be automated. Someone
> (me) is going to forget if an image needs to be updated. (The automation image needs to
> be updated someday because it's missing some examples.)

Agree, something like that would be nice - though images rarely
change, so it might not be worth setting up some automation.

Options are (Mac Only):
https://imageoptim.com/command-line.html

Could also use something like imageMagik.

Having said that, current PR already saves 500Kb. It's already a
pretty big win, even if new images don't get compressed.

marcoscaceres:
Oh, I accidentally stumbled on this ... which could maybe run on Travis CI (or as a git hook or whatever... not tried it, but keeping it here as a reminder):

https://github.com/ImageOptim/node-imageoptim-api

rtoy:
Thanks for the info.  Could you run tidy on the file?

marcoscaceres:
> Thanks for the info. Could you run tidy on the file?

done.

rtoy:
Thanks so much for these optimizations!

lgtm.

PR #945, opened 2016-08-25
Polishing AudioWorkletGlobalScope section

hoch:
Preview: https://hoch.github.io/web-audio-api/#AudioWorkletGlobalScope

- Added an algorithm for `registerProcessor()` method in the ECMA script style.
- Moved the AWGS section after the concept, because this is the first phase of using AudioWorklet.
- Detailed the description and added an informative note.


hoch:
@bfgeek Could you take a look at the refined sections? (The concepts + AudioWorkletGlobalScope)

index.html
```diff
@@ -5164,6 +5164,114 @@ <h2 id="AudioWorklet-concepts">
...
+                  DOMString name
+                </dt>
+                <dd>
+                  A string key that represents a class definition to be
+                  registered. This key is used to look up the constructor of
+                  <a>AudioWorkletProcessor</a> during construction of an
+                  <a>AudioWorkletNode</a>.
+                </dd>
+                <dt>
+                  VoidFunction audioWorkletProcessorCtor
```
joeberkovitz: Call this arg `processorCtor` to reconcile it with the above algorithmic description

index.html
```diff
@@ -5164,6 +5164,114 @@ <h2 id="AudioWorklet-concepts">
...
+                  P="parameterDescriptors")</code> is not an array or
+                  <code>undefined</code>, throw a <code>TypeError</code> and
+                  abort all these steps.
+                </li>
+                <li>Let definition be a new <a>AudioWorkletProcessor</a>
+                definition with:
+                  <ul>
+                    <li>node name being <em>name</em>
+                    </li>
+                    <li>processor class constructor being <em>constructor</em>
```
joeberkovitz: I believe you meant `<em>processorCtor</em>` here.

hoch: Ah, also I made mistakes at line 5233 and 5243. Will fix them.

hoch:
Any opinion @rtoy and @padenot?

index.html
```diff
@@ -5164,6 +5164,116 @@ <h2 id="AudioWorklet-concepts">
...
+          </h2>
+          <p>
+            This special execution context is designed to enable the
+            generation, processing, and analysis of audio data directly using
+            JavaScript in the audio <a>rendering thread</a>. The user-supplied
+            script code is evaluated in this scope to define an
+            <a>AudioWorkletProcessor</a>.
+          </p>
+          <p>
+            <a>AudioWorkletGlobalScope</a> has a <dfn>node name to processor
```
rtoy: You mark "node name to processor definition map" as a definition, but I don't see how this is defined in any way here. I think you need to expand this somehow if you really want to define this here.

index.html
```diff
@@ -5164,6 +5164,116 @@ <h2 id="AudioWorklet-concepts">
...
+              void registerProcessor ()
+            </dt>
+            <dd>
+              <p>
+                Registers a class definition derived from
+                <a>AudioWorkletProcessor</a>.
+              </p>
+              <p>
+                When the <a><code>registerProcessor(<em>name</em>,
+                <em>processorCtor</em>)</code></a> method is called, the user
```
rtoy: "processorCtor" or "processorConstructor"?

index.html
```diff
@@ -5164,6 +5164,116 @@ <h2 id="AudioWorklet-concepts">
...
+                agent must run the following steps:
+              </p>
+              <ol>
+                <li>If the <code><em>name</em></code> exists as a key in the
+                <a>node name to processor definition map</a>, throw a
+                <code>NotSupportedError</code> and abort all these steps.
```
rtoy: I think you should explain why an error is thrown.  I assume you're disallowing redefinition.  If so, isn't redefinition a natural part of Javascript?

joeberkovitz: I agree with the conclusion. In my mind the analogy is to setting a new value for an existing key in any arbitrary map -- why would this be disallowed for any other map?  It might be that there is some intention to override an existing definition and although this might be awkward today, I don't think we can assume that there isn't some purpose to registering the same name twice.

hoch: The reason behind this, this method can get called anytime. So if we allow the dynamic redefinition of the processor, this becomes the data race. For example, the processor A('foo') is rendering the audio and its definition gets overwritten by the processor B('foo').

I'd say, we block the redefinition for V1. This is what Worklet spec does at the moment. This explanation can be a part of the spec as an informative section, if you will.

joeberkovitz: Thanks @hoch, I see now. That feels reasonable to me.

index.html
hoch: I expanded it. However, `dfn` here does not work well. It still complains about a linkless `a` on `node name to processor definition map`.

hoch: I've added a short description about it.

index.html
```diff
@@ -5193,24 +5195,25 @@ <h2 id="AudioWorkletGlobalScope">
...
                 When the <a><code>registerProcessor(<em>name</em>,
-                <em>processorCtor</em>)</code></a> method is called, the user
+                <em>processorConstructor</em>)</code></a> method is called, the user
                 agent must run the following steps:
               </p>
               <ol>
                 <li>If the <code><em>name</em></code> exists as a key in the
                 <a>node name to processor definition map</a>, throw a
-                <code>NotSupportedError</code> and abort all these steps.
+                <code>NotSupportedError</code> and abort all these steps because
```
rtoy: NotSupportedError exception

index.html
```diff
@@ -5176,7 +5176,9 @@ <h2 id="AudioWorkletGlobalScope">
           </p>
           <p>
             <a>AudioWorkletGlobalScope</a> has a <dfn>node name to processor
-            definition map</dfn>. Initially it is empty and populated when
+            definition map</dfn>. This map stores definitions of
```
rtoy: Maybe say somewhere here that the map is internal and is not directly exposed to the user?

hoch:
@joeberkovitz @padenot Is there any more to do here?

rtoy:
lgtm

PR #957, opened 2016-09-06
Remove AudioWorker from spec

### LINKED ISSUE #949: Remove AudioWorker section
Now that the AudioWorklet section has landed, we should remove the old AudioWorker section completely.

joeberkovitz:
Fixes #949

### DIFF
--- a/index.html
+++ b/index.html
@@ -186,7 +186,7 @@
         objects are connected together to define the overall audio rendering.
         The actual processing will primarily take place in the underlying
         implementation (typically optimized Assembly / C / C++ code), but
-        <a href="#AudioWorker">direct JavaScript processing and synthesis</a>
+        <a href="#AudioWorklet">direct JavaScript processing and synthesis</a>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -6142,7 +5294,7 @@ <h2 id="AudioWorklet-Examples">
           This is an <code>Event</code> object which is dispatched to
           <a><code>ScriptProcessorNode</code></a> nodes. It will be removed
           when the ScriptProcessorNode is removed, as the replacement
-          <a>AudioWorker</a> uses the <a>AudioProcessEvent</a>.
+          <a>AudioWorkletNode</a> uses a different approach.
```
rtoy: Not for this PR, but I think everything would read much better if we just removed such explanations.  We've already marked ScriptProcessorNode as deprecated; no need to keep saying AudioWorkletNodes do something else.

rtoy:
lgtm

Issue #954, opened 2016-09-02
[TAG] The AudioWorklet class seems redundant


domenic:
All it does is define an import() method, but that is already defined by its base class, Worklet.

Instead, like https://drafts.css-houdini.org/css-paint-api-1/#dom-window-paintworklet, it seems best to not have this class at all, and just use Worklet.

hoch:
Yes, it makes sense. Will fix that.

hoch:
Fixed by #963.

Issue #950, opened 2016-09-02
[TAG] Remove non-UA-implemented stuff from AudioWorkletProcessor


domenic:
I raised this point before in previous threads, but it doesn't seem to have been reflected. Chrome's intent to implement makes this particularly urgent, as it's important for implementers to be able to get an accurate idea of what they are expected to implement.

As far as I can tell, AudioWorkletProcessor should only have postMessage, onmessage, and contextInfo. process and parameterDescriptors are user-supplied and do not belong in the IDL.

joeberkovitz:
@hoch is there any reason why this is not a matter of simply removing the noted methods and attributes from the IDL?

hoch:
@joeberkovitz Agreed. I am removing them from the IDL.

@domenic What is the best practice to present an attribute or a method that cannot be captured by IDL? Perhaps an informative section? an example? I am planning to do something similar to this: https://www.w3.org/TR/css-paint-api-1/#paint-worklet.

domenic:
@hoch an example sounds perfect to me. Maybe something like:

Developers will typically create subclasses which extend AudioWorkletProcessor. These must implement a method named process() that (does X) and have a static property named parameterDescriptors which is an iterable of AudioParamDescriptors. An example of a minimal subclass is as follows:

```js
class MyProcessor extends AudioWorkletProcessor {
  static get parameterDescriptors() { return [{
      name: 'bitDepth',
      defaultValue: 12,
      minValue: 1
      maxValue: 16 
    }];
  }

  process() {
    // do X
  }
}
```

domenic:
More specifically, I actually kind of don't like the example from the Paint API which says "This is how the class should look" and then uses IDL. Web developers don't know how to read IDL.

hoch:
Okay, I'll go with the example then. Thanks!

hoch:
Fixed by #963.

PR #963, opened 2016-09-09
Rearranging AudioWorkletProcessor section

hoch:
PTAL. 

This PR is to address:
- #950: removed non-UA-implemented stuff and created a non-normative section to provide an example with short explanation.
- #954: removed redundant AudioWorklet interface. Now it is just Worklet.

Preview at: https://hoch.github.io/web-audio-api/#AudioWorkletProcessor


index.html
```diff
@@ -5466,70 +5442,66 @@ <h2 id="AudioWorkletProcessor">
...
-                  optional sequence&lt;Object&gt; transfer
-                </dt>
-                <dd></dd>
               </dl>
             </dd>
           </dl>
+          <p>
+            User can define a custom audio processor by extending
+            <a>AudioWorkletProcessor</a>. The subclass must define a method
+            named <strong>process()</strong> that implements the audio
```
rtoy: Why strong and not code? It is code.

index.html
```diff
@@ -5466,70 +5442,66 @@ <h2 id="AudioWorkletProcessor">
...
+            processing algorithm and have a static property named
+            <strong>parameterDescriptors</strong> which is an iterable of
```
rtoy: Use code isntead of strong for parameterDescriptors?

hoch:
Any opinion @joeberkovitz and @padenot?

index.html
```diff
@@ -5466,70 +5443,72 @@ <h2 id="AudioWorkletProcessor">
...
                 </dt>
                 <dd></dd>
-                <dt>
-                  optional sequence&lt;Object&gt; transfer
-                </dt>
-                <dd></dd>
               </dl>
             </dd>
           </dl>
+          <section class="informative">
```
joeberkovitz: Why is this informative and not normative?

hoch: Because this section is built around the example of process() method, which cannot be a part of IDL. The text is also not normative; it is more of an explanation on how to extend the AWP class.

hoch: Is there anything more to do here?

joeberkovitz: The fact that `process()` isn't part of the interface's IDL doesn't mean that we can go with an informative description of it alone. If we keep this piece informative, I feel we still need a normative description of `process()` including the exact data types of its parameters and the exact way it fits into the processing model.

hoch: So, we need to describe the API in a normative description without using WebIDL? By being "normative", do you want more of an algorithm?

hoch: @domenic Could you take a look? What can be done to improve the normativeness of this section?

hoch: @joeberkovitz I've changed the prose into a more normative description of the method. Should I remove 'informative' class as well?

joeberkovitz: Yes, I think the goal is to provide implementors/developers with a hard-edged definition of the features that a valid AudioWorkletProcessor must expose. I've been giving some thought to what might make this section more normative and here are my suggestions:

1. remove the `informative` class.

2. Retitle the section from "Subclassing AudioWorkletProcessor" to "Defining A Valid AudioWorkletProcessor". This makes it clear that the section is not merely about subclassing (which affords complete freedom as to what is added to the superclass) but defines a set of _mandatory elements_ that must be present for the AWP to function. Subclassing is merely the means to this more important end.

3. In this section, reference or link back to steps 5 and 6 in the description of `registerProcessor()`, which actually check the validity of a given AWP subclass. This emphasizes the connection between

hoch:
I am not sure how to respond the "requested change", but the line 5461 has been fixed (removed the informative label) with https://github.com/WebAudio/web-audio-api/pull/963/commits/81268260239f842699721359e0d17f43d77c314a.

hoch:
@joeberkovitz Thanks for your review!

PR #964, opened 2016-09-09
Fix #344: cancelScheduledValues and hold

### LINKED ISSUE #344: cancelScheduledValues should set the current value as a parameter change

rtoy:
Specify the algorithm (mentioned in #344) for how
cancelScheduledValues should behave to hold the value.


padenot:
Looks good, but I'm concerned about retro-compatibility. An alternative would be to introduce a new method, we can discuss at TPAC.

rtoy:
PTAL. I still need to add the explanatory diagrams, but the text is complete, I think.

padenot:
Yep, the text look good. We can merge this, or we can wait for the diagram, as you want.

rtoy:
Graphs added.  See https://rawgit.com/WebAudio/web-audio-api/86166ec39b9b291bab8f52ffdf0c116f78f8ab91/index.html#widl-AudioParam-cancelAndHoldAtTime-AudioParam-double-cancelTime

They're not great (I'm learning OmniGraffle at the same time), but perhaps passable for now until we get better diagrams.

padenot:
This has an unrelated commit.

rtoy:
Yeah, I used a different machine to create the graphs and checked them into the wrong branch. I'll remove those changes shortly.

PR #976, opened 2016-09-15
Adding normative description of instantiation process

hoch:
PTAL.

This PR adds a section that describes the instantiation process of AudioWorkletNode and AudioWorkletProcessor in a normative fashion.

CC: @bfgeek @domenic 

Preview branch: https://hoch.github.io/web-audio-api/#instantiation-of-AudioWorkletNode-and-AudioWorkletProcessor

### DIFF
--- a/index.html
+++ b/index.html
@@ -4657,14 +4657,14 @@ window.audioWorklet.import("bypass.js").then(function () {
               <ol>
                 <li>If the <code><em>name</em></code> exists as a key in the
                 <a>node name to processor definition map</a>, throw a
-                <code>NotSupportedError</code> exception and abort all these
-                steps because registering a definition with a duplicated key is
-                not allowed.
+                <code>NotSupportedError</code> exception and abort these steps
+                because registering a definition with a duplicated key is not
+                allowed.
                 </li>
                 <li>If the result of <code><a href=
                 "http://www.ecma-international.org/ecma-262/6.0/#sec-isconstructor">
                   IsConstructor</a>(argument=<em>processorConstructor</em>)</code>
-                  is false, throw a <code>TypeError</code> and abort all these
+                  is false, throw a <code>TypeError</code> and abort these
                   steps.
                 </li>
                 <li>Let <code><em>prototype</em></code> be the result of <code>
@@ -4683,14 +4683,14 @@ window.audioWorklet.import("bypass.js").then(function () {
                 "http://www.ecma-international.org/ecma-262/6.0/#sec-iscallable">
                   IsCallable</a>(argument=Get(O=<em>prototype</em>,
                   P="process"))</code> is false, throw a <code>TypeError</code>
-                  and abort all these steps.
+                  and abort these steps.
                 </li>
                 <li>If the result of <code><a href=
                 "http://www.ecma-international.org/ecma-262/6.0/#sec-get-o-p">
                   Get</a>(O=<em>processorConstructor</em>,
                   P="parameterDescriptors")</code> is not an array or
                   <code>undefined</code>, throw a <code>TypeError</code> and
-                  abort all these steps.
+                  abort these steps.
                 </li>
                 <li>Let <em>definition</em> be a new
                 <a>AudioWorkletProcessor</a> definition with:
@@ -5062,10 +5062,10 @@ class MyProcessor extends AudioWorkletProcessor {
             <li>Let <var>this</var> be the instance being created by
             constructor of <a>AudioWorkletNode</a> or its subclass.
             </li>
-            <li>If the <var>name</var> does not exists as a key in the
+            <li>If <var>name</var> does not exists as a key in the
             <a>AudioWorkletGlobalScope</a>’s <a>node name to processor
             definition map</a>, throw a <code>NotSupportedError</code>
-            exception and abort all these steps.
+            exception and abort these steps.
             </li>
             <li>Let <var>processorConstructor</var> be the result of looking up
             <var>name</var> on the <a>AudioWorkletGlobalScope</a>’s <a>node
@@ -5112,8 +5112,8 @@ class MyProcessor extends AudioWorkletProcessor {
                   </ul>
                 </li>
                 <li>If <var>options</var> contains a dictionary member with
-                name <var>paramName</var> whose type is <code>Number</code>,
-                perform <a href=
+                name <var>paramName</var> whose value is of type
+                <code>Number</code>, perform <a href=
                 "https://tc39.github.io/ecma262/#sec-set-o-p-v-throw">Set</a>(<var>audioParam</var>,
                 "value", <a href="https://tc39.github.io/ecma262/#sec-get-o-p">
                   Get</a>(O=<var>options</var>, P=<var>paramName</var>),
--- a/index.html
+++ b/index.html
@@ -4558,8 +4558,9 @@ if ((loopStart || loopEnd) &amp;& loopStart &gt;= 0 &amp;& loopEnd &gt; 0 &amp;&
               <dfn>node name to processor definition map</dfn><br>
               Belongs to <a>AudioWorkletGlobalScope</a>. This map stores
               definitions of <a>AudioWorkletProcessor</a> with the associated
-              string key. Initially it is empty and populated when
-              <a href="widl-AudioWorkletGlobalScope-registerProcessor-void-DOMString-name-VoidFunction-processorConstructor">registerProcessor</a> method is called.
+              string key. Initially it is empty and populated when <a href=
+              "widl-AudioWorkletGlobalScope-registerProcessor-void-DOMString-name-VoidFunction-processorConstructor">
+              registerProcessor</a> method is called.
             </li>
             <li>
               <dfn>node instance set</dfn><br>
@@ -4572,6 +4573,7 @@ if ((loopStart || loopEnd) &amp;& loopStart &gt;= 0 &amp;& loopEnd &gt; 0 &amp;&
           </ul>
           <pre class="example" title=
           "Defining and registering an AudioWorkletProcessor in AudioWorkletGlobalScope">
+
 // bypass.js script file
 registerProcessor("Bypass", class extends AudioWorkletProcessor {
   process (inputs, outputs) {
@@ -4583,6 +4585,7 @@ registerProcessor("Bypass", class extends AudioWorkletProcessor {
           </pre>
           <pre class="example" title=
           "Importing a script and creating AudioWorkletNode in the main global scope">
+
 window.audioWorklet.import("bypass.js").then(function () {
   var context = new AudioContext();
   var bypass = new AudioWorkletNode(context, "Bypass");
@@ -5123,8 +5126,8 @@ window.audioWorklet.import("bypass.js").then(function () {
             <li>Set <a>processor reference</a> in <var>this</var> to
             <var>processorInstance</var>.
             </li>
-            <li>Add <var>nodeInstance</var> to the <a>node
-            instance set</a> in the <a>BaseAudioContext</a>.
+            <li>Add <var>nodeInstance</var> to the <a>node instance set</a> in
+            the <a>BaseAudioContext</a>.
             </li>
             <li>Return <var>this</var> as a result of the construction.
             </li>
--- a/index.html
+++ b/index.html
@@ -4925,7 +4925,7 @@ registerProcessor("Bypass", class extends AudioWorkletProcessor {
               <a>AudioParam</a> in the node. The step 5 and 6 of <a href=
               "#widl-AudioWorkletGlobalScope-registerProcessor-void-DOMString-name-VoidFunction-processorCtor">
               registerProcessor()</a> ensure the validity of a given
-              AudioWorkletProcessor subclass.
+              <a>AudioWorkletProcessor</a> subclass.
             </p>
             <p>
               An example of subclass is as follows:
--- a/index.html
+++ b/index.html
@@ -4573,6 +4573,7 @@ if ((loopStart || loopEnd) &amp;& loopStart &gt;= 0 &amp;& loopEnd &gt; 0 &amp;&
           </ul>
           <pre class="example" title=
           "Defining and registering an AudioWorkletProcessor in AudioWorkletGlobalScope">
+
 // bypass.js script file
 registerProcessor("Bypass", class extends AudioWorkletProcessor {
   process (inputs, outputs) {
@@ -4584,6 +4585,7 @@ registerProcessor("Bypass", class extends AudioWorkletProcessor {
           </pre>
           <pre class="example" title=
           "Importing a script and creating AudioWorkletNode in the main global scope">
+
 // Main global scope
 window.audioWorklet.import("bypass.js").then(function () {
   var context = new AudioContext();
--- a/index.html
+++ b/index.html
@@ -4550,7 +4550,7 @@ if ((loopStart || loopEnd) &amp;& loopStart &gt;= 0 &amp;& loopEnd &gt; 0 &amp;&
             "https://drafts.css-houdini.org/worklets/#dom-worklet-import">import(moduleUrl)</a>
             method registers class definitions of <a>AudioWorkletProcessor</a>
             under the <a>AudioWorkletGlobalScope</a>. There are two internal
-            storages for the imported class definitions and the active
+            storage areas for the imported class definitions and the active
             instances created from the definition.
           </p>
           <ul>
@@ -4558,7 +4558,8 @@ if ((loopStart || loopEnd) &amp;& loopStart &gt;= 0 &amp;& loopEnd &gt; 0 &amp;&
               <dfn>node name to processor definition map</dfn><br>
               Belongs to <a>AudioWorkletGlobalScope</a>. This map stores
               definitions of <a>AudioWorkletProcessor</a> with the associated
-              string key. Initially it is empty and populated when <a href=
+              string key. Initially it is empty. It becomes populated when
+              <a href=
               "widl-AudioWorkletGlobalScope-registerProcessor-void-DOMString-name-VoidFunction-processorConstructor">
               registerProcessor</a> method is called.
             </li>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -5061,6 +5061,143 @@ <h2 id="AudioContextInfo">
...
+            <code><em>parameterDescriptors</em></code> perform the following
+            substeps:
+              <ol>
+                <li>Let <code><em>audioParam</em></code> be a new
+                <a>AudioParam</a> instance with:
+                  <ul>
+                    <li>
+                      <code><em>paramName</em></code> being the result
+                      <code><a href=
+                      "https://tc39.github.io/ecma262/#sec-get-o-p">Get</a>(O=item,
```
joeberkovitz: For `item` here and in the following few lines, I think you mean `descriptor` which is the iteration variable above.

index.html
```diff
@@ -5061,6 +5061,143 @@ <h2 id="AudioContextInfo">
...
+            </li>
+            <li>Let <code><em>processorConstructor</em></code> be the result of
+            looking up name on the <a>AudioWorkletGlobalScope</a>’s <a>node
+            name to processor definition map</a>.
+            </li>
+            <li>Let <code><em>baseNodeInstance</em></code> be a new instance of
+            <a>AudioWorkletNode</a>.
+            </li>
+            <li>Let <code><em>options</em></code> be the optional
+            <code>options</code> argument in the constructor.
```
joeberkovitz: For the `Get` operations below, does `options` need to be defaulted to an empty object?

index.html
```diff
@@ -5061,6 +5061,143 @@ <h2 id="AudioContextInfo">
...
+            <code>DOMString</code> or does not exists as a key in the
+            <a>AudioWorkletGlobalScope</a>’s <a>node name to processor
+            definition map</a>, throw a <code>NotSupportedError</code>
+            exception and abort all these steps.
+            </li>
+            <li>Let <code><em>processorConstructor</em></code> be the result of
+            looking up name on the <a>AudioWorkletGlobalScope</a>’s <a>node
+            name to processor definition map</a>.
+            </li>
+            <li>Let <code><em>baseNodeInstance</em></code> be a new instance of
```
joeberkovitz: Does this and other references to `AudioWorkletNode` as a class need to be generalized in light of the possibility that the node being instantiated is actually a _subclass_ of `AudioWorkletNode`?

index.html
```diff
@@ -5061,6 +5061,143 @@ <h2 id="AudioContextInfo">
...
+            <a>AudioWorkletNode</a>.
+            </li>
+            <li>Let <code><em>options</em></code> be the optional
+            <code>options</code> argument in the constructor.
+            </li>
+            <li>Let <code><em>processorInstance</em></code> be a new instance
+            of <a>AudioWorkletProcessor</a> created by processorConstructor.
```
joeberkovitz: A structured clone of the `options` object needs to be passed to the `AudioWorkletProcessor` constructor (this is how we agreed one communicates node options to the rendering side).

joeberkovitz: Should there be a formal call to `Construct(...)` here to represent the construction of the AWP?

index.html
hoch: I am still thinking about how to reconcile both cases into the one step: (a) direct instantiation from AWN constructor and (b) instantiation from AWN subclass. If you have a better prose/description for this, please feel free to jump in.

index.html
```diff
@@ -5061,6 +5061,143 @@ <h2 id="AudioContextInfo">
             </dl>
           </section>
         </section>
+        <section>
+          <h2 id="instantiation-of-AudioWorkletNode-and-AudioWorkletProcessor">
```
domenic: Will you remove the existing description of the AudioWorkletNode constructor? It seems redundant now.

index.html
```diff
@@ -5061,6 +5061,143 @@ <h2 id="AudioContextInfo">
...
+            maintain the internal reference to each other until the
+            <a>AudioWorkletNode</a> instance is destroyed.
+          </p>
+          <p>
+            When <code><a>AudioWorkletNode</a>(<em>context</em>, <em>name</em>,
+            <em>options</em>)</code> constructor is called, the user agent must
+            run the following steps in a synchronous fashion:
+          </p>
+          <ol>
+            <li>If the <code><em>context</em></code> is not an instance of <a>
```
domenic: Nit: through this PR it seems you are misusing `<em>` when it should be `<var>`.

domenic: "an instance of" should probably be "a platform object that implements"

index.html
```diff
@@ -5061,6 +5061,143 @@ <h2 id="AudioContextInfo">
...
+              BaseAudioContext</a>, throw a <code>TypeError</code> and abort
+              all these steps.
+            </li>
+            <li>If the <code><em>name</em></code> is not a valid
+            <code>DOMString</code> or does not exists as a key in the
```
domenic: It's impossible for name to not be a valid DOMString; the Web IDL layer takes care of that before this algorithm is ever called.

index.html
domenic: Yeah, the existing setup is very strange if the goal is to allow subclassing. I would instead not create any objects at all. Just operate on `this` (which will be an instance of either AWN or a subclass). That is a bit unusual for Web IDL-defined constructors but it is the clearest thing to do.

index.html
domenic: Yes, a Construct call would be good. And yes, create a structured clone (using https://html.spec.whatwg.org/multipage/infrastructure.html#structuredclone, with target realm being the relevant realm of processorConstructor).

index.html
```diff
@@ -5061,6 +5061,143 @@ <h2 id="AudioContextInfo">
...
+                      <code><a href=
+                      "https://tc39.github.io/ecma262/#sec-get-o-p">Get</a>(O=item,
+                      P="maxValue")</code>.
+                    </li>
+                  </ul>
+                </li>
+                <li>Let <code><em>initialValue</em></code> be the result of
+                <code><a href=
+                "https://tc39.github.io/ecma262/#sec-get-o-p">Get</a>(O=<em>options</em>,
+                P=<em>paramName</em>)</code>.
```
domenic: Since options is an argument to your constructor, it will already be processed by Web IDL, and converted to a number (if it is one of the unsigned long members of the dictionary). So these two steps should instead be something like "If options contains a dictionary member with name _paramName_ whose type is unsigned long, perform Set(audioParam, "value", the value of that dictionary member)."

index.html
```diff
@@ -5061,6 +5061,143 @@ <h2 id="AudioContextInfo">
...
+                      <code>Number</code>, perform <code><a href=
+                      "https://tc39.github.io/ecma262/#sec-set-o-p-v-throw">Set</a>(O=<em>audioParam</em>,
+                      P="value", V=<em>initialValue</em>)</code>. Otherwise
+                      ignore and go to the next step.
+                    </li>
+                  </ul>
+                </li>
+                <li>Perform <code><a href=
+                "https://tc39.github.io/ecma262/#sec-set-o-p-v-throw">Set</a>(O=<em>baseNodeInstance</em>,
+                P=<em>paramName</em>, V=<em>audioParam</em>, false)</code>.
```
domenic: Do you really want to perform a Set() here? That means you will trigger any overridden setters users have monkeypatched onto the object. Probably instead you want to set the appropriate internal sots or associated values. (Right now I don't see AudioWorkletNode having any internal state to represent these concepts, but maybe I just didn't find it...)

index.html
```diff
@@ -5061,6 +5061,143 @@ <h2 id="AudioContextInfo">
...
+                abort all these steps.
+                </li>
+              </ul>
+            </li>
+            <li>Create the internal association between
+            <code><em>nodeInstance</em></code> and
+            <code><em>processorInstance</em></code>. Note that these are not
+            exposed to users:
+              <ol>
+                <li>Set <code><em>processorReference</em></code> in <code><em>
```
domenic: processorReference and nodeReference are not discussed anywhere else. Probably they are part of internal state of the objects. In which case, in the definition of AudioWorkletNode, you should say something like "Every AudioWorkletNode has an associated **processor reference**, initially null." and then cross-reference it here. (You should not use code font for it, since it is not user-exposed.)

index.html
domenic: This is already taken care of by Web IDL, since it is declared optional, so it will be converted to an empty dictionary (not object).

index.html
hoch: Yes, through a different PR.

index.html
hoch: Oh, I simply didn't know about `<var>`. Thanks!

index.html
hoch: Oops. You're correct.

index.html
hoch: I wanted to create an AudioParam instance and make it a property belongs to `this`, which is AudioWorkletNode.

domenic: Hmm. The more normal way to do this would be to have attributes defined in Web IDL on AudioWorkletNode that return "this AudioWorkletNode's associated xParam". Or, is this an open-ended set of possible params? Then I guess the way you're doing it makes sense.

hoch: Yes, this allows to create the AudioParam based on `parameterDescriptors()`, which is defined by the user. If using `Set()` is not really practical, how do I describe this in a normative style?

domenic: Ah OK, now that I understand it, it makes more sense. It's an unusual design, but this is a good way to do it. It would be slightly better if you used [CreateDataProperty](https://tc39.github.io/ecma262/#sec-createdataproperty); that will avoid triggering any setters (e.g. if they create a param named `onmessage` or something similar).

hoch: Thanks! `CreateDataProperty` looks like what I want.

index.html
hoch: Good point. I'll fix them.

index.html
hoch: Done.

hoch:
Sorry about the messy commit log; Tidy is adding a blank line after the new code example, and it breaks Travis CI.

index.html
hoch: Done.

index.html
hoch: Done.

index.html
hoch: Done.

index.html
hoch: Done.

index.html
hoch: Done.

index.html
hoch: Done.

hoch: Done.

index.html
hoch: Done.

index.html
hoch: As Domenic pointed out, I left this one out.

index.html
hoch: I've added a sentence to address this. PTAL at https://github.com/WebAudio/web-audio-api/pull/976/commits/d6e40532409f0cf5fdf43fdba58aa5a71e5d6788.

index.html
```diff
@@ -4523,20 +4523,19 @@ <h2 id="AudioWorklet-concepts">
...
             An associated pair of objects must be defined in order to realize
-            this mechanism: <a><code>AudioWorkletNode</code></a> and
-            <a><code>AudioWorkletProcessor</code></a>. The former represents
-            the interface for the main global scope similar to other AudioNode
-            objects and the latter implements the internal audio processing
-            within a special scope named
-            <a><code>AudioWorkletGlobalScope</code></a>.
+            this mechanism: <a>AudioWorkletNode</a> and
+            <a>AudioWorkletProcessor</a>. The former represents the interface
+            for the main global scope similar to other <a>AudioNode</a> objects
```
rtoy: Typo: "objects" -> "objects,"

index.html
```diff
@@ -4546,44 +4545,58 @@ <h2 id="AudioWorklet-concepts">
...
           <p>
-            Importing a JavaScript file via <code>import(url)</code> method
-            registers class definitions of
-            <a><code>AudioWorkletProcessor</code></a> under the
-            <a><code>AudioWorkletGlobalScope</code></a>. Imported definitions
-            are stored in the <a><em>node name to processor definition
-            map</em></a> and any <a><code>AudioContext</code></a> instance from
-            the associated main global scope can access this map to create a
-            custom node.
+            Importing a JavaScript file via <a href=
```
rtoy: "via" -> "via the"

index.html
```diff
@@ -4546,44 +4545,58 @@ <h2 id="AudioWorklet-concepts">
...
+            under the <a>AudioWorkletGlobalScope</a>. There are two internal
+            storage areas for the imported class definitions and the active
+            instances created from the definition.
           </p>
-          <blockquote>
-            <pre>
-// 'bypass.js' in AudioWorkletGlobalScope
+          <ul>
+            <li>
+              <dfn>node name to processor definition map</dfn><br>
```
rtoy: I think using `<br>` here isn't great. Perhaps a `<dl>` list is better.

index.html
```diff
@@ -4546,44 +4545,58 @@ <h2 id="AudioWorklet-concepts">
...
+              Belongs to <a>AudioWorkletGlobalScope</a>. This map stores
+              definitions of <a>AudioWorkletProcessor</a> with the associated
+              string key. Initially it is empty. It becomes populated when
```
rtoy: It seems you haven't stated what the "string key" is or where it comes from.  A link to where this is defined is needed. Or some other text defining this before this is used.

rtoy: "Initially it is empty". The "it" here is ambiguous. Are you referring to the "key" or the map"?

index.html
```diff
@@ -4546,44 +4545,58 @@ <h2 id="AudioWorklet-concepts">
...
+            <li>
+              <dfn>node instance set</dfn><br>
+              Belongs to an instance of <a>BaseAudioContext</a>. This set
+              stores the reference to all the active instances of
+              <a>AudioWorkletNode</a> that are created by the associated
+              <a>BaseAudioContext</a>. The context keeps the track of instances
+              until they are destroyed by UA.
+            </li>
+          </ul>
+          <pre class="example" title=
```
rtoy: I think these examples are in the wrong place.  For the new reader, it's hard to make any sense of this example when you haven't defined exactly what registerProcessor looks like and what's supposed to be inside it.

index.html
```diff
@@ -4720,12 +4733,12 @@ <h2 id="AudioWorkletNode">
...
           <p>
-            This interface represents a user-defined
-            <a><code>AudioNode</code></a> which lives on the <a>control
-            thread</a>. The user can create an <code>AudioWorkletNode</code>
-            from an <a><code>AudioContext</code></a> and such a node can be
-            connected with other built-in <a><code>AudioNode</code></a>s to
-            form an audio graph.
+            This interface represents a user-defined <a>AudioNode</a> which
+            lives on the <a>control thread</a>. The user can create an
+            <a>AudioWorkletNode</a> from an <a>BaseAudioContext</a> and such a
```
rtoy: "BaseAudioContext and" -> "BaseAudioContext, and"

index.html
hoch: This is to provide the conceptual overview of AudioWorklet system without going into the detail. FWIW, this change was made based on @domenic's feedback.

index.html
hoch: Done.

index.html
hoch: Done.

index.html
hoch: Done.

index.html
hoch: Done. Also rephrased the `This map...` part.

index.html
```diff
@@ -5061,6 +5034,108 @@ <h2 id="AudioContextInfo">
...
+          <p>
+            When <a>AudioWorkletNode</a>(<var>context</var>, <var>name</var>,
+            <var>options</var>) constructor is called, the user agent must run
+            the following steps <a data-lt="atomic">atomically</a>.
+          </p>
+          <ol>
+            <li>Let <var>this</var> be the instance being created by
+            constructor of <a>AudioWorkletNode</a> or its subclass.
+            </li>
+            <li>If the <var>context</var> is not a platform object which
```
hoch: Done.

index.html
```diff
@@ -5061,6 +5034,108 @@ <h2 id="AudioContextInfo">
...
+            implements <a>BaseAudioContext</a>, throw a <code>TypeError</code>
+            and abort all these steps.
+            </li>
+            <li>If the <var>name</var> does not exists as a key in the
+            <a>AudioWorkletGlobalScope</a>’s <a>node name to processor
+            definition map</a>, throw a <code>NotSupportedError</code>
+            exception and abort all these steps.
+            </li>
+            <li>Let <var>processorConstructor</var> be the result of looking up
+            name on the <a>AudioWorkletGlobalScope</a>’s <a>node name to
```
hoch: Done.

index.html
hoch: Done.

joeberkovitz:
@hoch I'm good.

index.html
rtoy: I still think these are too early, and would fit in the on AudioWorklet Examples.

But I'm not going to hold up this PR for that.  We can move these things around later, if necessary

index.html
```diff
@@ -5061,6 +5040,104 @@ <h2 id="AudioContextInfo">
...
+          <p>
+            When <a>AudioWorkletNode</a>(<var>context</var>, <var>name</var>,
+            <var>options</var>) constructor is called, the user agent must run
+            the following steps <a data-lt="atomic">atomically</a>.
+          </p>
+          <ol>
+            <li>Let <var>this</var> be the instance being created by
+            constructor of <a>AudioWorkletNode</a> or its subclass.
+            </li>
+            <li>If the <var>name</var> does not exists as a key in the
```
padenot: > If <code>name</code>

we're using the variable here.

index.html
```diff
@@ -5061,6 +5040,104 @@ <h2 id="AudioContextInfo">
...
+            <a>AudioWorkletGlobalScope</a>’s <a>node name to processor
+            definition map</a>, throw a <code>NotSupportedError</code>
+            exception and abort all these steps.
```
padenot: nit: this is usually written as "abort these steps" in other specs.

index.html
```diff
@@ -5061,6 +5040,104 @@ <h2 id="AudioContextInfo">
...
+                    <li>
+                      <var>maxValue</var> being the result <a href=
+                      "https://tc39.github.io/ecma262/#sec-get-o-p">Get</a>(O=<var>descriptor</var>,
+                      P="maxValue").
+                    </li>
+                  </ul>
+                </li>
+                <li>If <var>options</var> contains a dictionary member with
+                name <var>paramName</var> whose type is <code>Number</code>,
+                perform <a href=
```
padenot: nit: "whose value is of type <code>Number</code>" I'd say.

index.html
hoch: Done.

index.html
hoch: Done. Fixed other instances as well..

index.html
hoch: Done.

PR #978, opened 2016-09-16
Fix #924: Changing fftSize of Analyser

### LINKED ISSUE #924: Changing fftSize of AnalyserNode
Currently, the `fftSize` attribute of an `AnalyserNode` can be changed at any time.  This is fine.  However, there are memory implications with this flexibility.  Suppose it is set to 128 now.  Later on I want to set it to 32768 and then immediately ask for FFT data.  For this to work as specified, the node must keep all 32768 frames in an internal buffer so that if the user asks for a 32K-point FFT, we can compute the FFT now using the most recent 32K frames.

This is unnecessary extra memory if we never want to use 32K FFTs, and is wasted memory, especially on low-end devices with limited memory.

Is this the intent?  Would it be useful to have a maxFFTSize attribute that sets the max FFT that will be requested and hence limiting the amount of memory.  But this would have to an argument to `createAnalyser()` or the constructor for it to be useful.

rtoy:
If the fftSize is changed, reset the internal state for the FFT
computations as if a new AnalyserNode were created.

### DIFF
--- a/index.html
+++ b/index.html
@@ -6558,6 +6558,15 @@ function calculateNormalizationScale(buffer)
               thrown</span>. The default value is 2048. Note that large FFT
               sizes can be costly to compute.
             </p>
+            <p>
+              If the <code>fftSize</code> is changed, then all state associated
+              with smoothing of the frequency data (for <a href=
+              "#widl-AnalyserNode-getByteFrequencyData-void-Uint8Array-array"><code>
+              getByteFrequencyData</code></a> and <a href=
+              "#widl-AnalyserNode-getFloatFrequencyData-void-Float32Array-array">
+              <code>getFloatFrequencyData</code></a>) is reset, as if a new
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -6558,6 +6558,16 @@ <h3 id="Convolution-channel-configurations">
               thrown</span>. The default value is 2048. Note that large FFT
               sizes can be costly to compute.
             </p>
+            <p>
+              If the <code>fftSize</code> is changed to a different value, then
+              all state associated with smoothing of the frequency data (for
```
padenot: Can we spell out the state that is reset here? We can link to https://webaudio.github.io/web-audio-api/#dfn-previous-block I think.

padenot:
LGTM with the comment addressed.

rtoy:
Thanks, Paul.

PR #979, opened 2016-09-19
Consistently use IsDetachedBuffer and link to the spec when referencing a detach operation.

### LINKED ISSUE #687: decodeAudioData: "Neuter the audioData ArrayBuffer in such a way that JavaScript code may not access or modify the data anymore." uses weasel words
Link to <https://heycam.github.io/webidl/#dfn-detach>.

padenot:
There is no facility in WebIDL for now to check if a buffer has been detached, so we're using ES's concepts.
 
This fixes #687.


rtoy:
lgtm.

Issue #264, opened 2013-10-18
Need a AudioMediaStreamTrack-based StreamSourceNode factory method


olivierthereaux:
Comment from Stefan Håkansson LK:
http://lists.w3.org/Archives/Public/public-audio/2013OctDec/0078.html

In section 4.24 (_Note: probably referring to 4.25. The MediaStreamAudioSourceNode Interface_) it is said that "The first AudioMediaStreamTrack from 
the MediaStream will be used as a source of audio."

Which audio track that is the first one is undefined (in the case when 
there is more than one). I think it would make much more sense to deal 
directly with a MediaStreamTrack (of kind audio) in this API (which 
should as a consequence be renamed to "MediaStreamTrackAudioSourceNode").

And it would probably make sense to rename "MediaStreamAudioDestinationNode" to "MediaStreamTrackAudioDestinationNode", and have it produce a MediaStreamTrack.

joeberkovitz:
This is ready for editing since it can be taken together with #132 whose resolution is well defined.

padenot:
This issue is about speccing a way of choosing the right track in the stream.

PR #982, opened 2016-09-19
 Spec which track to use when using MediaStreamAudioSourceNode, and introduce MediaStreamTrackAudioSourceNode

padenot:
We needed a arbitrary but stable order, so I did what I could.

Also, multi-track MediaStream are not that common these days, so changing this is probably safe. People who rely on the order for multi-track stream are going to be using the new Node.

### DIFF
--- a/index.html
+++ b/index.html
@@ -9605,10 +9605,17 @@ odd function with period \(2\pi\).
         </h2>
         <p>
           This interface represents an audio source from a
-          <code>MediaStream</code>. The first
-          <code>AudioMediaStreamTrack</code> from the <code>MediaStream</code>
-          will be used as a source of audio. Those interfaces are described in
-          [[!mediacapture-streams]].
+          <code>MediaStream</code>. The track that will be used as the source of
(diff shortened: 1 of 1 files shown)


padenot:
Now tidied.

rtoy:
Intended to fix #264.

padenot:
@rtoy, have you had the change to have a look at this one ?

padenot:
@joeberkovitz, @hoch, @mdjp, @svgeesus, any opinion ?

index.html
```diff
@@ -1912,6 +1918,22 @@ <h2 id="AudioContext">
             </dl>
           </dd>
           <dt>
+            MediaStreamTrackAudioSourceNode createMediaStreamSource()
```
joeberkovitz: Doesn't this need a distinct factory method name?

padenot: Right, updated.

joeberkovitz:
LGTM modulo the conflicts -- should be good to merge

PR #991, opened 2016-09-24
Add a concept of 'allowed to start' and use it when creating an AudioContext and resuming it (closes #836).

### LINKED ISSUE #836: Expose whether AudioBufferSourceNode.start() succeeded
/CC @rtoy @padenot @jernoble 

As far as I can tell, AudioBufferSourceNode.start() is a no-op on Safari iOS unless the call came from a user gesture. Chrome is considering to do the same in order to have more consistency with regards to autoplay behaviour on mobile. However, the start() method doesn't expose if the call was successful which makes me uncomfortable doing this change.

Would it make sense to have a pattern similar to HTMLMediaElement.play() for AudioBufferSourceNode and return a Promise that rejects if the playback isn't allowed and fulfil when the playback starts?

mounirlamouri:
@rtoy and @padenot PTAL

### DIFF
--- a/index.html
+++ b/index.html
@@ -1014,6 +1014,10 @@ function setupRoutingGraph() {
               of the <a>AudioContext</a> is already <code>running</code>,
               resolve <em>promise</em>, return it, and abort these steps.
               </li>
+              <li>If the <a>AudioContext</a> is not <a>allowed to start</a>,
+              append <em>promise</em> to <a>pendingResumePromises</a> and abort
+              these steps, returning <em>promise</em>.
+              </li>
               <li>Set the <em>control thread state</em> flag on the
               <a>AudioContext</a> to <code>running</code>.
               </li>
@@ -1034,6 +1038,9 @@ function setupRoutingGraph() {
               <li>In case of failure, queue a task on the <a>control thread</a>
               to execute the following, and abort these steps
                 <ol>
+                  <li>Reject all promises from <a>pendingResumePromises</a> in
+                  order, then clear <a>pendingResumePromises</a>.
+                  </li>
                   <li>Reject <em>promise</em>
                   </li>
                 </ol>
@@ -1041,6 +1048,9 @@ function setupRoutingGraph() {
               <li>Queue a task on the <a>control thread</a>'s event loop, to
               execute these steps:
                 <ol>
+                  <li>Resolve all promises from <a>pendingResumePromises</a> in
+                  order, then clear <a>pendingResumePromises</a>.
+                  </li>
                   <li>Resolve <em>promise</em>.
                   </li>
                   <li>If the <a href="#widl-audiocontext-state">state</a>
@@ -1661,6 +1671,21 @@ function setupRoutingGraph() {
           use cases, only a single <a><code>AudioContext</code></a> is used per
           document.
         </p>
+        <p>
+          An <a><code>AudioContext</code></a> is said to be <dfn>allowed to
+          start</dfn> if the user agent and the system allow audio output in
+          the current context. In other words, if the
+          <a><code>AudioContext</code></a> <em>control thread state</em> is
+          allowed to transition from <code>suspended</code> to
+          <code>running</code>.
+        </p>
+        <p class='note'>
+          For example, a user agent could require that an
+          <a><code>AudioContext</code></a> control thread state change to
+          running is <a href=
+          "https://html.spec.whatwg.org/multipage/interaction.html#triggered-by-user-activation">
+          triggered by a user activation</a> (as described in [[HTML]]).
+        </p>
         <dl title=
         '[Constructor(optional AudioContextOptions contextOptions)] interface AudioContext : BaseAudioContext'
         class='idl'>
@@ -1676,6 +1701,12 @@ function setupRoutingGraph() {
               <li>Set a <code>control thread state</code> to
               <code>suspended</code> on the <a>AudioContext</a>.
               </li>
+              <li>Let <dfn>pendingResumePromises</dfn> be an empty ordered list
+              of promises.
+              </li>
+              <li>If the <a>AudioContext</a> is not <a>allowed to start</a>,
+              abort these steps.
+              </li>
               <li>Send a <a>control message</a> to start processing.
               </li>
             </ol>
@@ -1684,7 +1715,7 @@ function setupRoutingGraph() {
               executing the following steps:
             </p>
             <ol>
-              <li>Attempt to <a href="#acquiring">Acquire system resources</a>.
+              <li>Attempt to <a href="#acquiring">acquire system resources</a>.
               </li>
               <li>In case of failure, abort these steps.
               </li>


index.html
```diff
@@ -999,6 +999,10 @@ <h2 id="BaseAudioContext">
               suspended.
             </p>
             <p>
+              Let <em>pendingResumePromises</em> be a list of promises that
+              were lef pending after a call.
```
padenot: s/lef/left/

index.html
```diff
@@ -999,6 +999,10 @@ <h2 id="BaseAudioContext">
...
+            </p>
+            <p>
               <span class="synchronous">When resume is called, execute these
```
padenot: Markup around `resume`.

index.html
```diff
@@ -999,6 +999,10 @@ <h2 id="BaseAudioContext">
               suspended.
             </p>
             <p>
+              Let <em>pendingResumePromises</em> be a list of promises that
```
padenot: This needs to be initialized to an empty, ordered list in the `AudioContext` ctor so that the prose is a bit more stronger.

index.html
mounirlamouri: N/A

mounirlamouri:
Comments applied. PTAL.

index.html
```diff
@@ -1661,6 +1671,17 @@ <h2 id="AudioContext">
           use cases, only a single <a><code>AudioContext</code></a> is used per
           document.
         </p>
+        <p>
+          An <a><code>AudioContext</code></a> is said to be <dfn>allowed to
+          start</dfn> if the user agent and the system allow sound output in
```
rtoy: "sound" -> "audio" ?

I think this sentence needs to be expanded a bit. It's kind of described later what happens, but a little more information here would be a useful introduction.

index.html
```diff
@@ -1661,6 +1671,17 @@ <h2 id="AudioContext">
...
+          the current context.
+        </p>
+        <p class='note'>
+          For example, a user agent could require that an
+          <a><code>AudioContext</code></a> start is <a href=
```
rtoy: What does "start" mean here?  The meaning is really overloaded here; clarify what start really means.

index.html
```diff
@@ -1676,15 +1697,19 @@ <h2 id="AudioContext">
               <li>Set a <code>control thread state</code> to
               <code>suspended</code> on the <a>AudioContext</a>.
               </li>
-              <li>Send a <a>control message</a> to start processing.
+              <li>Let <dfn>pendingResumePromises</dfn> be an empty ordered list
+              of promises.
+              </li>
+              <li>If the <a>AudioContext</a> is <a>allowed to start</a>, send a
```
rtoy: I think you also need to describe here what happens if the context is not allowed to start. Presumably all steps are aborted here.

index.html
```diff
@@ -1014,6 +1014,10 @@ <h2 id="BaseAudioContext">
               of the <a>AudioContext</a> is already <code>running</code>,
               resolve <em>promise</em>, return it, and abort these steps.
               </li>
+              <li>If the <a>AudioContext</a> is not <a>allowed to start</a>,
```
rtoy: Need to describe what happens if the context *is* allowed to start.

index.html
```diff
@@ -1034,13 +1038,19 @@ <h2 id="BaseAudioContext">
               <li>In case of failure, queue a task on the <a>control thread</a>
               to execute the following, and abort these steps
                 <ol>
+                  <li>Reject all promises from <a>pendingResumePromises</a> in
+                  order, then clear <a>pendingResumePromises</a>.
+                  </li>
                   <li>Reject <em>promise</em>
```
rtoy: This is a little confusing?  You've already added the promise to `pendingResumePromises` above because the context is not allowed to start.  So you reject the promise when handling `pendingResumePromises`, and then reject the promise again in this step.

index.html
```diff
@@ -1034,13 +1038,19 @@ <h2 id="BaseAudioContext">
...
                   </li>
                 </ol>
               </li>
               <li>Queue a task on the <a>control thread</a>'s event loop, to
               execute these steps:
                 <ol>
+                  <li>Resolve all promises from <a>pendingResumePromises</a> in
+                  order, then clear <a>pendingResumePromises</a>.
+                  </li>
                   <li>Resolve <em>promise</em>.
```
rtoy: Same issues as for rejecting `pendingResumePromises`

index.html
```diff
@@ -1676,15 +1697,19 @@ <h2 id="AudioContext">
               <li>Set a <code>control thread state</code> to
               <code>suspended</code> on the <a>AudioContext</a>.
               </li>
-              <li>Send a <a>control message</a> to start processing.
+              <li>Let <dfn>pendingResumePromises</dfn> be an empty ordered list
```
rtoy: Why is this list needed? We don't need such a list for any of the other promises returned from suspend/resume/close.

padenot: Yes, because we don't store them for later use, we immediately return them to content.

Here, we need to keep them around to resolve or reject them at a later date. It's necessary to store them somewhere to do that.

index.html
mounirlamouri: I'm following the pattern used for the steps above: if the condition is true, stops the steps and return. That implies that otherwise, we should continue.

index.html
mounirlamouri: We don't 'queue a control message` when the context is `not allowed to start` so it's not possible that `promise` was in the `pendingResumePromises`.

index.html
mounirlamouri: ditto

index.html
mounirlamouri: I've reversed the condition, it's now aborting the steps if not allowed to start instead.

index.html
mounirlamouri: I'm not sure how much I can say. This is a normative text that describes what `allowed to start` means. The note tries to give some more colour but it sounds fair for the reader to search in the spec how this is being used.

mounirlamouri: I've added an "In other words, [...]" sentence. HTH

index.html
mounirlamouri: I tried to clarify a bit. Though, it's a "note" so being hand-wavy is fine, it's not normative.

mounirlamouri:
Comments applied. PTAL :smile:

index.html
rtoy: Perhaps I'm confused, but lines 1017-1018 says that if the context is not allowed, the promise is appended to pendingResumePromises. And if the control message is not queued, how is pendingResumePromises ever handled?

mounirlamouri: If you create an AudioContext and call `resume()` when audio is not allowed, the promise will be left pending until you call `resume()` while audio is allowed. Then, the two promises will be resolved.

rtoy: My thinking on this was backwards.  This looks good.

padenot:
Merged. Thank you for this, @mounirlamouri !

mounirlamouri:
:dancer:

PR #994, opened 2016-09-27
Explain how suspend and resume work in the processing model.

### LINKED ISSUE #919: Interaction of suspend/resume and processing model is unclear
The Processing Model does not refer to the AudioContext's `state`, and the definition in https://webaudio.github.io/web-audio-api/#widl-BaseAudioContext-state does not itself reference any effect on the Processing Model.

The suspend/resume documentation alone https://webaudio.github.io/web-audio-api/#widl-BaseAudioContext-suspend-Promise-void talks about the effects of `state` changes. It feels as though we are missing a link that describes the linkage between control thread `state` change and the behavior of the rendering thread.

padenot:
This is for the real time part, another commit will explain how it works with
offline contexts.

This fixes #919.

Preview here: https://rawgit.com/WebAudio/web-audio-api/98c135ac98435de3dfba1b9699f61cad52d72f14/index.html

### DIFF
--- a/index.html
+++ b/index.html
@@ -954,6 +954,9 @@ function setupRoutingGraph() {
             <ol>
               <li>Attempt to <a>release system resources</a>.
               </li>
+              <li>Set the <a>rendering thread state</a> on the
+              <a>AudioContext</a> to <code>suspended</code>.
+              </li>
               <li>Queue a task on the <a>control thread</a>'s event loop, to
               execute these steps:
                 <ol>
@@ -1038,6 +1041,12 @@ function setupRoutingGraph() {
             </p>
             <ol>
               <li>Attempt to <a href="#acquiring">acquire system resources</a>.
+              </li>
+              <li>Set the <a>rendering thread state</a> flag on the
+              <a>AudioContext</a> to <code>running</code>.
+              </li>
+              <li>Start <a href="#rendering-loop">rendering the audio
+              graph</a>.
               </li>
               <li>In case of failure, queue a task on the <a>control thread</a>
               to execute the following, and abort these steps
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -9023,6 +9041,30 @@ <h2 id="MediaStreamAudioSourceNode">
...
+            buffer is often larger than a rendering quantum. In this case,
+            multiple invocations of the rendering algorithm will be called in a
+            rapid succession, in the same callback, before returning. After
+            some time, the underlying audio system will call the callback
+            again, and the algorithm will be executed again. This is an
+            implementation detail that should not be observable, apart from the
+            latency implications.
+          </p>
+          <p>
+            <a>OfflineAudioContext</a> will execute the algorithm continuously.
```
rtoy: Maybe add "continuously, until the AudioBuffer of the OfflineContext is filled". I read the current sentence as if the context runs forever.  And it's not really clear what you mean by continuously. I don't think we need to constrain the offline context in any way.

index.html
```diff
@@ -9023,6 +9041,30 @@ <h2 id="MediaStreamAudioSourceNode">
           "#control-message-queue">control message queue</a> <em>Q</em> is as
           follows.
         </p>
+        <p>
+          If the algorithm terminates normally (i.e. reaches its last step),
+          then it MUST be executed again in the future, to render the next
+          block of audio. Else, the <a>control thread</a> can make it run again
+          as needed.
```
rtoy: I don't understand what this is trying to say.

index.html
padenot: This is in relation with the paragraph above this one, where it is explained that the audio callback mechanism forces the rendering work to happen in chunks.

PR #996, opened 2016-09-27
Remove the section about JavaScript processing.

### LINKED ISSUE #45: Record all documentation that is considered developer documentation
> Originally reported on W3C Bugzilla [ISSUE-21548](https://www.w3.org/Bugs/Public/show_bug.cgi?id=21548) Tue, 02 Apr 2013 16:20:21 GMT
> Assigned to 

This is a placeholder to keep track of all the sections in the spec which are considered "developer documentation", to be split out from the spec and into a primer/developer doc type document.

padenot:
This fixes #45.

### DIFF
--- a/index.html
+++ b/index.html
@@ -10353,46 +10353,6 @@ function coneGain() {
           more work than is possible in real-time given the CPU's speed.
         </p>
       </section>
-      <section>
-        <h3 id="JavaScriptPerformance">
-          JavaScript Issues with Real-Time Processing and Synthesis:
-        </h3>While processing audio in JavaScript, it is extremely challenging
-        to get reliable, glitch-free audio while achieving a reasonably
-        low-latency, especially under heavy processor load.
-        <ul>
-          <li>JavaScript is very much slower than heavily optimized C++ code
-          and is not able to take advantage of SSE optimizations and
-          multi-threading which is critical for getting good performance on
-          today's processors. Optimized native code can be on the order of
-          twenty times faster for processing FFTs as compared with JavaScript.
-          It is not efficient enough for heavy-duty processing of audio such as
-          convolution and 3D spatialization of large numbers of audio sources.
-          </li>
-          <li>setInterval() and XHR handling will steal time from the audio
-          processing. In a reasonably complex game, some JavaScript resources
-          will be needed for game physics and graphics. This creates challenges
-          because audio rendering is deadline driven (to avoid glitches and get
-          low enough latency).
-          </li>
-          <li>JavaScript does not run in a real-time processing thread and thus
-          can be pre-empted by many other threads running on the system.
-          </li>
-          <li>Garbage Collection (and autorelease pools on Mac OS X) can cause
-          unpredictable delay on a JavaScript thread.
-          </li>
-          <li>Multiple JavaScript contexts can be running on the main thread,
-          stealing time from the context doing the processing.
-          </li>
-          <li>Other code (other than JavaScript) such as page rendering runs on
-          the main thread.
-          </li>
-          <li>Locks can be taken and memory is allocated on the JavaScript
-          thread. This can cause additional thread preemption.
-          </li>
-        </ul>The problems are even more difficult with today's generation of
-        mobile devices which have processors with relatively poor performance
-        and power consumption / battery-life issues.
-      </section>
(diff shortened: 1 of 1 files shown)


hoch:
Hmm. Are we moving this to the separate 'explainer' or 'introduction' doc? Or simply throwing away?

index.html
```diff
@@ -10353,46 +10353,6 @@ <h2 id="Performance">
...
-        to get reliable, glitch-free audio while achieving a reasonably
-        low-latency, especially under heavy processor load.
-        <ul>
-          <li>JavaScript is very much slower than heavily optimized C++ code
-          and is not able to take advantage of SSE optimizations and
-          multi-threading which is critical for getting good performance on
-          today's processors. Optimized native code can be on the order of
-          twenty times faster for processing FFTs as compared with JavaScript.
-          It is not efficient enough for heavy-duty processing of audio such as
-          convolution and 3D spatialization of large numbers of audio sources.
```
padenot: This is not true anymore. Web Assembly or asm.js are in the same ballpark as optimized C, can use SIMD, and can use multiple threads.

index.html
```diff
@@ -10353,46 +10353,6 @@ <h2 id="Performance">
...
-          </li>
-          <li>setInterval() and XHR handling will steal time from the audio
-          processing. In a reasonably complex game, some JavaScript resources
-          will be needed for game physics and graphics. This creates challenges
-          because audio rendering is deadline driven (to avoid glitches and get
-          low enough latency).
```
padenot: This is not true anymore because the JS will run on a different thread than the main thread.

index.html
```diff
@@ -10353,46 +10353,6 @@ <h2 id="Performance">
...
-          </li>
-          <li>JavaScript does not run in a real-time processing thread and thus
-          can be pre-empted by many other threads running on the system.
-          </li>
```
padenot: Not true anymore, js can run on the real-time thread with a worklet.

index.html
```diff
@@ -10353,46 +10353,6 @@ <h2 id="Performance">
...
-          <li>Garbage Collection (and autorelease pools on Mac OS X) can cause
-          unpredictable delay on a JavaScript thread.
```
padenot: It's perfectly possible to write js that does not GC. In fact, that's recommended. emscripten outputs js that does not gc.

index.html
```diff
@@ -10353,46 +10353,6 @@ <h2 id="Performance">
...
-          </li>
-          <li>Multiple JavaScript contexts can be running on the main thread,
-          stealing time from the context doing the processing.
```
padenot: Again, JS won't run on the main thread anymore.

index.html
```diff
@@ -10353,46 +10353,6 @@ <h2 id="Performance">
...
-          </li>
-          <li>Other code (other than JavaScript) such as page rendering runs on
-          the main thread.
```
padenot: This is less and less true in modern browsers.

index.html
```diff
@@ -10353,46 +10353,6 @@ <h2 id="Performance">
...
-          </li>
-          <li>Locks can be taken and memory is allocated on the JavaScript
-          thread. This can cause additional thread preemption.
```
padenot: This is something to avoid when writing js for audio rendering, and in practice you pre-allocate the memory.

index.html
```diff
@@ -10353,46 +10353,6 @@ <h2 id="Performance">
...
-          </li>
-        </ul>The problems are even more difficult with today's generation of
-        mobile devices which have processors with relatively poor performance
-        and power consumption / battery-life issues.
```
padenot: My phone is probably more powerful now than the computer on which this was written.

hoch:
Fair enough! I still think we need some sort of 'guidance' or 'note' about using API more efficiently, but I don't think that should be a part of the spec. A separate document like 'Introduction to Web Audio API' might be nice. (perhaps something like [this](https://www.w3.org/TR/2013/WD-components-intro-20130606/).)

lgtm

padenot:
Some bits of <https://padenot.github.io/web-audio-perf/> could be useful, and maybe we should extend it with information on how to write JavaScript that makes the JIT compiler happy. This will become more and more important when AudioWorklet will ship.

PR #999, opened 2016-09-27
Fix #968: special cases for channelCount and channelCountMode

### LINKED ISSUE #968: AudioNode.channelCount needs to describe special cases
As mentioned in https://groups.google.com/a/chromium.org/d/msg/blink-dev/J8XLxEnkLP0/MuwkC40kAQAJ, we can't really specify the special handling of the channelCount mode in the way that's specified; it's not possible in WebIDL.

Instead, in the description of channelCount, we need to specify here all the special cases, mentioned in https://groups.google.com/a/chromium.org/d/msg/blink-dev/J8XLxEnkLP0/MkSJpENGAQAJ

Perhaps it would be good enough to just list the nodes that have special cases and link to the relevant section of the spec that describe the behavior for the node.

rtoy:
Document the nodes that have special cases for the channelCount and channelCountMode.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2711,57 +2711,76 @@ function setupRoutingGraph() {
               constraints</dfn> on the possible values for the channel count:
             </p>
             <dl>
-              <dt><a>AudioDestinationNode</a></dt>
+              <dt>
+                <a>AudioDestinationNode</a>
+              </dt>
               <dd>
                 <p>
                   The behavior depends on whether the destination node is the
-                  destination of an <a>AudioContext</a> or <a>OfflineAudioContext</a>
+                  destination of an <a>AudioContext</a> or
+                  <a>OfflineAudioContext</a>
                 </p>
                 <dl>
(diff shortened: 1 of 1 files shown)


rtoy:
PTAL. There are some things missing, but it would be useful to have this in place before adding the special cases that have been recently defined (channelSplitter and convolver).

index.html
```diff
@@ -2722,6 +2799,64 @@ <h2 id="OfflineAudioCompletionEvent">
...
+              mode:
+            </p>
+            <dl>
+              <dt>
+                <a>AudioDestinationNode</a>
+              </dt>
+              <dd>
+                If the <a>AudioDestinationNode</a> is the <a href=
+                "#widl-BaseAudioContext-destination"
+                destination=""></a> node of an <a>OfflineAudioContext</a>, then
```
padenot: Markup looks weird here.

padenot:
This looks good for now.

PR #1001, opened 2016-09-27
Fix #977: clean up description of maxChannelCount

### LINKED ISSUE #977: AudioDestinationNode.maxChannelCount discusses channelCount
In https://webaudio.github.io/web-audio-api/#widl-AudioDestinationNode-maxChannelCount, the description includes lots of miscellaneous information not really related to `maxChannelCount`. It talks about the default `channelCount` and the valid range for it.

Then it goes on to talk about `channelCount` for an offline context and that it can't be changed, but doesn't say what it is set to.

I don't think this is the right place to say these things.  These should be moved out.  The constraints should be moved to the AudioNode channelCount section (see #968).

And as it currently stands, the default `channelCount`, `channelCountMode`, and `channelInterpretation` is the same for an destination node for an offline context.  This is not right (see #824 and #856).

rtoy:
Move the description of the channelCount mode to a different section.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2865,10 +2865,47 @@ function setupRoutingGraph() {
       numberOfInputs  : 1
       numberOfOutputs : 1
 
-      channelCount = 2;
-      channelCountMode = "explicit";
-      channelInterpretation = "speakers";
 </pre>
+        <p>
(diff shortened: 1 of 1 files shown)


padenot:
LGTM.

PR #1008, opened 2016-09-28
Spec what happens when the offset of the start() method is greater than the buffer.duration

padenot:
it's clamped if a buffer has been set, set to 0 otherwise.

### DIFF
--- a/index.html
+++ b/index.html
@@ -4451,7 +4451,7 @@ param.setValueCurveAtTime(curve, t7, t8 - t7);
                 <code>offset</code> is negative.</span> If <code>offset</code>
                 is greater than <code>loopEnd</code>, playback will begin at
                 <code>loopEnd</code> (and immediately loop to
-                <code>loopStart</code>). <code>offset</code> is silentely
+                <code>loopStart</code>). <code>offset</code> is silently
                 clamped to [0, <code>duration</code>], when
                 <code>startTime</code> is reached, where <code>duration</code>
                 is the value of the <code>duration</code> attribute of the
--- a/index.html
+++ b/index.html
@@ -4451,11 +4451,16 @@ param.setValueCurveAtTime(curve, t7, t8 - t7);
                 <code>offset</code> is negative.</span> If <code>offset</code>
                 is greater than <code>loopEnd</code>, playback will begin at
                 <code>loopEnd</code> (and immediately loop to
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -4432,11 +4432,16 @@ <h2 id="AudioBufferSourceNode">
...
-                exact sample frame offset within the buffer by multiplying by
-                the buffer's sample rate and rounding to the nearest integer
-                value. Thus its behavior is independent of the value of the
-                <a href=
+                <code>loopStart</code>). <code>offset</code> is clamped to [0,
+                <code>duration</code>], where <code>duration</code> is the
+                value of the <code>duration</code> attribute of the
+                <code>AudioBuffer</code> set to the <code>buffer</code>
+                attribute of this <code>AudioBufferSourceNode</code>, 0 if the
+                <code>buffer</code> attribute has not been set. This parameter
```
rtoy: Based on this text, if I do `start(currentTime + 1000, 1)` without a buffer set, the offset is clamped to 0.  Even if within the next 1000 sec I assign a 10 sec buffer to the ABSN.

Is that intended?

padenot: Kind of. I could not find something obvious to do. Maybe it would be better to do the clamping at the time when the audio is actually starting (currentTime + 1000 in your example), and hope that a buffer has been set ? Maybe it would be better ?

rtoy: I think this is what chrome does. If the buffer hasn't been set by the start time, there's nothing to play, so the node should stop instantly.

padenot:
@rtoy, updated.

index.html
```diff
@@ -4451,11 +4451,16 @@ <h2 id="AudioBufferSourceNode">
                 <code>offset</code> is negative.</span> If <code>offset</code>
                 is greater than <code>loopEnd</code>, playback will begin at
                 <code>loopEnd</code> (and immediately loop to
-                <code>loopStart</code>). This parameter is converted to an
-                exact sample frame offset within the buffer by multiplying by
-                the buffer's sample rate and rounding to the nearest integer
-                value. Thus its behavior is independent of the value of the
-                <a href=
+                <code>loopStart</code>). <code>offset</code> is silentely
```
rtoy: "silentely" -> "silently"

padenot:
Should be good now.

PR #1009, opened 2016-09-28
Add language describing the no-connected-inputs condition observable in an AWP

### LINKED ISSUE #970: Derive tail time from input connection status in AudioWorkletNode
As noted in https://github.com/WebAudio/web-audio-api/pull/959#discussion_r77810324, we need to have a way of telling the AudioWorkletNode if there are any inputs connected.  Without this, any tail time processing for the node can't be implemented correctly.

I can see this being indicated in at least two ways.  As @joeberkovitz mentions, we can have `contextInfo.connected`.  Or maybe `process(null, outputs)`; that is, instead of an input array, we pass in `null`.

joeberkovitz:
Fixes #970

### DIFF
--- a/index.html
+++ b/index.html
@@ -5279,17 +5279,35 @@ class MyProcessor extends AudioWorkletProcessor {
             </p>
             <ol>
               <li>
-                <code>input</code> of type
-                <code>sequence&lt;sequence&lt;Float32Array&gt;&gt;</code><br>
-                The input audio buffer from the incoming connections provided
-                by the user agent. <code>inputs[n][m]</code> is a
-                <code>Float32Array</code> of audio samples for the
-                <code>m</code>th channel of <code>n</code>th input. While the
-                number of inputs is fixed at the construction, the number of
-                channels can be changed dynamically.
+                <p>
+                  <code>inputs</code> of type
+                  <code>sequence&lt;sequence&lt;Float32Array&gt;&gt;</code><br>
+                  The input audio buffer from the incoming connections provided
+                  by the user agent. <code>inputs[n][m]</code> is a
+                  <code>Float32Array</code> of audio samples for the
+                  <code>m</code>th channel of <code>n</code>th input. While the
+                  number of inputs is fixed at the construction, the number of
+                  channels can be changed dynamically.
+                </p>
+                <p>
+                  If no connections exist to the <code>n</code>th input of the
+                  node during the current render quantum, then the content of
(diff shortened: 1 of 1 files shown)


joeberkovitz:
You can review this at https://rawgit.com/WebAudio/web-audio-api/85a1c03e0926a0e16703771b5b19d7d65f59982c/index.html#defining-a-valid-audioworkletprcoessor

joeberkovitz:
@rtoy @padenot PTAL. I split this off from the other PR detailing how the return value from `process()` works. The decision to use "zero channels" to indicate a disconnected input was motivated by the fact that if no inputs are connected, the algorithm to determine a "computed number of channels" is no longer well defined and may well result in zero channels at least some of the time (except when `channelCountMode` is `"explicit"`). So it felt better to just say that the channel count of a disconnected input appears to be zero, as observed by an AudioWorkletProcessor.

index.html
```diff
@@ -5279,17 +5279,35 @@ <h2 id="defining-a-valid-audioworkletprcoessor">
...
-                <code>input</code> of type
-                <code>sequence&lt;sequence&lt;Float32Array&gt;&gt;</code><br>
-                The input audio buffer from the incoming connections provided
-                by the user agent. <code>inputs[n][m]</code> is a
-                <code>Float32Array</code> of audio samples for the
-                <code>m</code>th channel of <code>n</code>th input. While the
-                number of inputs is fixed at the construction, the number of
-                channels can be changed dynamically.
+                <p>
+                  <code>inputs</code> of type
```
rtoy: "inputs"?  There's only one input to `process()`, right?

padenot:
Same, works for me.

index.html
joeberkovitz: There can be multiple inputs to a node and the parameter is actually called `inputs` already in the signature. So I thought the plural form made sense.

rtoy: Oh, sorry. You're referring to the actual parameter name.  (Is it possible to use markup like we do for other functions? I think that's what confused me.)

joeberkovitz: I don't think it is possible to use the usual dd/dt markup, because this is not IDL (as per TAG feedback).

rtoy: Oh, I didn't mean that you need use markup in this PR!

hoch: I simply used 'code' tag to describe this is an argument of this method. This is because, as Joe said, we cannot use the formal IDL notation here.

PR #1010, opened 2016-09-28
Fix #850: Allow setTargetAtTime timeConstant to be 0

### LINKED ISSUE #850: In case of setTargetAtTime's timeConstant is zero
The spec defines that the timeConstant of AudioParam.setTargetAtTime() is strictly positive or a TypeError exception must be thrown.

But it is more intuitive and convenient if it change the value immediately when the timeConstant is zero.

##
Just for information, current browsers cannot handle timeConstant=0 normally. it will not throw exception. but it stop sound generation or cause incomprehensible error.

rtoy:
Change the text to say that it must be non-negative instead of
strictly positive.

The formula when the time constant is 0 is pretty well-defined if the limit
is taken, so no text is added to indicate what happens.

### DIFF
--- a/index.html
+++ b/index.html
@@ -3536,8 +3536,8 @@ function setupRoutingGraph() {
                 The time-constant value of first-order filter (exponential)
                 approach to the target value. The larger this value is, the
                 slower the transition will be. <span class="synchronous">The
-                value must be strictly positive or a TypeError exception MUST
-                be thrown.</span>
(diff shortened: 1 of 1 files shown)


padenot:
I agree it's well defined if you take the limit, but it's still a division by zero when you compute it with a computer. We can simply add a note saying that a time constant of zero means that the transition is immediate.

PR #1013, opened 2016-09-29
Fix #997: Add defaults to dictionary members

### LINKED ISSUE #997: Inconsistent use of default values in Options dictionaries
We are inconsistent in whether default values are specified in the option dictionaries for each node. It was initially done this way because I didn't want to have two places specifying the default, but perhaps we should reconsider this and add default values for all dictionaries members that have defaults.

It's unfortunate to have two places, but I think dictionaries make it much clearer and is the future.

rtoy:
Defaults added for all dictionary members that could be defaulted.
Also fixed a couple of places where the member could be null, like for
the convolver buffer, the wave shaper curve, the oscillator
periodicwave wave.

### DIFF
--- a/index.html
+++ b/index.html
@@ -3897,7 +3897,7 @@ param.setValueCurveAtTime(curve, t7, t8 - t7);
           </p>
           <dl title="dictionary GainOptions : AudioNodeOptions" class="idl">
             <dt>
-              float gain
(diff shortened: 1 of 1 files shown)


rtoy:
Closing.  This is fixed in #1016.

PR #1016, opened 2016-09-30
Define `AudioNode` constructors

padenot:
@rtoy, PTAL

### DIFF
--- a/index.html
+++ b/index.html
@@ -2330,16 +2330,17 @@ function setupRoutingGraph() {
           sample-frames which corresponds to roughly 3ms at a sample-rate of
           44.1KHz.
         </p><a>AudioNode</a>s can be created in two ways:
-        <ul>
-          <li>First, an <a>AudioNode</a> can be created by calling the
+        <ol>
+          <li>An <a>AudioNode</a> can be created by calling the
           corresponding <dfn>factory method</dfn> on the <a>AudioContext</a>,
           passing in arguments. The <a>AudioNode</a> is associated with the <a>
             AudioContext</a> it was created from.
           </li>
-          <li>Second, an <a>AudioNode</a> can be create by calling its
+          <li>An <a>AudioNode</a> can be created by calling its
           constructor, passing in the <a>AudioContext</a> with which the
-          <a>AudioNode</a> being constructed will be associated with. A second
-          argument is a dictionnary that can contain attributes specific to the
+          <a>AudioNode</a> being constructed will be associated with. An
+          optional second
+          argument is a dictionary that can contain attributes specific to the
           type of <a>AudioNode</a> that is being created. When creating an <a>
             AudioNode</a> using the constructor, the following algorithm MUST
             be executed atomically:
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2329,7 +2329,68 @@ <h2 id="OfflineAudioCompletionEvent">
           value explicitly. <em>block-size</em> is defined to be 128
           sample-frames which corresponds to roughly 3ms at a sample-rate of
           44.1KHz.
-        </p>
+        </p><a>AudioNode</a>s can be created in two ways:
+        <ul>
```
rtoy: Since you say "First", "Second" below for each case, I think you should either make this an ordered list and remove "First", "Second" below.

index.html
```diff
@@ -2329,7 +2329,68 @@ <h2 id="OfflineAudioCompletionEvent">
...
+          <li>First, an <a>AudioNode</a> can be created by calling the
+          corresponding <dfn>factory method</dfn> on the <a>AudioContext</a>,
+          passing in arguments. The <a>AudioNode</a> is associated with the <a>
+            AudioContext</a> it was created from.
+          </li>
+          <li>Second, an <a>AudioNode</a> can be create by calling its
```
rtoy: s/create/created/

index.html
```diff
@@ -2329,7 +2329,68 @@ <h2 id="OfflineAudioCompletionEvent">
...
+          constructor, passing in the <a>AudioContext</a> with which the
+          <a>AudioNode</a> being constructed will be associated with. A second
+          argument is a dictionnary that can contain attributes specific to the
```
rtoy: s/dictionnary/dictionary/

Should we also say that the second argument is (sometimes) optional?

index.html
```diff
@@ -2329,7 +2329,68 @@ <h2 id="OfflineAudioCompletionEvent">
...
+              </li>
+              <li>If the <a>AudioNode</a> being constructed is a
+              <a>ConvolverNode</a>, set its<a href=
+              "#widl-ConvolverNode-normalize">normalize</a> attribute with the
+              inverse of the value of the <a href=
+              "#widl-ConvolverOptions-disableNormalization">disableNormalization</a>
+              dictionnary member, and then set its the <a href=
+              "#widl-ConvolverNode-buffer">buffer</a> attribute to the value of
+              the <a href="#widl-ConvolverOptions-buffer">buffer</a>
+              dictionnary member, in this order, and jump to the last step of
```
rtoy: s/dictionnary/dictionary/

index.html
```diff
@@ -3879,8 +3940,7 @@ <h2 id="GainNode">
           </dt>
           <dd>
             <p>
-              Represents the amount of gain to apply. Its default
-              <code>value</code> is 1 (no gain change). This parameter is
```
rtoy: I'm ok with adding defaults to the dictionary members, but I'm not sure I want it removed from the attribute definition too.  Now the reader has to magically know to go look at the dictionary definitions to get the default value.

padenot: I agree. I thought about keeping them, but having the same info in two locations is also problematic. I can revert part of this commit.

padenot:
@rtoy, comments addressed, I've reverted part of 339bd8d, adding the textual description back.

index.html
rtoy: Yeah, I didn't like specifying the defaults in two places, but what can we do now?

padenot:
Agreed, last patch harmonizes this.

PR #1017, opened 2016-09-30
Fix #960: Add AudioScheduledSourceNode base class for source nodes

### LINKED ISSUE #960: Add AudioSourceNode base class to capture common start()/stop() methods
The AudioBufferSourceNode, OscillatorNode, and the upcoming ConstantSourceNode all share common `start()` and `stop()` methods.  Factor this out in an `AudioSourceNode` (name TBD?) base class so we can put all of this information on one place instead of having nodes refer to AudioBufferSourceNode.

See https://github.com/WebAudio/web-audio-api/pull/902#discussion_r72907030 for the initial idea.

rtoy:
The AudioScheduledSourceNode is a base class for AudioBufferSourceNode,
ConstantSourceNode, and OscillatorNode.  This class provides the
basic start() and stop() methods as well as the onended attribute that
these source nodes support.


rtoy:
@joeberkovitz PTAL.

index.html
```diff
@@ -4666,7 +4773,7 @@ <h2 id="AudioBufferSourceNode">
             </p>
             <dl class="parameters">
               <dt>
-                optional double when = 0
+                double when
```
joeberkovitz: I don't get why we are leaving `start()`in this interface if it's inherited from `AudioScheduledSourceNode`. And why would `when` no longer be optional for an ABSN's implementation of this method?

rtoy: Yeah, this is confusing because you only see the diff.  `AudioScheduledSource` has the basic `start(when)` function that is applicable to all subclasses today.  But `AudioBufferSourceNode` has an additional `start(when, offset, duration)` method.  This change just makes `when` required for this 3-arg case. I guess this isn't really necessary, but it does make it slightly clearer which `start` method is being called when you say `start()`.

joeberkovitz: I see, thanks! In that case, LGTM

Issue #452, opened 2014-12-03
(CircularRouting): Circular graphs should exist while observable?


mark-buer:
Should cyclic graph fragments stay alive while they are observable through *any* means?

[The spec states](http://www.w3.org/TR/webaudio/#lifetime-AudioNode):
> Any AudioNodes which are connected in a cycle and are directly or indirectly connected to the AudioDestinationNode of the AudioContext will stay alive as long as the AudioContext is alive.

But, the output of the cycle may be observable via a MediaStreamAudioDestinationNode or an AnalyserNode.

What happens in these cases?

joeberkovitz:
Resolution: fix the spec to cite MediaStreamAudioDestinationNode as well as AudioDestinationNode (they don't share a super-interface).

PR #1019, opened 2016-09-30
Define circular routing lifetime correctly

joeberkovitz:
This addresses #452 which pointed out that MediaStreamAudioDestinationNode also creates a valid sink causing attached cyclic graphs to be retained.


joeberkovitz:
@rtoy @padenot PTAL

Issue #966, opened 2016-09-09
[TAG] Clarify the relationship/ownership of AudioWorkletGlobalScope 


hoch:
The number of potential `AudioWorkletGlobalScope` instances that can be running at the same time doesn't seem to be spelled out anywhere. The example code at the front of the section makes me think that it could be any of:
- One per AudioWorkletNode
- One per "Main global scope"
- One per AudioWorkletProcessor subclass type

...although I assume it's the first?

The issued is raised by @slightlyoff at #869.

hoch:
I believe each `AudioContext` should have one `AudioWorkletGlobalScope`. This conflicts with the case of other Worklet variants, because they use a single WorkletGlobalScope for a frame.

Or we can be bold and use a single `AudioWorkletGlobalScope` for all the AudioContext in the same frame. Let developers do the inter-context communication!!

padenot:
`AudioNode`s, as they are today, don't really communicate in an observable way with each other. However, some state is shared between them (mainly large buffers, for exampe, wave tables), to optimize memory and CPU usage. 

Considering one of the design goal for `AudioWorklet` is to be able to reimplement the native nodes, we should make it so that it's possible to implement those optimizations.

It would be acceptable to be able to share a read-only `ArrayBuffer` slice ? I don't know what the DOM concept for this exists or has been implemented yet ? Something like this: <https://gist.github.com/dherman/5401735>.

hoch:
`AudioBuffer` is not associated with any `BaseAudioContext`. So sharing an AudioBuffer object between `AudioWorkletNode`s created by different audio contexts should not be an issue. Or do you mean that `AudioWorkletGlobalScope` can be shared by multiple `AudioContext`s? Or do we want to be able to share `ArrayBuffer`s through the `WorkletGlobalScope`?

Before taking a look into the link you posted, I wanted to clarify your intention first.

padenot:
I meant that we should ensure that big buffers can be shared between `AudioWorkletNode`s. This should ideally be supported for things other than `AudioBuffer`.

For now, if `postMessage` works the same to a worklet than it does to a worker, an `ArrayBuffer` that is sent through `postMessage` is detached, and can't be sent again, iirc.

hoch:
Yes, I wanted to discuss that in TPAC F2F: __Do we really want to support the complete `postMessage` mechanism like Worker?__

[...]

PR #1020, opened 2016-09-30
Clarify relationships between AudioWorkletGlobalScope and AudioContext.

joeberkovitz:
This is a fix for #966.


joeberkovitz:
@rtoy @hoch @padenot PTAL.
Reviewers please see:
- https://rawgit.com/WebAudio/web-audio-api/c8645ea156620e15a7765be688b0890fa744d1a4/index.html#AudioWorkletGlobalScope
- https://rawgit.com/WebAudio/web-audio-api/c8645ea156620e15a7765be688b0890fa744d1a4/index.html#AudioWorklet-Sequence

index.html
```diff
@@ -5543,26 +5544,42 @@ <h2 id="AudioWorklet-Sequence">
...
-            <li>As part of the global scope's initialization, the set of imported module scripts is run, including the one that was previously imported.</li>
-            <li>As part of running the imported script, an <a>AudioWorkletProcessor</a> is registered under the
-            key <code>"Custom1"</code> within the <a>AudioWorkletGlobalScope</a>.</li>
-            <li>In the main scope, an <a>AudioWorkletNode</a> is created using the key <code>"Custom1"</code> along
-            with an <code>opts</code> dictionary of options.</li>
-            <li>As part of the node's creation, this key is used to look up the correct <a>AudioWorkletProcessor</a>
-            subclass for instantiation.</li>
-            <li>An instance of the <a>AudioWorkletProcessor</a> subclass is instantiated with a structured clone of the same
-            <code>opts</code> dictionary.</li>
+            <li>In the main scope, <code>window.audioWorklet</code> is
```
hoch: I am not entirely sure why this non-normative section is necessary. Being user-friendly is certainly nice, but this information is too non-normative and the algorithm above is good enough to explain what's happening technically. @rtoy @padenot WDYT?

joeberkovitz:
@rtoy @padenot I would like to get this PR merged and I've only heard from @hoch so far. Although a separate intro doc would be nice, I think that can be deferred.

index.html
```diff
@@ -4998,6 +5000,18 @@ <h2 id="AudioWorkletGlobalScope">
...
             is internal and is not directly exposed to the user.
           </p>
+          <div class="note">
+            The <a>AudioWorkletGlobalScope</a> may also contain any other data
+            and code to be shared by these instances. As an example, multiple
+            processors might share an ArrayBuffer defining a wavetable or an
+            impulse response.
+          </div>
+          <div class="note">
+            Every <a>AudioWorkletGlobalScope</a> is associated with at most one
```
rtoy: How is this enforced (association to at most one AudioContext)? Can't seem to find a link for that, so if it exists a link to that would be good.

index.html
```diff
@@ -5510,14 +5524,65 @@ <h2 id="instantiation-of-AudioWorkletNode-and-AudioWorkletProcessor">
...
+          <h2 id="AudioWorklet-Sequence">
+            AudioWorklet Sequence of Events
+          </h2>
+          <p>
+            The following figure illustrates an idealized sequence of events
+            occurring relative to an <a>AudioWorklet</a>:
+          </p>
+          <figure>
+            <img alt="AudioWorklet sequence" src=
+            "images/audioProcessorInstantiation.png" width="917" height="546">
```
rtoy: Would this figure be better if it really were a sequence diagram?

index.html
```diff
@@ -4987,8 +4977,20 @@ <h2 id="AudioWorkletGlobalScope">
             This special execution context is designed to enable the
             generation, processing, and analysis of audio data directly using
             JavaScript in the audio <a>rendering thread</a>. The user-supplied
-            script code is evaluated in this scope to define an
-            <a>AudioWorkletProcessor</a>.
+            script code is evaluated in this scope to define one or more
+            <a>AudioWorkletProcessor</a> subclasses, which in turn are used to
+            instantiate <a>AudioWorkletProcessor</a>s, in 1:1 association with
```
joeberkovitz: done

index.html
joeberkovitz: This actually _is_  the statement of the principle to be enforced. It's what is required to satisfy some of the Worklet motivations described in https://drafts.css-houdini.org/worklets/#motivations, namely that global scopes may be spun up in a rendering environment as needed. Each scope must be able to run consistently, without races.

index.html
joeberkovitz: I considered that, but I think this is somewhat more effective in that it shows the relationship of the various pieces to each other.

Certainly open to changing it based on further review.

joeberkovitz: @rtoy is there further need for changes or may I merge this?

joeberkovitz:
@hoch, @rtoy asked me to ask you to take a look at this :) Want to merge it if possible...

index.html
rtoy: The picture is pretty complex, but I think a sequence diagram would also be complex.

I'm fine with this as it is.  If it proves to confusing, someone will file and issue and we can fix it then.

index.html
```diff
@@ -4987,8 +4977,20 @@ <h2 id="AudioWorkletGlobalScope">
...
-            <a>AudioWorkletProcessor</a>.
+            script code is evaluated in this scope to define one or more
+            <a>AudioWorkletProcessor</a> subclasses, which in turn are used to
+            instantiate <a>AudioWorkletProcessor</a>s, in a 1:1 association
+            with <a>AudioWorkletNode</a>s in the main scope.
+          </p>
+          <p>
+            At least one <a>AudioWorkletGlobalScope</a> exists for each
+            <a>AudioContext</a> that contains one or more
+            <a>AudioWorkletNode</a>s. The running of imported module scripts is
```
hoch: Can we leave 'module' out? Or do you have a reference this is actually the 'module' pattern?

index.html
hoch: I feel like the wording is slightly confusing. I believe we need to highlight that the one-to-one relationship between an BaseAudioContext and an AWGS. I think we have already decided that we do not support per-node global scope. No?

By the way, it's better to use BaseAudioContext everywhere instead of AudioContext.

index.html
hoch: Agreed. We land the diagram as it is and can fix it later if necessary. When we freeze the spec for CR, I would like to rework on every diagram in the spec.

index.html
```diff
@@ -5510,14 +5524,65 @@ <h2 id="instantiation-of-AudioWorkletNode-and-AudioWorkletProcessor">
...
+          </h2>
+          <p>
+            The following figure illustrates an idealized sequence of events
+            occurring relative to an <a>AudioWorklet</a>:
+          </p>
+          <figure>
+            <img alt="AudioWorklet sequence" src=
+            "images/audioProcessorInstantiation.png" width="917" height="546">
+            <figcaption>
+              <a><code>AudioWorklet</code></a> sequence
```
hoch: `<code>` tags are redundant. Or it doesn't work in the `<figcaption>`?

index.html
```diff
@@ -5510,14 +5524,65 @@ <h2 id="instantiation-of-AudioWorkletNode-and-AudioWorkletProcessor">
...
+            <li>An <a>AudioContext</a> is created.
+            </li>
+            <li>An <a>AudioWorkletGlobalScope</a> is created in association
+            with the context's audio rendering thread.
+            </li>
+            <li>As part of the global scope's initialization, the set of
+            imported module scripts is run, including the one that was
+            previously imported.
+            </li>
+            <li>As part of running the imported script, an
```
hoch: Consider this scenario:

1. worklet.import('moduleA');
2. Context A created. (with AWGS A)
3. worklet.import('moduleB');
4. Context B created. (with AWGS B)

What is the status of definition map in AWGS A and AWGS B? Are they supposed to be same or different?

index.html
```diff
@@ -5510,14 +5524,65 @@ <h2 id="instantiation-of-AudioWorkletNode-and-AudioWorkletProcessor">
...
+            <a>AudioWorkletProcessor</a> is registered under the key
+            <code>"Custom1"</code> within the <a>AudioWorkletGlobalScope</a>.
```
hoch: Not sure about using an arbitrary keyword `Custom1` in the spec. We might have to consult the spec specialist here. (I have a strong feeling that she/he will encourage us to use the ECMA style algorithm for this.)

index.html
```diff
@@ -5510,14 +5524,65 @@ <h2 id="instantiation-of-AudioWorkletNode-and-AudioWorkletProcessor">
...
+            <li>In the main scope, an <a>AudioWorkletNode</a> is created using
+            the key <code>"Custom1"</code> along with an <code>opts</code>
+            dictionary of options.
+            </li>
+            <li>As part of the node's creation, this key is used to look up the
+            correct <a>AudioWorkletProcessor</a> subclass for instantiation.
+            </li>
+            <li>An instance of the <a>AudioWorkletProcessor</a> subclass is
+            instantiated with a structured clone of the same <code>opts</code>
+            dictionary.
```
hoch: When I reached here, this section reads like a rough description of 'instantiation process'. Perhaps we should focus on explaining relationship between classes.

index.html
```diff
@@ -4987,8 +4977,20 @@ <h2 id="AudioWorkletGlobalScope">
...
             JavaScript in the audio <a>rendering thread</a>. The user-supplied
-            script code is evaluated in this scope to define an
-            <a>AudioWorkletProcessor</a>.
+            script code is evaluated in this scope to define one or more
+            <a>AudioWorkletProcessor</a> subclasses, which in turn are used to
+            instantiate <a>AudioWorkletProcessor</a>s, in a 1:1 association
+            with <a>AudioWorkletNode</a>s in the main scope.
+          </p>
+          <p>
+            At least one <a>AudioWorkletGlobalScope</a> exists for each
```
joeberkovitz: I don't think it makes any difference how many AWGSs are created, since if there is more than one, the effects of multiple AWGSs must not be observable. We are not mandating any sort of parallel processing here, but also not ruling it out provided that its effects are unobservable (which is important for the future).

However I believe this is in fact normative language because of the requirement of "At least one". What _is_ observable is that at least one AWGS exists, allowing processing to take place at all.

joeberkovitz: I guess the non-observability of multiple AWGS is also normative: we are formally stating that they are not allowed to have observable effects.

index.html
joeberkovitz: Agreed, I'm removing `module`.

index.html
joeberkovitz: They would be the same. moduleB gets added to AWGS A, and moduleA is placed into AWGS B prior to moduleB. This is just regular Worklet behavior.

index.html
joeberkovitz: Let's mark that up when it happens.

index.html
joeberkovitz: I don't understand what relationship you want to explain - can you elaborate?

joeberkovitz:
@hoch I've responded to your comments and my hope is that this is ready to merge - PTAL. Thanks!

PR #1021, opened 2016-09-30
Fix #942: Clarify behavior of convolver node

### LINKED ISSUE #942: ConvolverNode behavior underspecified and inconsistent
From the [diagram](https://webaudio.github.io/web-audio-api/#Convolution-channel-configurations), it looks like there are six possible configurations of the `ConvolverNode`.  Some issues:

- The general N-input/K-convolvers/M-outputs is not possible and should be removed.
- It's impossible to distinguish the 1-input/1-convolver/1-output case from the 1-input/1-convolver/2-output case.  (Chrome and FF apparently both do the latter).  
- The relationship between channelCount and number of channels in the impulse response is not specified.
  - What happens with a stereo input and a mono impulse response?  Chrome outputs silence because it's not well defined.

Proposal:

- Remove the general case from the diagram
- Remove the mono to copied stereo case. (**This is a breaking change!**)
  - Alternatively, remove the mono case, which I think isn't implemented by anyone.
  - I find the mono to copied stereo case really confusing. (I was doing a test with two cascaded convolvers mono response with an oscillator input.  No audio is heard because the 2-channel input and mono response isn't specified.)
- The channelCount of the node is set based on the number of channels in the impulse response. Then channelCount = min(2, channels in impulse response).
 
With these changes, I think it's possible to reason exactly what will happen based on the channelCount, mode, and interpretation.  For example, the case of n-channel input with a one channel response has a channelCount = 1, so the input is mixed to mono, producing a mono output.  Easy!

rtoy:
This specifies how the impulse response works in various
configurations.  The constraints on channelCount and channelCountMode
were applied in #968.


rtoy:
Still need to update the diagrams.

padenot:
What program have you used to generate this image? All the svg renderers I have tried apart from Chrome render something strange when it comes to kerning the figure description, for example 



This in Firefox, but I have a similar rendering in Inkscape or the image viewer of my desktop.

rtoy:
I used Omnigraffle to create the graph and export to svg.  Perhaps it's a font issue? I used whatever deafult Omnigraffle used.  I'll try to update the graph and test with Firefox to make sure it looks right.  If it doesn't work out, I'll save as an png instead of svg.

rtoy:
I changed the font from Helvetica Nueue (default for Omnigraffle?) to Arial.  It looks fine now in Firefox. Please test when you get a chance.

padenot:
Yep looks good, thanks !

PR #1024, opened 2016-10-04
Fix #915 The case for sample-accurate start()

### LINKED ISSUE #915: The case for subsample-accurate start()
Bob: Can I create a sine wave using WebAudio? I only got as far as

```
```

Alice: Of course! You can just add the following:

```
osc.connect(context.destination);
osc.start();
```

Bob:  Hey, that's cool.  But I actually want the first quarter wave of the signal to be zero.  Can we do that?
Alice:  That's not hard either.  We can just add a gain node to zero out that part like so:

```
var g = context.createGain();
g.gain.setValueAtTime(0, 0);
var quarterWaveTime = (1/4)/osc.frequency.value;
g.gain.setValueAtTime(1, quarterWaveTime);
osc.connect(g);
g.connect(context.destination);
osc.start();
```

Bob: That's exactly what I wanted!  Thanks so much!  But you know, that looks like a cosine wave that starts after a quarter wave.  Can I do that too with WebAudio?
Alice:  Sure.  You have to use a custom oscillator to create a cosine wave, but it's not hard:

```
var osc2 = context.createOscillator();
var wave = context.createPeriodicWave(Float32Array.from([0, 1]), Float32Array.from([0, 0]));
osc2.setPeriodicWave(wave);
osc2.connect(context.destination);
var quarterWaveTime = (1/4)/osc2.frequency.value;

osc2.start(quarterWaveTime);
```

Bob:  That's so cool!  But is it really same?  How can I tell?
Alice:  Of course it's the same!  Just to prove it we can take the two bits of code, add another gain node and subtract one from the other like so:

```
var g = context.createGain();
g.gain.setValueAtTime(0, 0);
var quarterWaveTime = (1/4)/osc.frequency.value;
g.gain.setValueAtTime(1, quarterWaveTime);
osc.connect(g);

var osc2 = context.createOscillator();
var wave = context.createPeriodicWave(Float32Array.from([0, 1]), Float32Array.from([0, 0]));
osc2.setPeriodicWave(wave);
var quarterWaveTime = (1/4)/osc2.frequency.value;

var inverter = context.createGain();
inverter.gain.value = -1;
osc2.connect(inverter);

g.connect(context.destination);
inverter.connect(context.destination);
osc.start();
osc2.start(quarterWaveTime);
```

Bob:  Thanks Alice, but my machine must be broken.  When I try this with hoch.github.io/canopy and paste your code in and use a sampleRate of 24000, I don't get zero output.  In fact it looks like a sine wave of amplitude 0.04 or so:
screen shot 2016-07-26 at 7 55 06 pm

Alice:  Oh.

What happened?  Assuming that automations are sampled at frame boundaries, which is what Chrome does, the example using gain automation is truly what we want.  But the start time of the oscillator in the second case isn't sample accurate.  The delayed cosine wave started with an output value of 1 at a sample-frame boundary.  if it had started at the actual start time, and we sampled the resulting continuous signal at the sample-frame points, the resulting difference would have been much, much closer to zero.  I'd show a plot, but no browser I know of implements sample-accurate start for oscillators.

This would also a problem for AudioBufferSourceNodes if we don't have sample-accurate start.  If you replaced the oscillators with source nodes with buffers created using `Math.sin` and `Math.cos` as you would expect to use, the difference wouldn't be zero either.  (I leave the reader to test this out;  I haven't actually tried it.)

mdjp:
WG should we explicitly statethat this is not quantised to samples?
eg.
It is in the same time coordinate system as the AudioContext's currentTime attribute.

### DIFF
--- a/index.html
+++ b/index.html
@@ -4444,7 +4444,9 @@ param.setValueCurveAtTime(curve, t7, t8 - t7);
               between 0 and the duration of the buffer. If
               <code>loopStart</code> is less than 0, looping will begin at 0.
               If <code>loopStart</code> is greater than the duration of the
-              buffer, looping will begin at the end of the buffer.
+              buffer, looping will begin at the end of the buffer. Its behavior is 
+              independent of the value of the <a href="#widl-AudioBufferSourceNode-
+              playbackRate"><code>playbackRate</code></a> parameter.
             </p>
           </dd>
           <dt>
@@ -4461,7 +4463,9 @@ param.setValueCurveAtTime(curve, t7, t8 - t7);
               between 0 and the duration of the buffer. If <code>loopEnd</code>
               is less than 0, looping will end at 0. If <code>loopEnd</code> is
               greater than the duration of the buffer, looping will end at the
-              end of the buffer.
+              end of the buffer. Its behavior is independent of the value of 
+              the <a href="#widl-AudioBufferSourceNode-playbackRate">
+              <code>playbackRate</code></a> parameter.
             </p>
           </dd>
           <dt>
@@ -4507,6 +4511,7 @@ param.setValueCurveAtTime(curve, t7, t8 - t7);
                 is the value of the <code>duration</code> attribute of the
                 <code>AudioBuffer</code> set to the <code>buffer</code>
                 attribute of this <code>AudioBufferSourceNode</code>.
+                Its behavior is independent of the value of the <a href="#widl-AudioBufferSourceNode-playbackRate"><code>playbackRate</code></a> parameter.
               </dd>
               <dt>
                 optional double duration
--- a/index.html
+++ b/index.html
@@ -4444,9 +4444,10 @@ param.setValueCurveAtTime(curve, t7, t8 - t7);
               between 0 and the duration of the buffer. If
               <code>loopStart</code> is less than 0, looping will begin at 0.
               If <code>loopStart</code> is greater than the duration of the
-              buffer, looping will begin at the end of the buffer. Its behavior is 
-              independent of the value of the <a href="#widl-AudioBufferSourceNode-
-              playbackRate"><code>playbackRate</code></a> parameter.
+              buffer, looping will begin at the end of the buffer. Its behavior
+              is independent of the value of the <a href=
+              "#widl-AudioBufferSourceNode-%20playbackRate"><code>playbackRate</code></a>
+              parameter.
             </p>
           </dd>
           <dt>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -4444,13 +4444,7 @@ <h2 id="AudioBufferSourceNode">
               between 0 and the duration of the buffer. If
               <code>loopStart</code> is less than 0, looping will begin at 0.
               If <code>loopStart</code> is greater than the duration of the
-              buffer, looping will begin at the end of the buffer. This
-              attribute is converted to an exact sample frame offset within the
-              buffer by multiplying by the buffer's sample rate and rounding to
-              the nearest integer value. Thus its behavior is independent of
-              the value of the <a href=
```
rtoy: This part about the behavior with respect to the `playbackRate` is important and should be left in, here and below.

mdjp: Let me review this again. i was under the impression that we were not quantising to the closest sample. I might have got my wires crossed with this one.

rtoy:
This seems tricky. I'm not sure what the best way to deal with this would be. Removing the rounding to the sample frame is certainly right, but I'm not sure if we need to say anything more.

Perhaps a diagram that illustrates how time and sampling interact would be appropriate.  That would also help explain how time in AudioParams work too. (This is not to say you need to draw this diagram for this PR.)

For the oscillator case, I think it's pretty clear.  Let s(t) be the basic waveform of the oscillator. If you have osc.start(ts), and t0 <= ts < t1, where t0, and t1 are the sample points closest to ts, the output value at t0 is 0.  The output value at t1, must be s(t1-ts), which is typically not 0.

I think the same holds for an ABSN. You have to interpolate between the samples of the buffer.  Thus, the output at time t1 is not exactly equal to the first sample of the AudioBuffer.  It can be a linear interpolation between 0 and s(0), where s(n) are the samples in the AudioBuffer.

Once we have this, I think it's pretty clear how loopStart and loopEnd behave.  (Even if it is actually hard to describe the exact behavior.)

index.html
joeberkovitz: Seconding @rtoy's observation that we have to specify that the loop points (though now unquantized) are defined in terms of the *buffer's* sample rate, not the computed playback rate.

If loop points were evaluated in terms of computed playback rate, the behavior with a modulated playback rate would be chaotic and unpredictable.

joeberkovitz:
@mdjp I think @rtoy's feedback is valid - do you think we can wrap this up this week?

joeberkovitz:
@mdjp is this ready to merge?

joeberkovitz:
@mdjp and I just talked and agree it makes sense to wait until #95 is closed, then this becomes a trivial change to the Oscillator and other schedulable nodes other than ABSN.

joeberkovitz:
We agreed to close this w/o merging since it's superseded by #1155 which fixes #95. A new PR will be created for the general case of AudioSchedulableSourceNodes.

PR #1025, opened 2016-10-05
Make `maxChannelCount` always be the number of output channels, and error out when setting `channelCount` with an invalid value.

padenot:


index.html
```diff
@@ -3158,7 +3158,10 @@ <h2 id="AudioDestinationNode">
           Giving a concrete example, if the audio hardware supports 8-channel
           output, then we may set <a href=
           "#widl-AudioNode-channelCount">channelCount</a> to 8, and render 8
-          channels of output.
+          channels of output. <span class="synchronous">An
+          <code>NotSupportedError</code> MUST be fired if the implementation
+          does not support changing the <a href=
+          "#widl-AudioNode-channelCount">channelCount</a></span>.
```
rtoy: The previous sentence (not shown here) already says we thrown an error for invalid values of `channelCount`.  Are you saying that I can have, say a 7 channel device and I'm not allowed to set `channelCount` to any value from 1 to 7?

rtoy: Can't set the count to 1 and have it upmixed to stereo before going out?  And why is this not considered a bug in the implementation? It seems to me that you should set maxChannelCount to 2 (always). (Chrome on Linux had this issue where maxChannelCount = 2 no matter how many channels the output device actually had.  This was only very recently fixed.)

padenot: Quoting the current spec:

>  maxChannelCount is the maximum number of channels that this hardware is capable 
> of supporting

The hardware can supports more than two, Gecko does not right now. `maxChannelCount` reports the hardware count, not what is available. Of course we can change this, but as it is right now, it's not illegal per spec.

I think it would be more useful to report the number of channels of the hardware on `maxChannelCount` and throw with `NotSupportedError` when setting the `channelCount` to something that the engine does not support, or `IndexSizeError` when it's outside [1, `maxChannelCount`].

I don't feel strongly about this though, a number of solutions are acceptable.

rtoy: Thanks for the explanation; I understand what Gecko is doing.

I'd rather not bake this condition into the spec. I think it would be ok for Gecko to return 2 even if the hardware has more channels (just like Chrome on Linux did until a couple of weeks ago).  And write it off as a bug in the implementation.  I think that's more useful from a user's viewpoint with fewer exceptional cases to consider.

rtoy: Ping.

padenot:
Updated, please have a look.

PR #1027, opened 2016-10-05
Make the ChannelSplitterNode have constant channel* attributes.

### LINKED ISSUE #975: ChannelSplitterNode channelCount/Mode/Intepretation settings?
As currently spec'ed the [ChannelSplitterNode](http://localhost:7100/#the-channelsplitternode-interface) has a default count of 2, mode of "max", and interpretation of "speakers", and these can be changed. (A mode of "max" means the count is ignored, in case you forget, like I did.)

Allowing this complicated behavior for the splitter node complicates how to reason about the splitter's behavior.  It would be much simpler if we just said the count is the number of outputs, mode is "explicit", and interpretation is "discrete".  Then, the inputs are simply summed, and we take just the first n channels to go to the n outputs.

If the complicated behavior is needed, just stick an appropriate gain node before the splitter.

I think there are three alternatives here:

1. Leave things as is (allowing garbage in/garbage out)
1. Change the defaults to number of outputs, "explicit", "discrete"
1. Like 2, but don't allow the user to change the defaults.

I think I prefer the last option (as a user, not implementor). This also makes the splitter more like the merger where things are fixed and it is easy to reason about the behavior.

padenot:
This fixes #975.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2853,6 +2853,14 @@ function setupRoutingGraph() {
                 </dl>
               </dd>
               <dt>
+                <a>ChannelSplitterNode</a>
+              </dt>
+              <dd>
+                The channel count cannot be changed, and an <span class=
+                "synchronous">InvalidStateError exception MUST be thrown for
+                any attempt to change the value.</span>
+              </dd>
+              <dt>
                 <a>ChannelMergerNode</a>
               </dt>
               <dd>
@@ -2928,6 +2936,14 @@ function setupRoutingGraph() {
                 value.</span>
               </dd>
               <dt>
+                <a>ChannelSplitterNode</a>
+              </dt>
+              <dd>
+                The channel count mode cannot be changed from "explicit" and an
+                <span class="synchronous">InvalidStateError exception must be
+                thrown for any attempt to change the value.</span>
+              </dd>
+              <dt>
                 <a>ChannelMergerNode</a>
               </dt>
               <dd>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2851,6 +2851,14 @@ <h2 id="OfflineAudioCompletionEvent">
                 </dl>
               </dd>
               <dt>
+                <a>ChannelSplitterNode</a>
+              </dt>
+              <dd>
+                The channel count cannot be changed, and an <span class=
+                "synchronous">InvalidStateError exception MUST be thrown for
```
rtoy: Add markup for InvalidStateError.

index.html
```diff
@@ -2926,6 +2934,14 @@ <h2 id="OfflineAudioCompletionEvent">
                 value.</span>
               </dd>
               <dt>
+                <a>ChannelSplitterNode</a>
+              </dt>
+              <dd>
+                The channel count mode cannot be changed from "explicit" and an
+                <span class="synchronous">InvalidStateError exception must be
```
rtoy: Add markup for InvalidStateError

index.html
```diff
@@ -7537,8 +7558,8 @@ <h3 id="Convolution-channel-configurations">
     numberOfInputs  : 1
     numberOfOutputs : Variable N (defaults to 6) // number of "active" (non-silent) outputs is determined by number of channels in the input
 
-    channelCountMode = "max";
-    channelInterpretation = "speakers";
+    channelCountMode = "explicite";
```
rtoy: s/explicite/explicit/

PR #1028, opened 2016-10-05
Add `alt` attribute in the new img tags.

padenot:


padenot:
(merging this without review, it's trivial enough).

PR #1029, opened 2016-10-05
Use Mixin pattern for contexts

hoch:
PTAL. Two simple changes.
- Moved `suspend` from BAC to AC.
- Added `EventTarget` and `implements` to AC and OAC.

Having the `onstatechange` event handler while BaseAudioContext does not extend EventTarget is still weird. I believe we might have to move this to each context.

Preview: https://rawgit.com/hoch/web-audio-api/fixing-context-idl/index.html

### DIFF
--- a/index.html
+++ b/index.html
@@ -784,8 +784,8 @@ function setupRoutingGraph() {
             output latency. Lowest power consumption.
           </dd>
         </dl>
-        <dl title="interface BaseAudioContext" class="idl"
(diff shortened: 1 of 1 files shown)


hoch:
This proposal has an issue: https://github.com/WebAudio/web-audio-api/issues/1022#issuecomment-251756006

rtoy:
From the conclusion of #1022, we're not going to use a mixin.

Closing this PR.

PR #1035, opened 2016-10-06
Small typo fix in `rolloffFactor` description

alvov:
Changed `panningModel` to `distanceModel` in `rolloffFactor` attribute description where `linear`, `inverse` and `exponential` models are mentioned. Also added **s** to `source move`, sorry, if I'm wrong here.

### DIFF
--- a/index.html
+++ b/index.html
@@ -6364,7 +6364,7 @@ var node = context.createScriptProcessor(bufferSize, numberOfInputChannels,
           </dt>
           <dd>
             <p>
-              A reference distance for reducing volume as source move further
+              A reference distance for reducing volume as source moves further
               from the listener. The default value is 1.
             </p>
           </dd>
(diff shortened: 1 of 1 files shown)


rtoy:
Thanks for the fix!  If possible, can you run tidy or apply the tidy fix?

alvov:
Sure, I will update the pr after a while.

alvov:
@rtoy done! Pardon me for not doing `make check` before creating this pr. What could possibly go wrong, I naively thought.

rtoy:
No worries!  I forget to run tidy all the time.

rtoy:
Hmm.  Your version of tidy unindented many `<pre>` tags. Not wrong, but not really needed.

The checks passed, so I think this is good.

alvov:
> Your version of tidy unindented many \<pre\> tags

Sorry again, my version is 5.0.0, it actually did show some errors with `</pre>`, and most tags were unindented, so I thought that's the right way :)

Thanks a lot!

PR #1036, opened 2016-10-06
Fix #1030: Define value for ChannelSplitterNode.channelCount

### LINKED ISSUE #1030: Default ChannelSplitterNode.channelCount not specified
Reviewer of #1027 overlooked the fact that we didn't actually specify the `channelCount` value.  It's supposed to be fixed to the `numberOfOutputs`.

rtoy:
The text never said that channelCount should be set to be the same as
the number of outputs.  Make it so, and add a note that there are
channelCount and channelCountMode constraints.

### DIFF
--- a/index.html
+++ b/index.html
@@ -7593,8 +7593,9 @@ function calculateNormalizationScale(buffer)
         </p>
         <p>
           The <code>channelCount</code> is set equal to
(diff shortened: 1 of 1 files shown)


padenot:
LGTM after tidy.

PR #1038, opened 2016-10-07
reject a TypeError from decodeAudioData

### LINKED ISSUE #1037: decodedAudioData() should probably reject a TypeError
Hi, I just found something which I think might be a little mistake. The current version of the spec says the decodeAudioData() promise might be rejected with a "DOMException whose name is "TypeError"". Everywhere else in the spec TypeErrors are regular TypeErrors which are not somehow wrapped in a DOMException, that's why I guess this might be a mistake. As far as I remember it was once a NotSupportError DOMException, which might be the reason for that.

If my assumptions are right and you agree, I'm happy to turn this into a pull request.

chrisguttandin:
This pull request is meant to fix #1037.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1209,8 +1209,7 @@ function setupRoutingGraph() {
               </li>
               <li>Else, execute the following steps:
                 <ol>
-                  <li>Let <var>error</var> be a <code>DOMException</code> whose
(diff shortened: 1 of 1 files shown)


padenot:
Thanks !

PR #1051, opened 2016-10-18
Fix #713: Specify clamping of automation times.

### LINKED ISSUE #713: Behavior of events scheduled in the past?
The spec doesn't say much about this, but what is supposed to happen if you schedule an event in the past?  Say currentTime = 1 sec, and you scheduled something at time 0.9.  What should happen.

I'm inclined to say that behavior is undefined if the time parameter to the event is in the past.

rtoy:
Specify clamping of the automation time for all automation methods.
Some were missed in #849.


rtoy:
@padenot PTAL.

PR #1054, opened 2016-10-26
Apply changes and clarifications regarding event scheduling and interaction.

### LINKED ISSUE #573: Interaction of automation methods is unclear
The [spec entry for automation methods](https://webaudio.github.io/web-audio-api/#the-audioparam-interface) says:

> The following rules will apply when calling these methods:
>
>  * If one of these events is added at a time where there is already an event of the exact same type, then the new event will replace the old one.
>  * If one of these events is added at a time where there is already one or more events of a different type, then it will be placed in the list after them, but before events whose times are after the event. 

It's not really clear, but "time" above probably refers to the `time` parameter in each of the methods.

One odd behaviour in Chrome is that if you do the following (where `g` is a gain node)
```
var g = context.createGain();
g.gain.linearRampToValueAtTime(0, context.currentTime + 1);
```
Chrome doesn't actually do anything.  You need to add a `setValueAtTime(1,0)` to establish a previous event for the linear ramp to use.

It's not clear what happens if a different event type is added at the same time as an existing event.  Should this be allowed?  What does it really mean to have different events at the same time?

In Chrome, you can do this:
```
g.gain.setValueAtTime(1,0);
g.gain.linearRampToValueAtTime(0, 1);
g.gain.exponentialRampToValueAtTime(100, 1);
```
What appears to happen is that you get a linear ramp from time 0 to time 1. Then you get an instantaneous jump to 100.  If you reversed the order of the calls, you get something different: an exponential ramp up to 100 and then an jump to 0 at time 1. I find this odd.

Also, what happens if you do something like:
```
g.gain.setValueAtTime(1,context.currentTime);
g.gain.linearRampToValue(0, 1000 + context.currentTime);
// Start the sources, etc., connecting to this gain.
// Wait 10 sec
g.gain.exponentialRampToValue(10, 100 + context.currentTime);
```
What does the gain curve look like now?  Does the linear ramp continue towards 0, and then at time 100+currentTime, an exponential ramp starts?  Something else?

What should the following do?
```
g.gain.setValueAtTime(1, 1+ context.currentTime);
g.gain.linearRampToValueAtTime(0, 1000 + context.currentTime);
// Wait 100 sec
g.gain.setValueAtTime(100, -10 + context.currentTime)
```

I don't have any real suggestions on how to resolve these.  We could say that the these corner cases are undefined and leave it to the implementation.  But we should be clear on what happens for the most common cases.

joeberkovitz:
Fix #573

### DIFF
--- a/index.html
+++ b/index.html
@@ -3343,13 +3343,13 @@ function setupRoutingGraph() {
           The following rules will apply when calling these methods:
         </p>
         <ul>
-          <li>If one of these events is added at a time where there is already
-          an event of the exact same type, then the new event will replace the
-          old one.
+          <li>Automation event times are not quantized with respect to the
+          prevailing sample rate. Formulas for determining curves and ramps are
+          applied to the exact numerical times given when scheduling events.
           </li>
           <li>If one of these events is added at a time where there is already
-          one or more events of a different type, then it will be placed in the
-          list after them, but before events whose times are after the event.
+          one or more events, then it will be placed in the list after them,
+          but before events whose times are after the event.
           </li>
           <li>
             <span class="synchronous">If setValueCurveAtTime() is called for


joeberkovitz:
@rtoy @padenot PTAL

PR #1055, opened 2016-10-26
Propagate settable-once behavior of ABSN.buffer to Convolver, WaveShaper

### LINKED ISSUE #667: The ConvolverHandler.prototype.buffer setter doesn't handle null
https://webaudio.github.io/web-audio-api/#ConvolverNode

"A mono, stereo, or 4-channel AudioBuffer containing the (possibly multi-channel) impulse response used by the ConvolverNode. The AudioBuffer must have 1, 2, or 4 channels or a NotSupportedError exception must be thrown. This AudioBuffer must be of the same sample-rate as the AudioContext or a NotSupportedError exception must be thrown. At the time when this attribute is set, the buffer and the state of the normalize attribute will be used to configure the ConvolverNode with this impulse response having the given normalization. The initial value of this attribute is null."

This doesn't say what should happen if `buffer` is set to null. In Chrome it silently does nothing. Should it be possible to clear the buffer by setting it to null?

See also https://github.com/WebAudio/web-audio-api/issues/666

joeberkovitz:
Fixes #667

### DIFF
--- a/index.html
+++ b/index.html
@@ -1293,8 +1293,6 @@ function setupRoutingGraph() {
           <dd>
             <p>
               Creates an <a><code>AudioBufferSourceNode</code></a>.
-              <a>AudioBufferSourceNode</a> are created with an internal flag
-              <code>buffer set</code>, initially set to false.
             </p>
           </dd>
           <dt>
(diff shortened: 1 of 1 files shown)


joeberkovitz:
@rtoy @padenot PTAL

index.html
```diff
@@ -8874,6 +8902,24 @@ <h2 id="WaveShaperNode">
               attribute therefore have no effect: the attribute must be set
               again in order to change the curve.
             </p>
+            <p>
+              To set the <code>curve</code> attribute, execute these steps:
```
joeberkovitz: I agree with @rtoy. Anyone else want to weigh in? If not, I'll remove the WaveShaper piece.

rtoy: But the same could be said for the buffer in an `AudioBufferSourceNode`; the buffer is used "atomically" in some sense, with some odd behavior if the new buffer is much shorter than the previous.  The `ConvolverNode` also has this property of being used "atomically" in principle, but at least in Chrome, various parts of the response are used in multiple threads to speed up long convolutions.

So perhaps for consistency, make `WaveShaper` also have this constraint as well.

Issue #469, opened 2015-01-29
Audio nodes should expose their intrinsic latency.


cristiano-belloni:
Stackoverflow thread here: http://stackoverflow.com/questions/25807887/using-an-offline-context-in-the-web-audio-api-shift-the-signal-of-264-samples

Compression, panning and waveshaping add latency in the current Chrome and Firefox implementation. Filters and gains do not.

You can see the behaviour here: https://wav.hya.io/#/fx/node_compressor - If you add a wave and then press the "Apply" button, the Compressor effect (based uniquely on a CompressorNode) adds a shift, which is clearly visible if you zoom down your waveform. I can provide a plnkr example, if needed. If you use a Filter (example: lowpass - https://wav.hya.io/#/fx/lowpass_filter), this does not happen.

@cwilso says:

> The behavior of the nodes isn't strictly defined by the spec (yet), so it's not well-defined. I'd expect an HRTF panner to add some latency, though. Not positive about compression.

so I thought it was worth to post it here.

I believe the nodes should automatically handle the shift by spec, when in an offline context (while it's inevitable in a real-time context). Alternatively, there should be a way to calculate the amount of shift each node applies, and counteract it application-side.

rtoy:
On Thu, Jan 29, 2015 at 6:18 AM, Cristiano Belloni <notifications@github.com
> wrote:

> Stackoverflow thread here:
> http://stackoverflow.com/questions/25807887/using-an-offline-context-in-the-web-audio-api-shift-the-signal-of-264-samples
>
> Compression, panning and waveshaping add latency in the current Chrome and
> Firefox implementation. Filters and gains do not.
>
​I think waveshapers should only add latency when you oversample.
Otherwise, there shouldn't be any latency.

Filters also add latency​,typically pretty small.

I think compressors add latency.  @hongchan can say more about that for
Chrome's implementation.

The HRTF panner (now no longer the default) also adds a small latency
 of about 64 samples, I think. This is what Chrome does. I think it also
depends on the sample rate of the context.

> You can see the behaviour here: https://wav.hya.io/#/fx/node_compressor -
> If you add a wave and then press the "Apply" button, the Compressor effect
> (based uniquely on a CompressorNode) adds a shift, which is clearly visible
> if you zoom down your waveform. I can provide a plnkr example, if needed.
> If you use a Filter (example: lowpass -
> https://wav.hya.io/#/fx/lowpass_filter), this does not happen.
>
[...]

PR #1058, opened 2016-10-28
Note the existence of audio processing latency.

joeberkovitz:
DelayNode and WaveShaperNode with oversampling are covered.


joeberkovitz:
This addresses #469 but does not close it, since the request is for programmatic determination of latency; rather, it supplies some useful information for V1 of the API.

PR #1064, opened 2016-11-08
Fix #611: Clamp maxDistance and refDistance

### LINKED ISSUE #611: Distance models: maxDistance vs refDistance
What should happen if someone sets maxDistance < refDistance?

What should happen if someone sets refDistance = 0?  

Since these are all distances, presumably, we should throw an InvalidStateError if they are set to negative values.

rtoy:
For the linear distance model, make d'max = max(dref, dmax) and d'ref
= min(dref,dmax).  And specify the value of the model is 1-f if
d'ref=d'max. This makes the model well-defined for all values of dref
and dmax.

Note that this change allows a user to set dref > dmax, but the
behavior is well-defined in this case (by reversing the order). Not
sure if we want that or not.

Also, specify that setting refDistance and maxDistance to a
non-positive value throws a TypeError.

### DIFF
--- a/index.html
+++ b/index.html
@@ -6238,12 +6238,17 @@ var node = context.createScriptProcessor(bufferSize, numberOfInputChannels,
             </p>
             <pre class="nohighlight">
             $$
-              1 - f\frac{\max(\min(d, d_{max}), d_{ref}) - d_{ref}}{d_{max} - d_{ref}}
+              1 - f\frac{\max(\min(d, d'_{max}), d'_{ref}) - d'_{ref}}{d'_{max} - d'_{ref}}
             $$
             
 </pre>
             <p>
-              That is, \(d\) is clamped to the interval \([d_{ref},\,
+              where \(d'_{ref} = \min(d_{ref}, d_{max})\) and \(d'_{max} =
+              \max(d_{ref}, d_{max})\). In the case where \(d'_{ref} =
+              d'_{max}\), the value of the linear model is taken to be \(1-f\).
+            </p>
+            <p>
+              Note that \(d\) is clamped to the interval \([d_{ref},\,
               d_{max}]\).
             </p>
           </dd>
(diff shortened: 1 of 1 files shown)


rtoy:
PTAL. 

Note that the change allows dref > dmax, but the linear model is still well-defined in this case and also the case where dref = dmax (having the value 1-f).  Not sure we want that.

I also took the opportunity to say that maxDistance and refDistance must be strictly positive values.

index.html
```diff
@@ -6396,7 +6401,9 @@ <h2 id="AudioWorklet-Examples">
           <dd>
             <p>
               A reference distance for reducing volume as source moves further
-              from the listener. The default value is 1.
+              from the listener. The default value is 1. A
+              <code>TypeError</code> exception must be thrown if this is set to
+              a non-positive value.
```
padenot: This is not a type issue, this is a range issue.

index.html
```diff
@@ -6406,7 +6413,8 @@ <h2 id="AudioWorklet-Examples">
             <p>
               The maximum distance between source and listener, after which the
               volume will not be reduced any further. The default value is
-              10000.
+              10000. A <code>TypeError</code> exception must be thrown if this
```
padenot: Same.

index.html
rtoy: RangeError then?

padenot: Usually "IndexSizeError" for that kind of things, although it's weird. RangeError does not exist any more apparently.

rtoy: Really?  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RangeError still lists RangeError and https://heycam.github.io/webidl/#exceptiondef-rangeerror still lists it too.

padenot:
You're right, I got something mixed up here.

PR #1065, opened 2016-11-08
Fix #324: Handle degenerate azimuth/elevation calculation

### LINKED ISSUE #324: Degenerate azimuth/elevation calculation
In https://webaudio.github.io/web-audio-api/#Spatialization the given algorithm says that if the source and listener are at the same position, the azimuth and elevation are both set to zero.

This seems incorrect.  If the source is moving towards the listener with an azimuth of 90, I would expect the azimuth to be continuous when it reaches the listener. If you don't do this, the azimuth suddenly changes to 0, causing a change in the sound.

This effect can be heard at http://jsfiddle.net/UxwGP/8/.

rtoy:
In the current code, if the source to listener vector is zero, the
algorithm immediately returns azimuth = 0 and elevation = 0.  This
causes a discontinuity.

Instead, remove the early exit and let the source-listener vector
propagate through. This removes this discontinuity, but in general
there is still the discontinuity when the source and listener
positions are the same.  This is unavoidable.

Also add a short note on what normalize() does in general and when the
vector is the 0 vector. (Does nothing).

### DIFF
--- a/index.html
+++ b/index.html
@@ -10657,10 +10657,11 @@ Quad up-mix:
           Azimuth and Elevation
         </h3>
         <p>
-          The following algorithm must be used to calculate the <em>azimuth</em>
-          and <em>elevation</em> for the <a><code>PannerNode</code></a>. The
-          <code>normalize()</code> function makes the vector a unit vector,
-          except when the vector has zero length. In this case, nothing is done.
+          The following algorithm must be used to calculate the
+          <em>azimuth</em> and <em>elevation</em> for the
+          <a><code>PannerNode</code></a>. The <code>normalize()</code> function
+          makes the vector a unit vector, except when the vector has zero
+          length. In this case, nothing is done.
         </p>
         <pre>
   // Calculate the source-listener vector.
--- a/index.html
+++ b/index.html
@@ -10657,20 +10657,14 @@ Quad up-mix:
           Azimuth and Elevation
         </h3>
         <p>
-          The following algorithm must be used to calculate the
-          <em>azimuth</em> and <em>elevation</em> for the
-          <a><code>PannerNode</code></a>:
+          The following algorithm must be used to calculate the <em>azimuth</em>
+          and <em>elevation</em> for the <a><code>PannerNode</code></a>. The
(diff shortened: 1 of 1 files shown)


padenot:
@karlt, does that sound good ?

rtoy:
An alternative would be to keep the algorithm as is, but change the values when the source and listener are the same. I think if you follow through, azimuth = 90 (not 0), and elevation = 90, (not 0).

joeberkovitz:
@padenot @karlt checking to see if we can expedite this review...

joeberkovitz:
@padenot to ping @karlt and if no answer, @padenot will resolve.

joeberkovitz:
@padenot reminder - please resolve this if @karlt isn't going to be involved

padenot:
So:
- When the source to listener vector is zero, then the azimuth will be 0, and the normalized azimuth (in degrees) will be 90, so it's right in the front of the listener, and the gain is equal power, which makes sense. This is the case when the listener and source are the same point, so we should not pan.
- When it's not zero, but the projected source on the listener plane is zero, then the azimuth will be zero, and the normalized azimuth will be 90, so it's equal power, which makes sense. This is the case where the source and the listener are above or below each other. We should not pan.
- When going from left to right around the center, I hear or see no discontinuity.

What am I missing ?

rtoy:
Ok. I think I remember the issue now.  

If you go through the computations, but skip the early exit for the case when the source and listener are the same, and also ignore the fact that you can't normalize a zero vector (leave it unchanged). you'll find that you end up with `projectedSource` = [0,0,0] (using the notation from the algorithm), Then, the azimuth is basically `acos(0)` which is 90 deg.

`frontBack` will be 0 because `projectedSource` is the zero vector.  Thus, the rest of the steps leave the azimuth as 90.

But I'm no longer able to reason about any of these clearly anymore.  The spec is just too confusing now (and even before?).  See #1208. I don't know how to compute the equal-power panner output anymore. (I could look at what Chrome and Firefox do, but that's cheating.)

rtoy:
Irrelevant now that we decided that when the source and listener are at the same location, audio should be heard from both left and right, as Firefox does today.

PR #1069, opened 2016-11-11
Fix #1068: ConstantSourceNode output behavior

### LINKED ISSUE #1068: ConstantSourceNode behavior on start() / stop()
I think these points are not clear in spec. : 

* What value is output the ConstantSourceNode before start(), maybe zero?
* Is the ConstantSourceNode after stop() output return to zero or remain the last output?

the AudioScheduledSourceNode start()/stop() description use the word 'sound', but it
may not match to ConstantSourceNode.

rtoy:
Specify that all source nodes produce silence before the node has
started and produces silence after the node has stopped.

### DIFF
--- a/index.html
+++ b/index.html
@@ -4102,6 +4102,14 @@ param.setValueCurveAtTime(curve, t7, t8 - t7);
           <a>OscillatorNode</a>.
         </p>
         <p>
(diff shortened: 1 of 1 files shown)


joeberkovitz:
Looks good!

PR #1071, opened 2016-11-11
Fix #1070: ABSN.start arguments should all be optional

### LINKED ISSUE #1070: AudioBufferSourceNode.start must have optional args
Currently, https://webaudio.github.io/web-audio-api/#AudioBufferSourceNode says
```
void start (double when, double offset, optional double duration);
```

But, @joeberkovitz was right in his review.  They should all be optional.  If they're not optional, `AudioBufferSourceNode.start()` will throw an error that there are not enough arguments.

rtoy:
Previously, not all arguments were optional so ABSN.start() would
throw an error that not enough arguments were passed.  This is
incorrect; we want ABSN.start() to start, as it used to before
AudioScheduledSourceNode was added.

### DIFF
--- a/index.html
+++ b/index.html
@@ -4790,7 +4790,7 @@ param.setValueCurveAtTime(curve, t7, t8 - t7);
             </p>
             <dl class="parameters">
               <dt>
-                double when
+                optional double when = 0
               </dt>
               <dd>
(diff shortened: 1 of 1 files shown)


joeberkovitz:
LGTM

PR #1072, opened 2016-11-11
Fix #1047: Simplify AudioNode dictionary section

### LINKED ISSUE #1047: AudioNode dictionary section inconsistent
The section https://webaudio.github.io/web-audio-api/#the-audionode-interface has a full section on dictionaries: https://webaudio.github.io/web-audio-api/#dictionaries.  But there's only one, so we could probably simplify this a bit to match the dictionary section used for all other nodes.

(This is a left over from the initial design where there were two dictionaries for an AudioNode.)

rtoy:
Basically rename the section from "Dictionaries" to
"AudioNodeOptions" to match how other nodes handle their
dictionaries.

Also add a bit of text saying the members are optional, and that the
specific default values depend on the actual node.

### DIFF
--- a/index.html
+++ b/index.html
@@ -3036,8 +3036,13 @@ function setupRoutingGraph() {
         </dl>
         <section>
           <h2>
-            Dictionaries
+            AudioNodeOptions
           </h2>
+          <p>
+            This specifies the options that can be used in constructing all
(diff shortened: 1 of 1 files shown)


joeberkovitz:
LGTM

PR #1074, opened 2016-11-14
Fix #1067: Add arguments to event constructors

### LINKED ISSUE #1067: Event constructors are missing arguments
The constructors added in #566 don't follow the pattern of other event constructors.  I believe just saying `[Constructpr]`on `OfflineAudioCompletionEvent` means you can do `new OfflineAudioCompletionEvent()` but you can't do anything interesting with that because there's no good way to set the event properties (including the event type).  Instead you want to take the event type string and a dictionary (which inherits from the `EventInit` dictionary) so that developers can do `new OfflineAudioCompletionEvent("complete", {renderedBuffer: buf})`.

Eg. see the definition of [CustomEvent](https://dom.spec.whatwg.org/#customevent) and [MouseEvent](https://w3c.github.io/uievents/#mouseevent).

rtoy:
To be useful, the event constructors needs arguments.  Add them and
also define the needed dictionaries.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2270,8 +2270,8 @@ function setupRoutingGraph() {
             <a><code>OfflineAudioContext</code></a> for legacy reasons.
           </p>
           <dl title=
-          "[Constructor]interface OfflineAudioCompletionEvent : Event" class=
-          "idl">
+          "[Constructor(DOMString type, OfflineAudioCompletionEventInit eventInitDict)]interface OfflineAudioCompletionEvent : Event"
+          class="idl">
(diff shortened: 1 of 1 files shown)


rtoy:
@joeberkovitz PTAL.

joeberkovitz:
@rtoy LGTM

rtoy:
Thanks. I'd like @RByers to take a look at the proposed changes before landing.

RByers:
LGTM, thanks!

PR #1075, opened 2016-11-15
Fix #1073: Move AudioProcessingEvent section into ScriptProcessorNode

### LINKED ISSUE #1073: Inconsistency of OfflineAudioCompletionEvent and AudioProcessingEvent sections
Currently, [OfflineAudioCompletionEvent](https://webaudio.github.io/web-audio-api/#OfflineAudioCompletionEvent) is a subsection of [OfflineAudioContext](https://webaudio.github.io/web-audio-api/#OfflineAudioContext).  This makes sense; the event is only relevant to the offline audio context.

However, [AudioProcessingEvent](https://webaudio.github.io/web-audio-api/#the-audioprocessingevent-interface---deprecated) is in its own top-level section at the same level as the the node sections.  But this event is only relevant to [ScriptProcessorNode](https://webaudio.github.io/web-audio-api/#the-scriptprocessornode-interface---deprecated).  

I think the AudioProcessingEvent section should be a subsection of ScriptProcessorNode.

rtoy:
Move the section for AudioProcessingEvent into the ScriptProcessorNode
section since the AudioProcessingEvent is for ScriptProcessorNode.
This makes this section match the style for
OfflineAudioCompletionEvent and OfflineAudioContext.

### DIFF
--- a/index.html
+++ b/index.html
@@ -6090,7 +6090,6 @@ var node = context.createScriptProcessor(bufferSize, numberOfInputChannels,
             </p>
           </dd>
         </dl>
-      </section>
       <section class="informative">
         <h2>
           The AudioProcessingEvent Interface - DEPRECATED
@@ -6185,6 +6184,7 @@ var node = context.createScriptProcessor(bufferSize, numberOfInputChannels,
             </dd>
           </dl>
         </section>
+      </section>
       </section>
       <section>
         <h2>
--- a/index.html
+++ b/index.html
@@ -6090,101 +6090,101 @@ var node = context.createScriptProcessor(bufferSize, numberOfInputChannels,
             </p>
           </dd>
         </dl>
-      <section class="informative">
-        <h2>
-          The AudioProcessingEvent Interface - DEPRECATED
-        </h2>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -6090,100 +6090,100 @@ <h2 id="AudioWorklet-Examples">
...
-            </p>
-          </dd>
-        </dl>
-        <section>
-          <h3>
-            AudioProcessingEventInit
-          </h3>
-          <dl title="dictionary AudioProcessingEventInit : EventInit" class=
-          "idl">
+        <section class="informative">
```
joeberkovitz: Did you accidentally lose the preceding `</section>` tag?

rtoy: I think this is what we want: https://rawgit.com/WebAudio/web-audio-api/9c13639f811292d6570e1506c060a5bbe38d73e0/index.html#the-scriptprocessornode-interface---deprecated

I think this looks right, and tidy doesn't complain about mismatched tags.

joeberkovitz: agreed, it looks fine

PR #1077, opened 2016-11-16
Fix #984: Patching AudioWorkletNodeOptions

### LINKED ISSUE #984: Should AudioWorkletNodeOptions inherit from AudioNodeOptions?
Currently, https://webaudio.github.io/web-audio-api/#AudioWorkletNodeOptions describes the dictionary by adding three members, `channelCount`, `channelCountMode`, and `channelInterpretation`.  Other than the default values, these are identical to `AudioNodeOptions`.

Also, https://webaudio.github.io/web-audio-api/#AudioWorkletNode has slightly different introductory text. All other nodes have a `<pre>` block describing the `'numberOfInputs`, `numberOfOutputs`, `channelCount`, `channelCountMode`, and `channelInterpretation`.

Perhaps it's best to follow the existing structure of the other nodes.

hoch:


hoch:
PTAL. A fairly simple and small change.

index.html
```diff
@@ -5425,39 +5425,23 @@ <h2 id="AudioWorkletNodeOptions">
...
             </p>
-            <dl title="dictionary AudioWorkletNodeOptions" class="idl">
+            <dl title="dictionary AudioWorkletNodeOptions : AudioNodeOptions"
+            class="idl">
               <dt>
                 unsigned long numberOfInputs
               </dt>
               <dd>
-                Same to <a>numberOfInputs</a> in <a><code>AudioNode</code></a>.
+                Same to <a href=
```
rtoy: Although you didn't actually change the content of this text, I think it reads better if you said something like

This is used to initialize the value of AudioNode numberOfInputs attribute.

Same comment for numberOfOutputs.

PR #1088, opened 2016-11-18
Fix #1082: AudioNode context is a BaseAudioContext

### LINKED ISSUE #1082: AudioNode "context" attribute is type of BaseAudioContext
But, the below document is not edited from AudioContext yet.

https://webaudio.github.io/web-audio-api/#the-audionode-interface

```ruby
interface AudioNode : EventTarget {
  // ..
  readonly        attribute AudioContext          context;
  // ..                      ^
}                            |
                            BaseAudioContext
```

```
2.4.1 Attributes

context of type AudioContext, readonly
    The AudioContext which owns this AudioNode.
         ^       ^
         |       |
        BaseAudioContext
```

rtoy:
Fix typo.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2824,11 +2824,11 @@ function setupRoutingGraph() {
             </dl>
           </dd>
           <dt>
-            readonly attribute AudioContext context
+            readonly attribute BaseAudioContext context
           </dt>
           <dd>
             <p>
-              The <a><code>AudioContext</code></a> which owns this
+              The <a><code>BaseAudioContext</code></a> which owns this
               <a><code>AudioNode</code></a>.
             </p>
           </dd>


index.html
```diff
@@ -2824,11 +2824,11 @@ <h2 id="OfflineAudioCompletionEvent">
...
           </dd>
           <dt>
-            readonly attribute AudioContext context
+            readonly attribute BaseAudioContext context
           </dt>
           <dd>
             <p>
-              The <a><code>AudioContext</code></a> which owns this
+              The <a><code>BaseAudioContext</code></a> which owns this
               <a><code>AudioNode</code></a>.
```
hoch: I don't think we need `<code>` tags here.

rtoy: I think we do this almost everywhere.  If this is the preferred style, we need to examine the entire spec to make it consistent.  So, I'd prefer to do that separately.

hoch: SGTM.

joeberkovitz:
Looks fine - I say merge!

PR #1090, opened 2016-11-28
Clarify behaviour of AudioParam.cancelScheduledValues()

SebastianZimmer:
On the W3C public-audio-dev mailing list, I noted that the spec is a bit vague on the behavior of AudioParam.cancelScheduledValues, especially concering currently running automation curves [1].

Since @rtoy agreed that this needs to be clarified in the spec [2], I created this PR as a suggestion.

I think, we should mention all automation methods here, not only `setTargetAtTime`. @joeberkovitz also suggested here [3], that possible discontinuities should be mentioned in the spec.

[1] https://lists.w3.org/Archives/Public/public-audio-dev/2016Nov/0001.html
[2] https://lists.w3.org/Archives/Public/public-audio-dev/2016Nov/0002.html
[3] https://github.com/WebAudio/web-audio-api/issues/344#issuecomment-151034935

### DIFF
--- a/index.html
+++ b/index.html
@@ -3851,13 +3851,15 @@ function setupRoutingGraph() {
           <dd>
             <p>
               Cancels all scheduled parameter changes with times greater than
-              or equal to <code>cancelTime</code>. Active
-              <code><a>setTargetAtTime</a></code> automations (those with
-              <code>cancelTime</code> less than the supplied time value) will
-              also be cancelled. Any hold values scheduled by <a href=
-              "#widl-AudioParam-cancelAndHoldAtTime-AudioParam-double-cancelTime">
-              <code>cancelAndHoldAtTime</code></a> are also removed if the hold
-              time occurs after <code>cancelTime</code>.
+              or equal to <code>cancelTime</code>. Cancelling a scheduled
+              parameter change means removing the scheduled event from the
+              event list. Any active automations whose event time is less than
+              <code>cancelTime</code> are also cancelled, and such
+              cancellations may cause discontinuities because the original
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -3852,9 +3852,14 @@ <h2 id="AudioParam">
...
               Cancels all scheduled parameter changes with times greater than
               or equal to <code>cancelTime</code>. Active
-              <code><a>setTargetAtTime</a></code> automations (those with
-              <code>cancelTime</code> less than the supplied time value) will
-              also be cancelled. Any hold values scheduled by <a href=
+              <code><a>linearRampToValueAtTime</a></code>,
+              <code><a>exponentialRampToValueAtTime</a></code>,
+              <code><a>setTargetAtTime</a></code> and
+              <code><a>setValueCurveAtTime</a></code> automations (those with
+              <code>cancelTime</code> less than the automation's end time) will
```
rtoy: I would simplify this to say something like "any automation scheduled to occur after `cancelTime`, even if the automation is currently running is removed.  This may cause discontinuities,..."

SebastianZimmer: Yes, this sounds better. I agree it is unnecessary to list every automation method. How about your suggestion with the addition "...scheduled to occur or end after `cancelTime`, even if..."

Because to me, currently running automations do not occur completely after `cancelTime`, but also before (partially).

Sorry for the nitpicking ;-)

SebastianZimmer:
@rtoy I've changed the commit as you suggested.
I'm not sure where and how to define event time, so I have to leave that up to you.

PR #1102, opened 2016-12-01
AudioWorklet: `maplike` and sendData/ondata

### LINKED ISSUE #990: Consider using maps for main->worklet inputs rather than creating individual properties on the AudioWorkletNode instance.
I've talked with some other folks on web platform, folks are pretty divided to whether we actually want to do this or not; we don't really have anything like this on the platform today, and it makes it difficult to expand the `AudioWorkletNode` later in time. 

As a concrete example, it would limit a future spec from adding a new property to `AudioWorkletNode`.

One option Instead is having two maps for both params & properties.

For example:

``` js
const ctx = ...;
const node = new AudioWorkletNode(ctx, 'foo');
node.parameters.get('freq').value = 5;

node.properties.set('buffer', buffer);
```

``` webidl
interface AudioParamMap {
  readonly maplike<DOMString, AudioParam>;
};
interface AudioPropertyMap {
  readonly maplike<DOMString, *>;
};

partial interface AudioWorkletNode {
  readonly attribute AudioParamMap parameters;
  readonly attribute AudioPropertyMap properties;
};
```

cc/ @domenic @esprehn @cwilso @hoch

hoch:
PTAL. This fixes #990.

Preview: https://rawgit.com/hoch/web-audio-api/a4bf075d8e5ee79e28ba75ad9dce8c0eeb42dc74/index.html#AudioWorklet

### DIFF
--- a/index.html
+++ b/index.html
@@ -5864,8 +5864,9 @@ class MyProcessor extends AudioWorkletProcessor {
                   false).
                 </li>
                 <li>Perform <a href=
-                "https://tc39.github.io/ecma262/#https://tc39.github.io/ecma262/#sec-set-o-p-v-throw">Set</a>(O=<var>parameterMap</var>,
-                P=<var>paramName</var>, V=<var>audioParam</var>).
+                "https://tc39.github.io/ecma262/#https://tc39.github.io/ecma262/#sec-set-o-p-v-throw">
+                  Set</a>(O=<var>parameterMap</var>, P=<var>paramName</var>,
+                  V=<var>audioParam</var>).
                 </li>
               </ol>
             </li>
--- a/index.html
+++ b/index.html
@@ -5458,17 +5458,28 @@ window.audioWorklet.import("bypass.js").then(function () {
               <p>
                 The <code>ondata</code> handler is called whenever the
                 associated <a>AudioWorkletProcessor</a> sends data back to the
-                AudioWorkletNode instance.
+                <a>AudioWorkletNode</a> instance.
               </p>
             </dd>
             <dt>
-              void sendData(any data)
+              void sendData()
             </dt>
             <dd>
               <p>
                 <code>sendData</code> may be called to send data to the
                 corresponding <a>AudioWorkletProcessor</a>.
               </p>
+              <dl class="parameters">
+                <dt>
+                  any data
+                </dt>
+                <dd>
+                  The object to deliver to the associated processor. This may
+                  be any value or JavaScript object handled by the <a href=
+                  "http://w3c.github.io/html/infrastructure.html#safe-passing-of-structured-data">
+                  structured clone</a> algorithm.
+                </dd>
+              </dl>
             </dd>
           </dl>
           <section>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -5850,11 +5864,16 @@ <h2 id="instantiation-of-AudioWorkletNode-and-AudioWorkletProcessor">
...
+                "https://tc39.github.io/ecma262/#https://tc39.github.io/ecma262/#sec-set-o-p-v-throw">
+                  Set</a>(O=<var>parameterMap</var>, P=<var>paramName</var>,
+                  V=<var>audioParam</var>).
                 </li>
               </ol>
             </li>
+            <li>Perform <a href=
+            "https://tc39.github.io/ecma262/#sec-createmethodproperty">CreateDataProperty</a>(O=<var>this</var>,
+            P="parameters", V=<var>parameterMap</var>).
+            </li>
```
joeberkovitz: Along the lines of some earlier comments by Elliott, an AWN's `parameters` attribute should be a read-only slot that gets the parameter map, rather than a read/write data property. We don't want someone instantiating an AWN and then setting `node.parameters` to some other non-maplike object.

hoch: See here: https://rawgit.com/hoch/web-audio-api/a4bf075d8e5ee79e28ba75ad9dce8c0eeb42dc74/index.html#AudioWorkletNode

The new IDL defines `AudioParamMap` interface which is a `readonly` maplike storage. The AudioWorkletNode also has this object as a `readonly` member.

Do we need anything else to protect it?

index.html
```diff
@@ -5517,29 +5533,23 @@ <h2 id="AudioWorkletProcessor">
...
+              attribute EventHandler ondata
             </dt>
             <dd>
               <p>
-                The <code>onmessage</code> handler is called whenever the
-                <a><code>AudioWorkletNode</code></a> posts a message back to
-                the corresponding processor object.
+                The <code>ondata</code> handler is called whenever the
+                <a><code>AudioWorkletNode</code></a> posts data to the
+                corresponding processor object.
```
hoch: I agree, but I think we should focus on the IDL/example change in this PR, then we can deal with the messaging part later.

hoch:
@joeberkovitz Any other feedback here?

joeberkovitz:
@hoch nope, lgtm!

hoch:
@rtoy I would like to have one more l-g-t-m before I merge this. PTAL.

index.html
```diff
@@ -5420,6 +5420,11 @@ <h2 id="AudioWorkletNode">
             the node to be retained in memory and perform audio processing in
             the absence of any connected inputs.
           </p>
+          <dl title="interface AudioParamMap" class="idl">
+            <dt>
+              readonly maplike&lt;DOMString, AudioParam&gt;
+            </dt>
```
rtoy: For some reason the closing `>` doesn't show up on my rawgit view.

index.html
```diff
@@ -5435,22 +5440,33 @@ <h2 id="AudioWorkletNode">
...
-              void postMessage(any message, optional sequence&lt;object&gt;
-              transfer)
+              attribute EventHandler ondata
             </dt>
             <dd>
               <p>
-                <code>postMessage</code> may be called to send a message to the
+                The <code>ondata</code> handler is called whenever the
+                associated <a>AudioWorkletProcessor</a> sends data back to the
+                AudioWorkletNode instance.
```
rtoy: Add `a` tags around AudioWorkletNode

index.html
```diff
@@ -5435,22 +5440,33 @@ <h2 id="AudioWorkletNode">
...
+              </p>
+            </dd>
+            <dt>
+              void sendData(any data)
```
rtoy: Perhaps add a description for the `data` parameter. Right now, there's none included.

index.html
hoch: That's weird. I can see it on both the remote preview and the local one. Could you please check again?

index.html
hoch: Done.

hoch:
@rtoy PTAL. I added parameter descriptions.

index.html
rtoy: Seems ok now.

PR #1130, opened 2017-01-03
Remove unnecessary AudioContext arg from AudioBuffer constructor.

### LINKED ISSUE #370: AudioBuffer should not be tied to a context
AudioBuffers shouldn't really be tied to an AudioContext*.  Therefore, AudioBuffers should have a constructor that works - i.e. "context.createBuffer(2, 88200, 44100);" should turn into "new AudioBuffer(2, 88200, 44100);"

*except when they're created by decodeAudioData - and even that shouldn't be "tied" to an AudioContext - i.e. I think they should be shareable - but the context currently resamples.  We should also make decodeAudioData have a flag that disables the resampling.

joeberkovitz:
Fix #370.

### DIFF
--- a/index.html
+++ b/index.html
@@ -4627,10 +4627,12 @@ param.setValueCurveAtTime(curve, t7, t8 - t7);
           </h2>
           <p>
             This specifies the options to use in constructing an
-            <a><code>AudioBuffer</code></a>. Only the <a href=
-            "#widl-AudioBufferOptions-length"><code>length</code></a> member is
-            required. A <code>NotFoundError</code> exception MUST be thrown if
-            any of the required members are not specified.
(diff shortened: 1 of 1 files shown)


bzbarsky:
@padenot Doesn't Gecko have this implemented with the old signature?

@joeberkovitz Were there implementation bugs filed when this spec change happened, by any chance?   I don't see one for Gecko offhand...

joeberkovitz:
I'm not aware of any implementation bugs filed; normally the WG members from the various browser vendors have been taking care of that. I think @padenot would know.

padenot:
Sorry I was on PTO when this got merged, I usually handle opening Gecko bugs.

PR #1140, opened 2017-01-19
Add diagram describing mono-input-to-matrixed-stereo case.

### LINKED ISSUE #1050: Convolver diagram needs mono input 4-channel response
Based on the convolver node settings of `channelCount` and `channelCountMode`, it is valid to have a mono source with a 4-channel response.  The mono source is upmixed to stereo and the normal true-stereo processing occurs.

There's no diagram for this.  Should there be?

joeberkovitz:
Fixes #1050


joeberkovitz:
@rtoy PTAL. My version of Graffle doesn't export to SVG

joeberkovitz:
@rtoy I would rather have used SVG, but don't consider it a problem and we have PNGs still in use in some places in the doc -- so I will merge.

There's about a 10x size difference in terms of byte count  but we're still talking a modestly sized file.

Issue #1117, opened 2016-12-19
Better description of what AudioNode Constructors should do


bakulf:
AudioNodes have two ways to be constructed: Constructors and Factory methods.
We should definitely describe better what the Constructors should do because sometimes, this is unclear.
If these Constructors are supposed to call factory methods, I suggest to see how each param should be passed.

bzbarsky:
I suspect the sane way to define this would be to:

1.  Define the constructor behavior in detail, probably by moving over the current definition of the factor method.
2.  Define that the factory method calls the constructor.

That's assuming the arguments to the two are meant to be interpreted the same way, of course.

padenot:
Since this is the same for each constructor, we specced it directly in the base class (`AudioNode`): https://webaudio.github.io/web-audio-api/#dfn-factory-method.

We did it the opposite way, the constructor calls the factory method, speccing how different things (attributes, `AudioParam`s, etc.) should be set and a special case for the `ConvolverNode`.

bzbarsky:
> Since this is the same for each constructor, we specced it directly in the base class (AudioNode): https://webaudio.github.io/web-audio-api/#dfn-factory-method.

That's totally not discoverable from reading the actual definition of an AudioNode subclass constructor.  At the very least, those should link to the definition here.  However....

> We did it the opposite way, the constructor calls the factory method

There are several problems with doing it this way:

1.  The way you have it written, nothing actually uses the AudioParams you are setting up in https://webaudio.github.io/web-audio-api/#dfn-factory-method step 2.3.1, as far as I can tell.

2.  The attributes being set in https://webaudio.github.io/web-audio-api/#dfn-factory-method step 2.3.2 are being set on an object that hasn't been created yet; it will be created in step 2.4.  This is pretty confusing.

3.  At first glance the algorithm can end up passing null where it's not allowed to be passed to the factory method.  Maybe all the options dictionaries enforce that the non-nullable arguments of the factory method are required dictionary properties, but that's pretty fragile.

4.  The result violates the Web IDL spec.  Per IDL spec, a constructor call produces an object associated with the global of the constructor.  But your constructor definition creates the objects via factory method on the audiocontext, which might be from a different global.

Consider this testcase:

     <iframe></iframe>
     <script>
       var ctx = new AudioContext();
       var node = new frames[0].WaveShaperNode(ctx);
       alert(Object.getPrototypeOf(node) == frames[0].WaveShaperNode.prototype)
     </script>

[...]

PR #1144, opened 2017-02-08
Define the constructor for AudioNodes (#1117)

padenot:
The first commit is the most important of this PR, it defines what happens when constructing AudioNodes (either via the factory method or the ctor). The second fixes the descriptions of the various factory method to link back to the algorithm used (this is mostly a mechanical change). The third patch fixes a mistake in the first patch, and the fourth one defines the constructors in terms of all the above, linking back to the definitions appropriately: it's mostly mechanical, but adjustments have been performed for the AudioWorklet.

### DIFF
--- a/index.html
+++ b/index.html
@@ -4427,6 +4427,31 @@ param.setValueCurveAtTime(curve, t7, t8 - t7);
         "[Constructor(&lt;a&gt;BaseAudioContext&lt;/a&gt; context, optional &lt;a&gt;DelayOptions&lt;/a&gt; options)]interface DelayNode : &lt;a&gt;AudioNode&lt;/a&gt;"
         class="idl">
           <dt>
+            Constructor
+          </dt>
+          <dd>
+          <p>
+          Let <var>node</var> be a new <a>DelayNode</a> object.
+          <a href=#audionode-constructor-common>Initialize the common part</a>
+          of <var>node</var>, and return <var>node</var>.
+          </p>
+          <dl class=parameters>
+            <dt>
+              BaseAudioContext context
+            </dt>
+            <dd>
+              The <a>BaseAudioContext</a> this new <a>DelayNode</a> will be
+              <a href="#associated">associated</a> with.
+            </dd>
+            <dt>
+              optional DelayOptions options
+            </dt>
+            <dd>
+              Optional initial parameter value for this <a>DelayNode</a>.
+            </dd>
+          </dl>
+          </dd>
+          <dt>
             readonly attribute AudioParam delayTime
           </dt>
           <dd>
@@ -4791,6 +4816,31 @@ param.setValueCurveAtTime(curve, t7, t8 - t7);
         "[Constructor(&lt;a&gt;BaseAudioContext&lt;/a&gt; context, optional &lt;a&gt;AudioBufferSourceOptions&lt;/a&gt; options)]interface AudioBufferSourceNode : &lt;a&gt;AudioScheduledSourceNode&lt;/a&gt;"
         class="idl">
           <dt>
+            Constructor
+          </dt>
+          <dd>
+          <p>
+          Let <var>node</var> be a new <a>AudioBufferSourceNode</a> object.
+          <a href=#audionode-constructor-common>Initialize the common part</a>
+          of <var>node</var>, and return <var>node</var>.
+          </p>
+          <dl class=parameters>
+            <dt>
+              BaseAudioContext context
+            </dt>
+            <dd>
+              The <a>BaseAudioContext</a> this new <a>AudioBufferSourceNode</a> will be
+              <a href="#associated">associated</a> with.
+            </dd>
+            <dt>
+              optional AudioBufferSourceOptions options
+            </dt>
+            <dd>
+              Optional initial parameter value for this <a>AudioBufferSourceNode</a>.
+            </dd>
+          </dl>
+          </dd>
+          <dt>
             attribute AudioBuffer? buffer
           </dt>
           <dd>
@@ -5113,6 +5163,31 @@ if ((loopStart || loopEnd) &amp;& loopStart &gt;= 0 &amp;& loopEnd &gt; 0 &amp;&
         "[Constructor(&lt;a&gt;BaseAudioContext&lt;/a&gt; context, optional &lt;a&gt;ConstantSourceOptions&lt;/a&gt; options)]interface ConstantSourceNode : &lt;a&gt;AudioScheduledSourceNode&lt;/a&gt;"
         class="idl">
           <dt>
+            Constructor
+          </dt>
+          <dd>
+          <p>
+          Let <var>node</var> be a new <a>ConstantSourceNode</a> object.
+          <a href=#audionode-constructor-common>Initialize the common part</a>
+          of <var>node</var>, and return <var>node</var>.
+          </p>
+          <dl class=parameters>
+            <dt>
+              BaseAudioContext context
+            </dt>
+            <dd>
+              The <a>BaseAudioContext</a> this new <a>ConstantSourceNode</a> will be
+              <a href="#associated">associated</a> with.
+            </dd>
+            <dt>
+              optional ConstantSourceOptions options
+            </dt>
+            <dd>
+              Optional initial parameter value for this <a>ConstantSourceNode</a>.
+            </dd>
+          </dl>
+          </dd>
+          <dt>
             readonly attribute AudioParam offset
           </dt>
           <dd>
@@ -5169,7 +5244,33 @@ if ((loopStart || loopEnd) &amp;& loopStart &gt;= 0 &amp;& loopEnd &gt; 0 &amp;&
         </p>
         <dl title=
         "[Constructor(&lt;a&gt;AudioContext&lt;/a&gt; context, &lt;a&gt;MediaElementAudioSourceOptions&lt;/a&gt; options)]interface MediaElementAudioSourceNode : &lt;a&gt;AudioNode&lt;/a&gt;"
-        class="idl"></dl>
+        class="idl">
+          <dt>
+            Constructor
+          </dt>
+          <dd>
+          <p>
+          Let <var>node</var> be a new <a>MediaElementAudioSOurceNode</a> object.
+          <a href=#audionode-constructor-common>Initialize the common part</a>
+          of <var>node</var>, and return <var>node</var>.
+          </p>
+          <dl class=parameters>
+            <dt>
+              BaseAudioContext context
+            </dt>
+            <dd>
+              The <a>BaseAudioContext</a> this new <a>MediaElementAudioSourceNode</a> will be
+              <a href="#associated">associated</a> with.
+            </dd>
+            <dt>
+              optional MediaElementAudioSourceOptions options
+            </dt>
+            <dd>
+              Optional initial parameter value for this <a>MediaELementAudioSourceNode</a>.
+            </dd>
+          </dl>
+          </dd>
+      </dl>
         <p>
           A <a>MediaElementAudioSourceNode</a> is created given an
           <code>HTMLMediaElement</code> using the <a>AudioContext</a>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2360,68 +2363,114 @@ <h2 id="OfflineAudioCompletionEvent">
...
-              </li>
-              <li>Call the corresponding <a>factory method</a> with the
-              parameters in <var>parameters</var>.
-              </li>
-            </ol>
+          <li>For each member of <var>dict</var> passed
+          in, execute these steps, with <var>k</var> the key of the member, and
+          <var>v</var> its value:
+          <ol>
+            <li>If<var>k</var> is <code>disableNormalization</code> or
```
rtoy: s/If/If /

index.html
```diff
@@ -4265,9 +4314,32 @@ <h2 id="GainNode">
...
+            </dt>
+            <dd>
+              The <a>BaseAudioContext</a> this new <a>GainNode</a> will be
+              <a href="#associated">associated</a> with.
+            </dd>
+            <dt>
+              optional GainOptions options
+            </dt>
+            <dd>
+              An optional initial gain value for this <a>GainNode</a>.
```
rtoy: |options| isn't the actual initial gain value. Not sure how best to describe this.  Maybe say nothing at all since there's a link to the description of the actual GainOptions dict?  Same goes for all of the following FooOptions dictionaries.

index.html
```diff
@@ -2360,69 +2367,124 @@ <h2 id="OfflineAudioCompletionEvent">
...
+            <a href="#audionode-constructor-common">Initialize the common
+            part</a> of <var>o</var> with <var>c</var> and <var>option</var> as
+            arguments.
+          </li>
+          <li>Return <var>o</var>
+          </li>
+        </ol>
+        <p>
+          <dfn id="audionode-constructor-common">Initializing the common
+          part</dfn> of an object <var>o</var> of interface <var>i</var> that
```
rtoy: This talks about initializing the common parts like the number of inputs/outputs and channel parameters, and it does.  But then it also talks about the non-common parts like the node-specific options.

I think this needs to be split up into two pieces or just renamed to so that's not about initializing the common part.

index.html
```diff
@@ -8950,6 +9369,31 @@ <h3 id="example-2">
...
+            <dl class="parameters">
+              <dt>
+                BaseAudioContext context
+              </dt>
+              <dd>
+                The <a>BaseAudioContext</a> this new <a>IIRFilterNode</a> will
+                be <a href="#associated">associated</a> with.
+              </dd>
+              <dt>
+                optional IIRFilterOptions options
```
rtoy: This isn't optional.

index.html
```diff
@@ -9744,7 +10240,35 @@ <h2 id="MediaStreamAudioSourceNode">
...
+              <dt>
+                BaseAudioContext context
+              </dt>
+              <dd>
+                The <a>BaseAudioContext</a> this new
+                <a>MediaStreamAudioSourceNode</a> will be <a href=
+                "#associated">associated</a> with.
+              </dd>
+              <dt>
+                optional MediaStreamAudioSourceOptions options
```
rtoy: This isn't optional.

index.html
padenot: I've renamed it, you're right.

index.html
```diff
@@ -2360,69 +2367,124 @@ <h2 id="OfflineAudioCompletionEvent">
...
+        <p>
+          To create a new <a>AudioNode</a> of a particular type <var>i</var>
+          using its <a>factory method</a>, called on a
+          <var>BaseAudioContext</var> <var>c</var>, execute these steps:
+        </p>
+        <ol>
+          <li>Let <var>o</var> be a new object of type <var>i</var>.
+          </li>
+          <li>Let <var>option</var> be a dictionnary of the type associated to
+          the interface associated to this factory method.
```
padenot: Done. I also linked more things.

index.html
```diff
@@ -4269,6 +4331,31 @@ <h2 id="GainNode">
         "[Constructor(&lt;a&gt;BaseAudioContext&lt;/a&gt; context, optional &lt;a&gt;GainOptions&lt;/a&gt; options)]interface GainNode : &lt;a&gt;AudioNode&lt;/a&gt;"
         class="idl">
```
padenot: apparently when you define it in the `<dl>`, then you don't have to do it in the title. I'll fix.

index.html
```diff
@@ -4269,6 +4331,31 @@ <h2 id="GainNode">
...
           <dt>
+            Constructor
+          </dt>
+          <dd>
+            <p>
+              Let <var>gain</var> be a new <a>GainNode</a> object. <a href=
+              "#audionode-constructor-common">Initialize the common part</a> of
+              <var>gain</var>, and return <var>gain</var>.
```
padenot: This is not called "initializing an AudioNode", hopefully less confusing.

index.html
```diff
@@ -2360,69 +2367,124 @@ <h2 id="OfflineAudioCompletionEvent">
...
+          </li>
+          <li>If the <a>AudioNode</a> being constructed is a
+          <a>ConvolverNode</a>, set its <a href=
+          "#widl-ConvolverNode-normalize">normalize</a> attribute with the
+          inverse of the value of the <a href=
+          "#widl-ConvolverOptions-disableNormalization">disableNormalization</a>
+          in <var>dict</var>, and then set its <a href=
+          "#widl-ConvolverNode-buffer">buffer</a> attribute to the value of the
+          <a href="#widl-ConvolverOptions-buffer">buffer</a> in <var>dict</var>
+          member, in this order, and jump to the last step of this algorithm.
```
padenot: Yes, this algorithm describes a way of getting the desired result. There are other ways to get the same result.

index.html
```diff
@@ -2360,69 +2367,124 @@ <h2 id="OfflineAudioCompletionEvent">
...
+          <li>For each member of <var>dict</var> passed in, execute these
+          steps, with <var>k</var> the key of the member, and <var>v</var> its
+          value:
             <ol>
-              <li>Let <var>parameters</var> be an array that has a size equal
-              to the number of parameters of the associated <a>factory
-              method</a> function for this <a>AudioNode</a> type. Initialize
-              all elements of this array to <code>null</code>.
+              <li>If <var>k</var> is <code>disableNormalization</code> or
+              <code>buffer</code>, jump to the beginning of this loop.
```
padenot: Fixed.

index.html
padenot: Fixed

index.html
padenot: Fixed

padenot:
@rtoy, sorry for the delays, I've addressed the issues you found, preview [here](https://rawgit.com/padenot/web-audio-api/a338555737720fc5002ec137a3b7ab8669a33db7/index.html).

index.html
```diff
@@ -2363,68 +2370,131 @@ <h2 id="OfflineAudioCompletionEvent">
...
-              <li>Call the corresponding <a>factory method</a> with the
-              parameters in <var>parameters</var>.
-              </li>
-            </ol>
-          </li>
-        </ol>
+        </p>
+        <section>
+          <h3>
+            AudioNode creation
```
rtoy: Nit: "creation" should be capitalized.

index.html
```diff
@@ -2363,68 +2370,131 @@ <h2 id="OfflineAudioCompletionEvent">
...
+            <dfn>factory method</dfn> on the <a>BaseAudioContext</a> or
+            <a>AudioContext</a>.
+          </p>
+          <p>
+            The <a>BaseAudioContext</a> passed as first argument of the
+            constructor of an <a>AudioNode</a>s is called the <dfn id=
+            "associated">associated <a>BaseAudioContext</a></dfn> of the
+            <a>AudioNode</a> to be created. Similarly, when using the factory
+            method, the <a>associated <code>BaseAudioContext</code></a> of the
+            <a>AudioNode</a> is the first argument of the factory method.
```
rtoy: This seems wrong since many factory methods don't have a first argument at all.  Or perhaps I don't understand what you're trying to say here.

index.html
```diff
@@ -2363,68 +2370,131 @@ <h2 id="OfflineAudioCompletionEvent">
...
+          </p>
+          <p>
+            To create a new <a>AudioNode</a> of a particular type <var>i</var>
```
rtoy: Can we use "n" instead of "i"? I find it hard to distinguish between an "i" and a "l" and a "1".

index.html
```diff
@@ -2363,68 +2370,131 @@ <h2 id="OfflineAudioCompletionEvent">
...
+            <var>option</var> as second argument, from the <a href=
+            "https://html.spec.whatwg.org/multipage/webappapis.html#concept-relevant-global">
+            relevant global</a> of <var>c</var>, execute these steps:
+          </p>
+          <ol>
+            <li>Let <var>o</var> be a new object of type <var>i</var>.
+            </li>
+            <li>
+              <a href="#audionode-constructor-init">Initialize the common
+              part</a> of <var>o</var>, with <var>c</var> and <var>option</var>
```
rtoy: I assume you don't want to say "Initializing the common part" here anymore because that section isn't about just the common part anymore.

index.html
```diff
@@ -2363,68 +2370,131 @@ <h2 id="OfflineAudioCompletionEvent">
...
+          </ol>
+          <p>
+            To create a new <a>AudioNode</a> of a particular type <var>i</var>
+            using its <a>factory method</a>, called on a
+            <var>BaseAudioContext</var> <var>c</var>, execute these steps:
+          </p>
+          <ol>
+            <li>Let <var>o</var> be a new object of type <var>i</var>.
+            </li>
+            <li>Let <var>option</var> be a dictionnary of the type <a href=
```
rtoy: Typo: "dictionnary" -> "dictionary"

index.html
```diff
@@ -2363,68 +2370,131 @@ <h2 id="OfflineAudioCompletionEvent">
...
+            "#dfn-associated-option-object">associated</a> to the interface
+            <a href="#dfn-associated-interface">associated</a> to this factory
+            method.
+            </li>
+            <li>For each parameter passed to the factory method, set the
+            dictionary member of the same name on <var>option</var> to the
+            value of this parameter.
+            </li>
+            <li>
+              <a href="#audionode-constructor-init">Initialize the common
```
rtoy: Ditto. It's not the common part anymore.

index.html
```diff
@@ -4430,9 +4548,7 @@ <h2 id="AudioBuffer">
           <a><code>OfflineAudioContext</code></a> and an
           <a><code>AudioContext</code></a>.
         </p>
-        <dl title=
-        "[Constructor(&lt;a&gt;AudioBufferOptions&lt;a&gt; options)]interface AudioBuffer"
-        class="idl">
+        <dl title="interface AudioBuffer" class="idl">
```
rtoy: I think you forget to define the actual constructor for this.  And the constructor for this doesn't follow the AudioNode constructors because context is not the first arg.  Not sure where the best place to describe the constructor should be.

index.html
```diff
@@ -5096,8 +5265,36 @@ <h2 id="MediaElementAudioSourceNode">
...
+              <dt>
+                BaseAudioContext context
+              </dt>
+              <dd>
+                The <a>BaseAudioContext</a> this new
+                <a>MediaElementAudioSourceNode</a> will be <a href=
+                "#associated">associated</a> with.
+              </dd>
+              <dt>
+                optional MediaElementAudioSourceOptions options
```
rtoy: This is not optional.

index.html
```diff
@@ -5096,8 +5265,36 @@ <h2 id="MediaElementAudioSourceNode">
...
+              </dt>
+              <dd>
+                Optional initial parameter value for this
+                <a>MediaELementAudioSourceNode</a>.
+              </dd>
+            </dl>
+          </dd>
+        </dl>
```
rtoy: Something is messed up somewhere.  The Media* nodes all used to be in successive sections.  Now the worklet node comes after MediaElementAudioSource, and the other nodes are all placed at the end.

index.html
```diff
@@ -8950,10 +9366,34 @@ <h3 id="example-2">
...
+              </dt>
+              <dd>
+                The <a>BaseAudioContext</a> this new <a>IIRFilterNode</a> will
+                be <a href="#associated">associated</a> with.
+              </dd>
+              <dt>
+                IIRFilterOptions options
+              </dt>
+              <dd>
+                Optional initial parameter value for this <a>IIRFilterNode</a>.
```
rtoy: Not optional.

index.html
padenot: Yeah this is wrong. I meant "The associated BaseAudioContext when calling a factory method is the BaseAudioContext this factory method is called on", which makes sense.

index.html
padenot: Done

index.html
padenot: Yeah I half assed my regex apparently. Fixed.

index.html
padenot: Those are not `AudioNode`s, the goal is to fix it in a different PR (this PR and issue is about AudioNodes, there are other issues about other objects).

This is different because the notion of "associated AudioContext" does not apply, among other things. I've reinstated this constructor.

Anyways, I plan on following up immediately with defining the ctors for other objects, this PR is big enough already.

index.html
padenot: Fixed.

index.html
padenot: This PR does not change this, but I can do a patch to reorder the section if you'd like.

rtoy:
For those following along:  https://rawgit.com/WebAudio/web-audio-api/591f0a7/index.html

index.html
rtoy: sgtm.

index.html
rtoy: If you haven't already, don't bother. Recall that we're going to reorder all the sections anyway to roughly alphabetical.

PR #1145, opened 2017-02-09
Prevent using the tidy from Apple, it's from 2006 and does not understand HTML5

padenot:
Effectively, we pin harder to 5.2 (patch release are ok). This works for me on macOS Sierra.


hoch:
Non-editor LGTM.

Issue #1105, opened 2016-12-07
PannerNode.refDistance shouldn't throw when it's set to 0.


asRIA:
Hi,
Changes introduced in https://github.com/WebAudio/web-audio-api/pull/1064 
contradict with OpenAL specification: https://www.openal.org/documentation/openal-1.1-specification.pdf  
Where `AL_REFERENCE_DISTANCE` takes `[0, any]` ($Table 4-12)

I understand that proposed changes prevent to have strange result for exponential mode, but in case of linear mode it's safe to have `0` as a value. 

I brought OpenGL as a reference, because when a game is ported with Emscripten to ASM.js/WebAssembly, and it uses OpenAL then it will throw an exception. 

CC:
@rtoy 
@padenot

rtoy:
First, I would say WebAudio is not OpenAL, but it is certainly inspired by OpenAL.

It's probably possible to give definite meanings for all of these distance models, but does anyone actually use a ref distance of 0?

asRIA:
OpenAL and linear model works well for 2D games.
I don't know if any other company uses this combination, and definitely I can't say if somebody has published HTML5 game. 
Major concern refers to games that are already published. Our tech stack can adjust to new situation. 

In case of removing `setVelocity` from `PannerNode` it was slightly different. It was nicely documented and we have seen messages about deprecation. Removing possibility to set reference distance to 0 was sudden.

rtoy:
I'm not really opposed to reverting this change.  Just now need to come up with appropriate formulas for the corner cases for the inverse and exponential models.  And since you found this, it seems clear that someone sets refDistance to 0 as part of their application.

Added WG review label for wider review.

rtoy:
Update formulas to allow 0.

asRIA:
Thank you! Feel free to close it.

PR #1147, opened 2017-02-11
#1105 Allow zero for a reference distance

asRIA:
This a continuation for #1105. 
It allows to set 0 for a reference distance, what is useful in a linear mode.


rtoy:
Thanks for this patch.  Because it was missing a few items, I took the liberty of creating a new PR (#1150) based on these changes.

Closing this.  Your help was much appreciated.

PR #1150, opened 2017-02-15
Fix #1105: Allow refDistance to be 0 for PannerNode

### LINKED ISSUE #1105: PannerNode.refDistance shouldn't throw when it's set to 0.

rtoy:
If refDistance is 0, define the inverse and exponential models to take
the value 0 independent of the values of d and f.

The limit as d and dref go to zero aren't well defined and depends on
the direction of approach.  So I took the lazy way and set the value
to be always be 0.

### DIFF
--- a/index.html
+++ b/index.html
@@ -6494,7 +6494,8 @@ var node = context.createScriptProcessor(bufferSize, numberOfInputChannels,
 </pre>
             <p>
               That is, \(d\) is clamped to the interval \([d_{ref},\,
-              \infty)\).
+              \infty)\). If \(d_{ref} = 0\), the value of the inverse model is
+              taken to be 0, independent of the value of \(d\) and \(f\).
             </p>
           </dd>
           <dt>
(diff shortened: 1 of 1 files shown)


rtoy:
I suppose we could use `max(dref, floatMin)` where `floatMin` is the smallest non-zero single float value (1.4012985e-45).  This gets rid of that discontinuity, I think. This allows the user to set dref to 0, without actually behaving as if it were zero.  Don't know if that matters or not to the reporter in #1105.

joeberkovitz:
@rtoy just a reminder to please close #1147. thanks!

rtoy:
Internally clamping dref to max(dref, eps) works.  But if the user really, truly wanted 0, he'll be able to tell, but not sure if that matters.

Some graphs.  This first graph is the exponential case with f = 1 and eps = 0.01 (to make it easier to see the effect).

exponential-panner

This graph is for the inverse case with f = 0.5 and eps = 0.01:
inverse-panner

rtoy:
After a quick discussion on the teleconf, we don't need to have any epsilon. Zero is fine.

PR #1158, opened 2017-03-01
Fix #1004: Specify copyFrom/To channel behavior

### LINKED ISSUE #1004: AudioBuffer copyFrom/To channel ambiguity
The function descriptions seem rather cryptic here as to what actually happens, especially if I only want to copy a portion of the buffer. Say I have a buffer of 2 channels, each 2048 samples long. If I do the following:

``` javascript
    var myBuffer = context.createAudioBuffer(2, 2048, context.sampleRate);
    var destination = new Float32Array(1024);
    myBuffer.copyFromChannel(destination, 0);
```

what should the behaviour be? Should I get an error because my buffer is not big enough or will it only copy the first 1024 samples (as I think it should).
Equally if the last line were `myBuffer.copyFromChannel(destination, 0, 2000)`, meaning only the last 48 samples are captured, would there be an error, fill the rest of the destination array with zeros, or completely ignore them.

The same issues exist in the copyToChannel.

The third argument is also strange, `startInChannel` doesn't immediately sound like an `offset`, even though that is what it is.

From the spec:

> When reading data from an AudioBuffer's channels, and the data can be processed in chunks, copyFromChannel should be preferred to calling getChannelData

it would seem that copyFromChannel can be used to access the buffer on a frame-by-frame basis but when handled in odd circumstances there is room for difference in the implementation.

rtoy:
Basically specify that copying is done until we run out of frames in
the source or run out of space in the destination.  In either case,
the remaining elements are not modified, so only a part of the array
is modified.

### DIFF
--- a/index.html
+++ b/index.html
@@ -4496,11 +4496,11 @@ param.setValueCurveAtTime(curve, t7, t8 - t7);
               Let <code>buffer</code> be the <a>AudioBuffer</a> buffer with
               \(N_b\) frames, let \(N_f\) be the number of elements in the
               <code>destination</code> array, and \(k\) be the value of
-              <code>startInChannel</code>. Then the number of frames copied from
-              <code>buffer</code> to <code>destination</code> is \(\min(N_b - k,
-              N_f)\).  If this is less than \(N_f\), then the remaining elements
-              of <code>destination</code> are not modified.
-              </p>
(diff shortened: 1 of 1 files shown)


joeberkovitz:
@rtoy this looks good to me. Merge at will!

PR #1161, opened 2017-03-03
Clarify normative aspects of AudioNode lifetime.

### LINKED ISSUE #1079: Need a more algorithmic definition of AudioWorkletNode lifetime
The definition of lifetime used in the fix to #475 may lack specificity in some way, so WG members have asked to add an issue reflecting the potential need for an algorithmic definition.

joeberkovitz:
Fixes #1079

### DIFF
--- a/index.html
+++ b/index.html
@@ -3096,9 +3096,7 @@ function setupRoutingGraph() {
             conditions under which an AudioNode is alive, meaning that it MUST
             be retained in the graph by an implementation. Where these
             conditions do not apply, AudioNodes MAY be released by an
-            implementation. Additionally, implementations may apply other
-            methods to avoid unnecessary resource usage and unbounded memory
-            growth of unused/finished nodes.
+            implementation.
           </p>
           <p>
             An <a><code>AudioNode</code></a> must remain alive as long as there
--- a/index.html
+++ b/index.html
@@ -3093,14 +3093,13 @@ function setupRoutingGraph() {
           </h3>
           <p>
             The following behaviors provide a normative description of the
-            conditions under which an AudioNode is alive, meaning that it MUST
-            be retained in the graph by an implementation. Where these
-            conditions do not apply, AudioNodes MAY be released by an
+            conditions under which an <a>AudioNode</a> is alive, meaning that
+            it MUST be retained in the graph by an implementation. Where these
+            conditions do not apply, <a>AudioNode</a>s MAY be released by an
             implementation.
           </p>
           <p>
-            An <a><code>AudioNode</code></a> must remain alive as long as there
-            are any references to it. There are several types of references:
+            There are several types of references:
           </p>
           <ol>
             <li>A <em>normal</em> JavaScript reference obeying normal garbage
--- a/index.html
+++ b/index.html
@@ -3088,21 +3088,21 @@ function setupRoutingGraph() {
           </dl>
         </section>
         <section>
-          <h3 id="lifetime-AudioNode" class="informative">
+          <h3 id="lifetime-AudioNode">
             Lifetime
           </h3>
           <p>
-            <i>This section is informative.</i>
+            The following behaviors provide a normative description of the
+            conditions under which an AudioNode is alive, meaning that it MUST
+            be retained in the graph by an implementation. Where these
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -3088,21 +3088,21 @@ <h2 id="OfflineAudioCompletionEvent">
...
           </h3>
           <p>
-            <i>This section is informative.</i>
+            The following behaviors provide a normative description of the
+            conditions under which an AudioNode is alive, meaning that it MUST
+            be retained in the graph by an implementation. Where these
+            conditions do not apply, AudioNodes MAY be released by an
+            implementation. Additionally, implementations may apply other
+            methods to avoid unnecessary resource usage and unbounded memory
+            growth of unused/finished nodes.
```
rtoy: Have we defined what "finished" means?

joeberkovitz:
@rtoy in https://github.com/WebAudio/web-audio-api/issues/1079#issuecomment-277033023, I recorded the resolution that the WG reached on our Feb 2 call, which was that the issue could be resolved by making the lifetime text normative.

To recap that discussion, we concluded that this is OK because the description of "active reference" for an AWN is normative, and the lifetime section piggybacks on this definition much as it does on tail-time references and other already-existing criteria for node retention.

index.html
joeberkovitz: I guess not. What language would you like to see here, exactly?

rtoy: Dunno. Just leave out the last sentence?  Of the couple of languages that I know that  have GC, GC isn't actually required as part of the language. Maybe javascript is one such language so we don't have to say anything about GCing "unused/finished" node?

index.html
```diff
@@ -3088,21 +3088,19 @@ <h2 id="OfflineAudioCompletionEvent">
...
           <p>
-            An implementation may choose any method to avoid unnecessary
-            resource usage and unbounded memory growth of unused/finished
-            nodes. The following is a description to help guide the general
-            expectation of how node lifetime would be managed.
-          </p>
-          <p>
-            An <a><code>AudioNode</code></a> will live as long as there are any
-            references to it. There are several types of references:
+            An <a><code>AudioNode</code></a> must remain alive as long as there
```
rtoy: Actually, I think "alive" isn't well-defined here.  Not exactly sure what to say here because we can certainly create an ABSN, start it, and drop the JS reference to it.  The node must still play.

joeberkovitz: But isn't this dealt with by item 2 in the list of reference types just after this paragraph, which describes a _playing reference_ for ABSNs and other nodes that emit sound? That item has been around for a long time. It reads:

> A _playing reference_ for AudioBufferSourceNodes, MediaElementAudioSourceNodes, MediaStreamAudioSourceNodes and OscillatorNodes. These nodes maintain a _playing reference_ to themselves while they are currently playing.

joeberkovitz:
@padenot PTAL

index.html
```diff
@@ -3088,21 +3088,19 @@ <h2 id="OfflineAudioCompletionEvent">
...
         </section>
         <section>
-          <h3 id="lifetime-AudioNode" class="informative">
+          <h3 id="lifetime-AudioNode">
             Lifetime
           </h3>
           <p>
-            <i>This section is informative.</i>
+            The following behaviors provide a normative description of the
+            conditions under which an AudioNode is alive, meaning that it MUST
```
padenot: `<a>AudioNode</a>`

index.html
```diff
@@ -3088,21 +3088,19 @@ <h2 id="OfflineAudioCompletionEvent">
...
+            be retained in the graph by an implementation. Where these
+            conditions do not apply, AudioNodes MAY be released by an
```
padenot: `<a>AudioNode</a>`s. Maybe it's better to not mention what happens if this node does not have to be retained.

index.html
padenot: Is this a MUST in the sense of RFC2119, or a must in english ? In any case, this behaviour is well defined on the web platform already, no need to define it here again. We can simply say that we have self-references in some cases, and everything will work out fine.

joeberkovitz:
@padenot I've incorporated your suggested changes; please re-review

PR #1168, opened 2017-03-16
Fix #1152: Define OfflineAudioContext constructor with dictionary

### LINKED ISSUE #1152: OfflineAudioContext options dictionary?
Now that we have constructors for all of the nodes and they take an options dictionary and even `AudioContext` has an optional dictionary, wouldn't it be nice and consistent if `OfflineAudioContext` also supported a dictionary?

Then you could do
```
new OfflineAudioContext({numberOfChannels: 3, length: 99, sampleRate: 12345})
```
in addition to
```
new OfflineAudioContext(3, 99, 12345)
```

rtoy:
Add new constructor for OfflineAudioContext that takes a required
OfflineAudioContextOptions dictionary to specify the necessary
parameters to create an offline context.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2119,7 +2119,8 @@ function setupRoutingGraph() {
           </dd>
         </dl>
         <dl title=
-        "[Constructor(unsigned long numberOfChannels, unsigned long length, float sampleRate)] interface OfflineAudioContext : &lt;a&gt;BaseAudioContext&lt;/a&gt;"
+        "[Constructor(unsigned long numberOfChannels, unsigned long length, float sampleRate),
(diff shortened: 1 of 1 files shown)


rtoy:
Thanks @padenot 
But I've kind of reconsidered and think `OfflineAudioContextOptions` should inherit from `AudioNodeOptions`, as mentioned in https://github.com/WebAudio/web-audio-api/issues/1152#issuecomment-287385747

rtoy:
@padenot Are you working on better defining the existing constructor for OfflineAudioContext?  If so, perhaps it's best if I don't do any more work on this until that is done.  If not, then perhaps this can include all the needed bits.

rtoy:
Closing until @padenot finishes the other in-progress constructor changes.

PR #1176, opened 2017-03-23
ConvolverNode outputs mono when input and response are mono

### LINKED ISSUE #1159: Reduce complexity of Convolver
First, in https://webaudio.github.io/web-audio-api/#ConvolverNode, there's a typo where we say the mode is "clamped-max".  It should be "explicit" as mentioned in https://webaudio.github.io/web-audio-api/#dfn-channelcountmode-constraints and as updated in #942.

However, with this update, a mono input with a mono response produces a stereo output, as shown in https://webaudio.github.io/web-audio-api/#Convolution-channel-configurations.  This instantly doubles the complexity of all downstream nodes. This is especially important if the convolver is doing a simple FIR filter and a downstream node includes a convolver with a huge response filter to simulate the environment.

Granted the user can add a splitter to fix this, but this seems excessive.  A mono response with a mono response should produce a mono output.  This is breaking change, but I doubt (without evidence) that anyone was actually depending on that fact.  And I know that at least two people (me and someone on Slack) had problems with a mono source with a mono convolver followed by another mono convolver.  Before #942 was fixed, Chrome would produce silence because we had a stereo input with a mono response which wasn't supported.

Proposal:  Change `channelCountMode` back to "clamped-max", but otherwise leaving `channelCount` at 2, and also leaving them fixed.  By doing so, a mono input and mono response will result in a mono output.  The existing diagrams show what should happen in all other cases; the mono-mono diagram needs to be updated to show a mono output.

(This came up while trying to implement the current spec in Chrome.  To make it work and work efficiently so that unnecessary convolutions were not done, a few hacks needed to be introduced to get the actual number of inputs because the mixing rules implied the input was always mixed to stereo.  Making it "clamped-max" got rid of these hacks.)

joeberkovitz:
Fixes #1159


index.html
```diff
@@ -7544,15 +7543,16 @@ <h2 id="ConvolverNode">
               this attribute is null.
             </p>
             <p>
-              When the <code>buffer</code> has a single channel, the
-              <a>ConvolverNode</a> treats it as if it had two channels with
-              both channels containing exactly the same data. This is needed to
-              support a stereo input.
+              When the <code>buffer</code> has a single channel and there are
```
rtoy: Perhaps add a new paragraph (or sentences) saying that mono input and mono response produces a mono output and for all other cases the output is stereo.  (As a companion to the new diagram.)

joeberkovitz:
@rtoy I believe we should be good to go now; PTAL

PR #1177, opened 2017-03-24
Fix #300: Configurable sample rate for AudioContext

### LINKED ISSUE #300: Configurable sample rate for AudioContext
1. It would be nice to be able to run the AudioContext in a predefined sample rate, which might differ from the underlying hardware. Some Windows machines default to 192k, which require a lot of processing power compared to 44.1k, especially when 192k quality is not required. Besides increased performance, it can also help optimize buffers and give more consistent output quality. Example:

new AudioContext(optional float sampleRate = _default_)

Being able to change the value at runtime would mean increased flexibility (e.g. in case of sudden performance problems). If that'd be easy to implement in a consistent way:

interface AudioContext {
  attribute float sampleRate;
  ...
}
1. Being able to query the host for default or preferred sample rate(s) would also be nice, so we can choose an optimal sample rate when the underlying hardware supports it. Not sure if this would make sense on all OS:es though. E.g.

float AudioContext.preferredSampleRate

or

float[] AudioContext.preferredSampleRates

If there is some kind of Sink interface coming up which abstracts the AudioContext.destination then I guess it would make sense to go there:

List<Sink> MediaStreamTrack.getSinks()
sink.defaultSampleRate

I might have misunderstood the proposed sink api completely so forgive me if this does not make sense.

rtoy:
Add a new sampleRate attribute to AudioContextOptions to specify the
sample rate of the AudioContext.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2063,16 +2063,17 @@ function setupRoutingGraph() {
               <p>
                 Set the <a href=
                 "#widl-BaseAudioContext-sampleRate"><code>sampleRate</code></a>
-                of the context to this value. The supported values are the same
-                as the sample rates for an <a>AudioBuffer</a>. <span class=
-                "synchronous">A <code>NotSupportedError</code> exception MUST
-                be thrown if the specified sample rate is not supported.</span>
+                to this value for the <a>AudioContext</a> that will be created.
+                The supported values are the same as the sample rates for an
+                <a>AudioBuffer</a>. <span class="synchronous">A
+                <code>NotSupportedError</code> exception MUST be thrown if the
+                specified sample rate is not supported.</span>
               </p>
               <p>
                 If <a href=
                 "#widl-AudioContextOptions-sampleRate"><code>sampleRate</code></a>
-                is not specified, the sample rate of the audio hardware is
-                used.
+                is not specified, the preferred sample rate of the output
+                device for this <a>AudioContext</a> is used.
               </p>
             </dd>
           </dl>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2056,6 +2056,25 @@ <h2 id="AudioContext">
...
+                "#widl-BaseAudioContext-sampleRate"><code>sampleRate</code></a>
+                of the context to this value. The supported values are the same
+                as the sample rates for an <a>AudioBuffer</a>. <span class=
+                "synchronous">A <code>NotSupportedError</code> exception MUST
+                be thrown if the specified sample rate is not supported.</span>
+              </p>
+              <p>
+                If <a href=
+                "#widl-AudioContextOptions-sampleRate"><code>sampleRate</code></a>
+                is not specified, the sample rate of the audio hardware is
```
padenot: A bit weird to say hardware here. It's possible that it's the system mixer and not the hardware, maybe we should use a broader term ? We use the term "preferred sample-rate" in gecko, fwiw.

Looks good otherwise.

rtoy: Yes, "preferred sample-rate" is better.  Or maybe "preferred sample rate of the default output device"?

padenot: There is going to be some text about non-default output device at some point, let's not couple it too much to default devices, and state something like "of the output device for this `AudioContext`" so that it's fine for when we do write something about non-default devices.

rtoy: Done.

PR #1197, opened 2017-04-07
Fix #1195: createPeriodicWave takes `sequence<float>` arrays

### LINKED ISSUE #1195: Inconsistent types for createPeriodicWave and PeriodicWave constructor
The new `PeriodicWave` constructor takes a dictionary where the real and imaginary coefficients are of type `sequence<float>`.  Hence, non-finite values will produce a `TypeError`.

However, `createPeriodicWave` takes `Float32Array`s for the coefficients.  These are unrestricted.

To make these more consistent, I propose `createPeriodicWave` be changed to take `sequence<float>` types.  This is backward compatible with existing code and makes the behavior between the factor method and the constructor more consistent.

rtoy:
Instead of using `Float32Array`'s for the real and imaginary
coefficients, use `sequence<float>`.  This will cause errors to be
signaled if non-finite values are used, just like for the
`PeriodicWave` constructor.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1422,14 +1422,13 @@ function setupRoutingGraph() {
               representing a waveform containing arbitrary harmonic content.
               <span class="synchronous">The <code>real</code> and
               <code>imag</code> parameters must be of type
-              <code>Float32Array</code> (described in [[!TYPED-ARRAYS]]) of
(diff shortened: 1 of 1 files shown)


joeberkovitz:
LGTM

PR #1200, opened 2017-04-11
Reword treatment of starting time for AudioScheduledSourceNode

### LINKED ISSUE #915: The case for subsample-accurate start()

joeberkovitz:
Fixes #915

### DIFF
--- a/index.html
+++ b/index.html
@@ -4286,10 +4286,14 @@ param.setValueCurveAtTime(curve, t7, t8 - t7);
                 time coordinate system as the
                 <a><code>AudioContext</code></a>'s <a href=
                 "#widl-BaseAudioContext-currentTime">currentTime</a> attribute.
-                If 0 is passed in for this value or if the value is less than
-                <b>currentTime</b>, then the sound will start playing
-                immediately. <span class="synchronous">A TypeError exception
-                MUST be thrown if <code>when</code> is negative.</span>
+                When the signal emitted by the <a>AudioScheduledSourceNode</a>
+                depends on the sound's start time, the exact value of
+                <code>when</code> is always used without rounding to the
+                nearest sample frame. If 0 is passed in for this value or if
+                the value is less than <b>currentTime</b>, then the sound will
+                start playing immediately. <span class="synchronous">A
+                TypeError exception MUST be thrown if <code>when</code> is
+                negative.</span>
               </dd>
             </dl>
           </dd>
@@ -10131,8 +10135,9 @@ $$
   computedFrequency(t) = frequency(t) * pow(2, detune(t) / 1200)
 </pre>
         <p>
-          The OscillatorNode's instantaneous phase at each time is the time
-          integral of <em>computedFrequency</em>. Its <a>nominal range</a> is
+          The OscillatorNode's instantaneous phase at each time is the definite
+          time integral of <em>computedFrequency</em>, assuming a phase angle
+          of zero at the node's exact start time. Its <a>nominal range</a> is
           [-<a>Nyquist frequency</a>, <a>Nyquist frequency</a>].
         </p>
         <pre>


index.html
```diff
@@ -4286,10 +4286,14 @@ <h2 id="AudioScheduledSourceNode">
...
-                <b>currentTime</b>, then the sound will start playing
-                immediately. <span class="synchronous">A TypeError exception
-                MUST be thrown if <code>when</code> is negative.</span>
+                When the signal emitted by the <a>AudioScheduledSourceNode</a>
+                depends on the sound's start time, the exact value of
+                <code>when</code> is always used without rounding to the
+                nearest sample frame. If 0 is passed in for this value or if
+                the value is less than <b>currentTime</b>, then the sound will
+                start playing immediately. <span class="synchronous">A
+                TypeError exception MUST be thrown if <code>when</code> is
```
padenot: I think we're going to use RangeError for this kind of stuff. However, we haven't really decided yet.

padenot:
LGTM.

PR #1201, opened 2017-04-12
Fix #1157: Add attributes to  Media{Element,Stream}AudioSourceNode

### LINKED ISSUE #1157: MediaElementAudioSourceNode and MediaStreamAudioSourceNode attributes
Currently `MediaElementAudioSourceNode` and `MediaStreamAudioSourceNode` have no attributes.  However, Chrome actually defines readonly `mediaElement` and `mediaStream` attributes, respectively, for these nodes.  I don't think they're strictly necessary, but it certainly is nice to be able to introspect these nodes to be able to find what element or stream are associated with these nodes.

Should these be defined?

rtoy:
Add readonly attribute to each of these nodes to hold a reference to
the media element or stream used to create the node.

### DIFF
--- a/index.html
+++ b/index.html
@@ -5659,6 +5659,15 @@ function process(numberOfFrames) {
               </dd>
             </dl>
           </dd>
+          <dt>
+            [SameObject] readonly attribute HTMLMediaElement mediaElement
+          </dt>
+          <dd>
+            <p>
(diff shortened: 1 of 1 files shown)


rtoy:
Teleconf: reviewed and ready to merge.

PR #1204, opened 2017-04-13
Disallow empty string as AWP registration key.

### LINKED ISSUE #1172: An empty string for AudioWorkletProcessor name.
```
registerProcessor('', class { /* ... */ });
```

Should we allow this? An empty string is still a valid string and can be defined a valid key within a JS object. I prefer to throw an exception, but would love to hear WG's opinion.

joeberkovitz:
Fixes #1172.
Also corrects missing # in link to registerProcessor() definition.


joeberkovitz:
@rtoy @hoch PTAL

PR #1214, opened 2017-04-20
Add a link to web-platform-tests to the top of the spec

foolip:
A few other specs that have something similar:
https://fetch.spec.whatwg.org/
https://w3c.github.io/IndexedDB/
https://notifications.spec.whatwg.org/
https://xhr.spec.whatwg.org/

Using the ReSpec support for this makes it slightly different.


padenot:
Thanks! We have a long way to go with those tests, but it's good to have nonetheless.

PR #1232, opened 2017-06-02
Fix #1222: Fix typo in clamping of linear distance model

### LINKED ISSUE #1222: Clamping of linear distance model
In https://webaudio.github.io/web-audio-api/#idl-def-DistanceModelType.linear, the last sentence says

> Note that dd is clamped to the interval `[dref,dmax]`

I think that should be `[dref', dmax']`. Otherwise, it doesn't match what the formula actually does with the distance.

rtoy:
Make the text match what the code says.


rtoy:
This is pretty trivial; I'm going to just merge this.

joeberkovitz:
Well OK then!

PR #1233, opened 2017-06-02
Introduce MessagePort on AudioWorkletNode/Processor

### LINKED ISSUE #1194: [audioworklet] AudioWorkletProcessor ondata should just be a callback rather than an EventHandler
I believe the intention here was to specifically deliver any ondata's sent from the main thread directly before the process method is called for that node? 

If so AudioWorkletProcessor shouldn't have `ondata` being a `EventHandler` just a callback similar to `process`.

hoch:
Preview: https://cdn.rawgit.com/hoch/web-audio-api/audioworklet-message-port/index.html

This fixes #1194.

### DIFF
--- a/index.html
+++ b/index.html
@@ -6073,9 +6073,8 @@ window.audioWorklet.import("bypass.js").then(function () {
                 Every <a>AudioWorkletNode</a> has an associated
                 <code>port</code> which is a <a href=
                 "https://html.spec.whatwg.org/multipage/comms.html#message-ports">
-                MessagePort</a>. It is connected to the other message port
-                belongs to the corresponding <a>AudioWorkletProcessor</a>
-                object.
+                MessagePort</a>. It is connected to the port on the
+                corresponding <a>AudioWorkletProcessor</a> object.
               </p>
             </dd>
           </dl>
@@ -6128,7 +6127,7 @@ window.audioWorklet.import("bypass.js").then(function () {
             <a>AudioWorkletProcessor</a> has an associated <dfn>node
             reference</dfn>, initially null.
           </p>
-          <dl title="interface AudioWorkletProcessor : EventTarget" class=
+          <dl title="interface AudioWorkletProcessor" class=
           "idl">
             <dt>
               readonly attribute AudioContextInfo contextInfo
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -6066,35 +6066,18 @@ <h2 id="AudioWorkletNode">
...
             </dt>
             <dd>
               <p>
-                The <code>ondata</code> handler is called whenever the
-                associated <a>AudioWorkletProcessor</a> sends data back to the
-                <a>AudioWorkletNode</a> instance.
+                Every <a>AudioWorkletNode</a> has an associated
+                <code>port</code> which is a <a href=
+                "https://html.spec.whatwg.org/multipage/comms.html#message-ports">
+                MessagePort</a>. It is connected to the other message port
```
rtoy: Sorry, I can't parse this sentence. Same issue below with the same description.

hoch:
Yes, this fixes #1194. (as shown in the description)

index.html
hoch: Okay, can you suggest something? We have two ports, one in AWN and the other in AWP. They are connected to each other.

hoch:
@rtoy Thanks!
@joeberkovitz Could you take a look at this change?

index.html
```diff
@@ -6066,35 +6066,17 @@ <h2 id="AudioWorkletNode">
...
             <dd>
               <p>
-                The <code>ondata</code> handler is called whenever the
-                associated <a>AudioWorkletProcessor</a> sends data back to the
-                <a>AudioWorkletNode</a> instance.
+                Every <a>AudioWorkletNode</a> has an associated
+                <code>port</code> which is a <a href=
+                "https://html.spec.whatwg.org/multipage/comms.html#message-ports">
+                MessagePort</a>. It is connected to the port on the
+                corresponding <a>AudioWorkletProcessor</a> object.
```
joeberkovitz: Suggest a little additional text here and on the other `MessagePort` stating that this is the way that Web Audio allows the AWN and AWP to send data to each other. Technically this is redundant but we could go a little further than just saying that the ports are connected.

padenot:
LGTM, merging.

PR #1237, opened 2017-06-02
Fix #1208: Clarify panning algorithms

### LINKED ISSUE #1208: Panner algorithm for equal-power panning confusing.
[Section 8.3.1](https://webaudio.github.io/web-audio-api/#Spatialzation-equal-power-panning) describes how to do equal-power panning for the `PannerNode`.  However, after step 1.3, the algorithm stops, and the text discusses `StereoPannerNode`.  It's really hard to tell, but the text after step 1.3 of the `StereoPannerNode` is (I think) the rest of the panning algorithm.

These needs to be separated out better to make it clearer that the common stuff is really common.

rtoy:
Basically split out the section on equal power panning into two
sections: One for PannerNode equalpower panning and one for
StereoPannerNode.  We don't try to pretend they share some bits; they
do, but it's not relevant here.

Because of the split, I moved the HRTF section after the equalpower
section, and put the StereoPanner section after these.  Section titles
changed to reflect better the section contents.

The algorithm for the PannerNode equalpower panning was taken from
Chrome's equalpower panning code.

### DIFF
--- a/index.html
+++ b/index.html
@@ -11951,12 +11951,11 @@ Quad up-mix:
         </p>
         <section>
           <h4 id="Spatialzation-equal-power-panning">
-            Equal-power panning
+            PannerNode "equalpower" Panning
           </h4>
           <p>
             This is a simple and relatively inexpensive algorithm which
-            provides basic, but reasonable results. It is used for the
-            <a><code>StereoPannerNode</code></a>, and for the
+            provides basic, but reasonable results. It is used for the for the
             <a><code>PannerNode</code></a> when the <a href=
             "#widl-PannerNode-panningModel"><code>panningModel</code></a>
             attribute is set to <code>"equalpower"</code>, in which case the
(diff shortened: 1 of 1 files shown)


rtoy:
Preview of the changes:  https://rawgit.com/WebAudio/web-audio-api/3d1375b/index.html#Spatialization-panning-algorithm

index.html
```diff
@@ -11997,25 +11995,79 @@ <h4 id="Spatialzation-equal-power-panning">
...
                   <p>
                     Or for a stereo input as:
                   </p>
                   <pre>
-  if (azimuth &lt;= 0) { // -90 ~ 0
+  if (azimuth &lt;= 0) { // -90 -&lt; 0
     // Transform the azimuth value from [-90, 0] degrees into the range [-90, 90].
     x = (azimuth + 90) / 90;
-  } else { // 0 ~ 90
+  } else { // 0 -&lt; 90
```
joeberkovitz: This is just a nit, but the above comments read as `// -90 -< 0` rather than `// -90 -> 0`

PR #1240, opened 2017-06-08
Remove reference to dead interface AudioMediaStreamTrack.

### LINKED ISSUE #1217: Missing AudioMediaStreamTrack definition
There is a changelog regarding the removal of AudioMediaStreamTrack in WebRTC and Media Capture and Streams specs. Is this a deprecated interface?

From [WebRTC](https://www.w3.org/TR/webrtc/):

>Changes since Dec 12, 2012
>* Changed AudioMediaStreamTrack to RTCDTMFSender and gave it its own section. Updated text to reflect most recent agreements. Also added examples section.

From [Media Capture and Streams](https://www.w3.org/TR/mediacapture-streams/):

>Changes since February 18, 2014
>* Removed videoMediaStreamTrack and audioMediaStreamTrack since they are no longer necessary.

joeberkovitz:
Fixes #1217.

Use MediaStreamTrack instead with condition that its kind equal “audio”.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2019,10 +2019,13 @@ function setupRoutingGraph() {
             </p>
             <dl class="parameters">
               <dt>
-                AudioMediaStreamTrack mediaStreamTrack
+                MediaStreamTrack mediaStreamTrack
               </dt>
               <dd>
-                The audio media stream track that will act as source.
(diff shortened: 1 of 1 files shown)


padenot:
Looks good, merging.

PR #1241, opened 2017-06-14
Define the AudioBuffer ctor.

padenot:
This fixes issue #1135

### DIFF
--- a/index.html
+++ b/index.html
@@ -4584,22 +4584,22 @@ param.setValueCurveAtTime(curve, t7, t8 - t7);
           <a><code>AudioContext</code></a>.
         </p>
         <p>
-          <a>AudioBuffer</a> have four internal slots:
+          <a>AudioBuffer</a> has four internal slots:
         </p>
         <dl>
           <dt>
             <var>[[number of channels]]</var>
           </dt>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -4600,14 +4649,16 @@ <h2 id="AudioBuffer">
           <dd>
             <p>
               The sample-rate for the PCM audio data in samples per second.
+              This MUST return the value of <var>[[sample rate]]</var>.
```
padenot: In theory, that's the style we should adopt, since we're describing something beneath the WebIDL layer. In practice, Anne told me it was OK (when really obvious) to simply refer to the attributes for now (like we do for other interfaces), since we're in a rush.

I figured that, because I was kind of rewriting stuff, that I would write it in the preferred style, and also because `duration` is not directly returned from another attribute.

padenot:
Updated with your comments addressed.

index.html
rtoy: We should probably make this all consistent. Eventually.

padenot: Yes. It's mostly just a bunch of repetitive manual work, and it's non blocking for CR, it seems.

PR #1243, opened 2017-06-14
Replace occurences of JavaScript by script and make adjustments where needed.

### LINKED ISSUE #1239: Use a term more generic than "JavaScript" for worklets
They should be able to run webassembly code as well.

padenot:
This fixes #1239.

### DIFF
--- a/index.html
+++ b/index.html
@@ -180,15 +180,15 @@
   <body>
     <section id="abstract">
       <p>
-        This specification describes a high-level JavaScript <abbr title=
+        This specification describes a high-level Web <abbr title=
         "Application Programming Interface">API</abbr> for processing and
         synthesizing audio in web applications. The primary paradigm is of an
         audio routing graph, where a number of <a><code>AudioNode</code></a>
         objects are connected together to define the overall audio rendering.
         The actual processing will primarily take place in the underlying
         implementation (typically optimized Assembly / C / C++ code), but
-        <a href="#AudioWorklet">direct JavaScript processing and synthesis</a>
-        is also supported.
+        <a href="#AudioWorklet">direct script processing and synthesis</a> is
+        also supported.
       </p>
       <p>
         The <a href="#introduction">introductory</a> section covers the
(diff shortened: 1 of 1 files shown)


hoch:
Hmm. Not sure if this is really good enough. I rather got the impression that we're losing bits by removing "JavaScript" from the entire spec. Also is WASM script? The web page says it is a "portable, size- and load-time-efficient binary format".

I really don't have a strong opinion about this, but I think we can keep the "JavaScript" until we find the appropriate substitution that includes both terms correctly.

padenot:
This is the technical term that seem to be in use throughout all web specs, for example https://html.spec.whatwg.org/.

svgeesus:
I think this is fine, particularly given the clarifying "allows developers to supply scripts (such as JavaScript or Web Assembly code)"

index.html
```diff
@@ -5771,11 +5771,12 @@ <h2 id="AudioWorklet-concepts">
             Concepts
           </h2>
           <p>
-            The <a>AudioWorklet</a> object allows developers to supply
-            JavaScript to process audio on the <a>rendering thread</a>,
-            supporting custom <a>AudioNode</a>s. This processing mechanism
-            ensures the synchronous execution of the script code with other
-            built-in <a>AudioNode</a>s in the audio graph.
+            The <a>AudioWorklet</a> object allows developers to supply scripts
+            (such as JavaScript or Web Assembly code) to process audio on the
```
hoch: WebAssembly? (without a space)

hoch:
Oh, sorry. I actually missed that part. Agreed, this is good enough.

index.html
```diff
@@ -3206,8 +3206,8 @@ <h3 id="lifetime-AudioNode">
             There are several types of references:
           </p>
           <ol>
-            <li>A <em>normal</em> JavaScript reference obeying normal garbage
-            collection rules.
+            <li>A <em>normal</em> reference obeying normal garbage collection
```
rtoy: Did you intentionally remove JavaScript instead of replacing it with script?

padenot: Yes, I think spec often talk about "references" and not "javascript references" because the fact that we're using JavaScript is an implementation detail.

Quoting the [HTML spec][html] for example:

>This specification does not require support of any particular network protocol, style sheet language, scripting language, or any of the DOM specifications beyond those required in the list above. However, the language described by this specification is biased towards CSS as the styling language, JavaScript as the scripting language, and HTTP as the network protocol, and several features assume that those languages and protocols are in use.

[html]: https://html.spec.whatwg.org/#infrastructure

index.html
padenot: WebAssembly, according to a developer working on it, thanks for the catch.

padenot:
Comments addressed, please have another look.

PR #1251, opened 2017-06-21
Clarify non-normative loop description

### LINKED ISSUE #1095: Inconsistent description of `loop` attribute behavior
@cwilso pointed out to me that https://webaudio.github.io/web-audio-api/#looping-AudioBufferSourceNode says

> If the loop attribute is true when start() is called, then playback will continue indefinitely until stop() is called and the stop time is reached.

This is inconsistent with what is described in https://webaudio.github.io/web-audio-api/#looping-AudioBufferSourceNode which implies that `loop` can be dynamically modified.

joeberkovitz:
Fix #1095 by bringing non-normative description into line with most recent changes.

### DIFF
--- a/index.html
+++ b/index.html
@@ -5034,10 +5034,9 @@ param.setValueCurveAtTime(curve, t7, t8 - t7);
           </dt>
           <dd>
             <p>
-              Indicates if the audio data should play in a loop. The default
-              value is false. If <code>loop</code> is dynamically modified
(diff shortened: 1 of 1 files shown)


joeberkovitz:
Preview at https://rawgit.com/WebAudio/web-audio-api/083762efab60b0125afe57d731a85b6389501539/index.html#looping-AudioBufferSourceNode

PR #1254, opened 2017-06-21
Force enteredLoop flag to false on next render quantum for ABSN, iff loop is false

### LINKED ISSUE #1252: Setting loop to false for ABSN should reset entered-loop state
Currently the speced algorithm for ABSN looping "remembers" that the looped region of the buffer was entered by the playhead, even when `loop` is set to `false`. This means that setting `loop` back to `true` will forcibly wrap the playhead into the looped region, even if it is outside the loop at that time.

Scenario:
- assume buffer length of 3
- set loopStart to 1
- set loopEnd to 2
- set loop to true
- start playback until playhead reaches 1.5
- set loop to false
- playhead reaches 2.5
- set loop to true again
- playhead now is forced to 1.5, whereas in a "fresh" ABSN with loop=false the playhead would continue from 2.5 (since the loop was never entered).

This behavior feels inconsistent. Setting `loop` to `false` should reset the remembered state that the loop was previously entered.

joeberkovitz:
Fixes #1252

### DIFF
--- a/index.html
+++ b/index.html
@@ -5445,6 +5445,10 @@ function process(numberOfFrames) {
       actualLoopEnd = buffer.duration;
     }
   }
+  else {
+    // If the loop flag is false, remove any record of the loop having been entered
+    enteredLoop = false;
+  }
 
   // Render each sample frame in the quantum
   for (let index = 0; index &lt; numberOfFrames; index++) {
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -5445,6 +5445,10 @@ <h3 id="playback-AudioBufferSourceNode">
       actualLoopEnd = buffer.duration;
     }
   }
+  else {
```
rtoy: It seems most of the other code uses this style:
```
} else {
```

PR #1257, opened 2017-06-21
Specify that AudioParam.value returns the most recent computed value.

### LINKED ISSUE #1081: Description for the value when reading AudioParam.value
The description, "Subsequent accesses to this attribute's getter will return the same value."
in 2.6.1 AudioParam.value description is confusable.

This explanation has caused a difference in interpretation between Chrome and Firefox for a long time.
I think it would be better to improve the description to make the specification unambiguous.

joeberkovitz:
Fixes #1081

### DIFF
--- a/index.html
+++ b/index.html
@@ -3479,8 +3479,9 @@ function setupRoutingGraph() {
           </li>
         </ul>
         <p>
-          When a new <a>AudioParam</a> is created, its internal <var>[[current
-          value]]</var> slot is set to the parameter's default value.
+          Each <a>AudioParam</a> has an internal slot <var>[[current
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -3478,6 +3478,10 @@ <h2 id="AudioParam">
             duration D.</span>
           </li>
         </ul>
+        <p>
+          When a new <a>AudioParam</a> is created, its internal <var>[[current
+          value]]</var> slot is set to the parameter's default value.
```
padenot: Usually written like so:

```html
Each <a>AudioParam</a> has an internal slot <var>[[current
value]]</var>, initially set to the value of the <a>AudioParam</a>'s <a href="/#widl-AudioParam-defaultValue">default value</a>. 
```

PR #1263, opened 2017-06-21
Canonicalize all the variations of 3-vector spec code to use common definitions.

### LINKED ISSUE #1236: Level of detail in code descriptions for algorithms?
In https://webaudio.github.io/web-audio-api/#Spatialization-distance-effects we have this bit of code:

```
function dotProduct(v1, v2) {
    var d = 0;
    for (var i = 0; i < Math.min(v1.length, v2.length); i++)
      d += v1[i] * v2[i];
    return d;
  }
  var v = panner.position - listener.position;
  var distance = Math.sqrt(dotProduct(v, v));
```
So we describe and give an algorithm for the `dotProduct`.  But then we also write
```
panner.position - listener.position
```
First, `panner.position` and `listener.position` doesn't really exist anywhere.  The intent is pretty clear that these represent a 3D vector of the positions.  But `panner.position - listener.position` isn't really valid JS either.

And curiously, https://webaudio.github.io/web-audio-api/#Spatialization-sound-cones has yet another (identical) definition for `dotProduct` and also introduces a `diff` function for the difference of two vectors.

We need to decide exactly what level of detail we need and decide if the code is pseudo-code or valid JS.  And remove the duplicated functions.

joeberkovitz:
Fixes #1236

### DIFF
--- a/index.html
+++ b/index.html
@@ -12419,18 +12419,18 @@ function coneGain() {
     new Vec3(panner.positionX, panner.positionY, panner.positionZ);
   let listenerPosition =
     new Vec3(listener.positionX, listener.positionY, listener.positionZ);
-  var sourceToListener = sourcePosition.diff(listenerPosition).normalize();
+  let sourceToListener = sourcePosition.diff(listenerPosition).normalize();
 
-  var normalizedSourceOrientation = sourceOrientation.normalize();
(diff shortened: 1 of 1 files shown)


padenot:
LGTM.

PR #1270, opened 2017-07-07
Update Makefile and gitignore

hoch:
@padenot PTAL. I had to do this to bypass the version check in the makefile so I can use my local tidy (OSX 5.4).


hoch:
Thanks for the quick response! :)

PR #1271, opened 2017-07-13
Fix 1089: Specify channel order

jdsmith3000:


jdsmith3000:
I don't quite like the table formatting, but think it's good enough.  Can someone advise me on the check failures?  I don't see any failures specific to my changes in "details".

hoch:
[Preview](https://cdn.rawgit.com/jdsmith3000/web-audio-api/87a4fac5202f3a95c714458a7d0f8a6c53774205/index.html#ChannelOrdering)

Things I've noticed on the log:
- line 202~203 goes beyond 80 cols.
- Line breaks are needed for all the tags and their contents.

You can simply install html-tidy (>= v5.4) and run it on the file via `make tidy`.

jdsmith3000:
The only error I see in Details is "make: *** [check] Error 1", which sounds like a makefile problem.  What am I missing here?

I've tried running html-tidy on index.html and get no warnings or errors, even when using tidyconf.txt from our repository as the tidy config file.

hoch:
Here's an excerpt from the log.

```html
         <p>
-          Channel ordering is defined by the following table.  Individual multichannel formats MAY not support all intermediate channels.  
-          Implementations MUST present the channels provided in the order defined below, skipping over those channels not present.
+          Channel ordering is defined by the following table. Individual
+          multichannel formats MAY not support all intermediate channels.
+          Implementations MUST present the channels provided in the order
+          defined below, skipping over those channels not present.
         </p>
```

For example, the lines starting with `-` means it did not pass the tidy check and must be replaced by the lines starting with `+`.

You get the `make:...` error because we use `make` to run the tidy check.

jdsmith3000:
@hoch Thanks for the instruction.  I decided to manually correct the edits to learn my formatting errors.  I took me a few passes to recognize dangling/extra spaces...

hoch:
@jdsmith3000 No problem. The table looks pretty!

PR #1276, opened 2017-07-20
deal with all the MUST and most of the SHOULD

svgeesus:
A few of the should are left lowercase where I wasn't sure.
Most should are plain-english and not conformance statements.

Turns our ReSpec automatically wraps MUST in RFC-2119 markup and leaves must alone, so the editing turned out to be super easy.


svgeesus:
Resolving conflicts before re-opening

index.html
```diff
@@ -7191,6 +7191,7 @@ <h2 id="AudioWorklet-Examples">
             <dd>
               <p>
                 An AudioBuffer where the output audio data should be written.
+                <!-- should? MUST? -->
```
joeberkovitz: I think this would be a MUST

svgeesus: Agreed

index.html
```diff
@@ -7190,8 +7190,8 @@ <h2 id="AudioWorklet-Examples">
             </dt>
             <dd>
               <p>
-                An AudioBuffer where the output audio data MUST be written. 
-                 It will have a number of channels equal to the
+                An AudioBuffer where the output audio data MUST be written. It
+                will have a number of channels equal to the
```
rtoy: "It will" or "It MUST"?  Not really sure if we should bother to change this here.

index.html
```diff
@@ -2716,14 +2716,14 @@ <h2 id="OfflineAudioCompletionEvent">
             <pre class="example">
     nodeA.connect(nodeB);
     nodeA.connect(nodeB);
-    
+
```
rtoy: Why the extra blank line?  Did tidy do this?

Same comment applies to all the other extra blank lines below at the end of a `<pre>` section.

index.html
```diff
@@ -2716,14 +2716,14 @@ <h2 id="OfflineAudioCompletionEvent">
...
 </pre>
             <p>
               will have the same effect as
             </p>
             <pre class="example">
       nodeA.connect(nodeB);
-    
+
```
rtoy: Why the extra blank line?

index.html
```diff
@@ -6558,13 +6558,13 @@ <h2 id="defining-a-valid-audioworkletprocessor">
               <ul>
                 <li>Nodes that transform their inputs, and are active only
                 while connected inputs and/or script references exist. Such
-                nodes should return <code>false</code> from
+                nodes SHOULD return <code>false</code> from
```
rtoy: I wonder if these "SHOULD" should actually be "MUST".

@hoch and @joeberkovitz: WDYT?

index.html
```diff
@@ -12933,7 +12933,7 @@ <h2 id="Performance">
           When an <a href="#acquire-the-content">acquire the content</a>
           operation is performed on an <a>AudioBuffer</a>, the entire operation
           can usually be implemented without copying channel data. In
-          particular, the last step should be performed lazily at the next
+          particular, the last step SHOULD be performed lazily at the next
```
rtoy: "MUST" or "SHOULD"?

index.html
joeberkovitz: I'm indifferent but MUST seems appropriate.

index.html
joeberkovitz: I'd say SHOULD is completely appropriate here since we're offering the user options on how to employ the procedure

index.html
joeberkovitz: seems like a SHOULD to me since other UA strategies could conceivably deliver the desired goal of minimizing copies and allocations... no normative difference could be observed if the UA failed to do so

index.html
rtoy: Fair enough.  This is ok.  But I also notice that in the bullet items after this one, there are also some "should" references that look pretty much like this one.  Should these be converted to "SHOULD" instead of "should" too?

index.html
rtoy: Works for me.

index.html
joeberkovitz: The other bullets seem to say SHOULD, at least in the current PR I'm looking at (lines 6567 and 6576).

rtoy:
Any reason not to merge this now? It's been approved by all reviewers and there are no conflicts with the base branch.

PR #1278, opened 2017-07-26
Specify the DynamicsCompressor

padenot:
Here is it. It should be fairly complete (I addressed the comments we made during the f2f).

### DIFF
--- a/index.html
+++ b/index.html
@@ -8940,122 +8940,161 @@ function calculateNormalizationScale(buffer)
             Processing
           </h2>
           <p>
-          Dyamics compression can be implemented in a variety of ways. The
-          <a>DynamicsCompressorNode</a> implements a dynamics processor that has
-          the following characteristics:
+            Dynamics compression can be implemented in a variety of ways. The
+            <a>DynamicsCompressorNode</a> implements a dynamics processor that
+            has the following characteristics:
+          </p>
           <ul>
             <li>Fixed look-ahead (this means that an
             <a>DynamicsCompressorNode</a> adds a fixed latency to the signal
-            chain).</li>
-            <li>Configurabe attack speed, release speed, threshold, knee
-            hardness and ratio.</li>
-            <li>Side-chaining is not supported</li>
-            <li>The gain reduction is reported</li>
+            chain).
+            </li>
+            <li>Configurable attack speed, release speed, threshold, knee
+            hardness and ratio.
+            </li>
+            <li>Side-chaining is not supported
+            </li>
+            <li>The gain reduction is reported <em>via</em> the
+            <code>reduction</code> property on the
+            <a>DynamicsCompressorNode</a>
+            </li>
+            <li>The compression curve is has three parts:
+              <ul>
+                <li>The first part is a linear function of equation \(f(x) =
+                y\).
+                </li>
+                <li>The second part is the soft-knee portion, which MUST be a
+                monotonically increasing function.
+                </li>
+                <li>The third part is a linear function of equation \(f(x) =
+                \frac{1}{ratio} y \).
+                </li>
+              </ul>This curve MUST be continuous.
+            </li>
           </ul>
+          <p>
+            Graphically, such a curve would look something like this:
           </p>
+          <figure>
+            <img alt="Graphical representation of a compression curve" src=
+            "images/compression-curve.svg" style="width: 50%">
+            <figcaption>
+              A typical compression curve, showing the knee portion (soft or
+              hard) as well as the threshold.
+            </figcaption>
+          </figure>
           <p>
             Internally, the <a>DynamicsCompressorNode</a> is described with a
             combination of other <a>AudioNode</a>s, as well as a special
             algorithm, to compute the gain reduction value.
           </p>
           <p>
-          The following <a>AudioNode</a> graph is used internally,
-          <code>input</code> and <code>output</code> respectively being the
-          input and output <a>AudioNode</a>, <code>context</code> the
-          <a>BaseAudioContext</a> for this <a>DynamicsCompressorNode</a>, and
-          <code>CompressionProcessor</code> a special object that behaves like
-          an <a>AudioNode</a>, described below:
-            <pre>
+            The following <a>AudioNode</a> graph is used internally,
+            <code>input</code> and <code>output</code> respectively being the
+            input and output <a>AudioNode</a>, <code>context</code> the
+            <a>BaseAudioContext</a> for this <a>DynamicsCompressorNode</a>, and
+            new class, <code>CompressionProcessor</code>, that instantiate a
+            special object that behaves like an <a>AudioNode</a>, described
+            below:
+          </p>
+          <pre>
               var delay = new DelayNode(context, {delayTime: 0.006});
               var gain = new GainNode(context);
+              var compression = new CompressionProcessor();
 
               input.connect(delay).connect(gain).connect(output);
-              input.connect(CompressionProcessor).connect(gain.gain);
+              input.connect(compression).connect(gain.gain);
             </pre>
-            <div class=note>
-              This implements the pre-delay and the application of the reduction
-              gain.
-            </div>
-          </p>
+          <div class="note">
+            This implements the pre-delay and the application of the reduction
+            gain.
+          </div>
           <p>
-          The following algorithm describes the
-          <code>CompressionProcessor</code> to be applied to the input signal to
-          produce the gain reduction value. First, a couple variables that
-          persist accross invocation of this algorithm, tied to this particular
-          <code>CompressionProcessor</code>.
+            The following algorithm describes the
+            <code>CompressionProcessor</code> object to be applied to the input
+            signal to produce the gain reduction value. First, a couple
+            variables that persist accross invocation of this algorithm, tied
+            to this particular <code>CompressionProcessor</code> instance.
+          </p>
           <ul>
             <li>Let <var>detector average</var> be a floating point number,
-            initialized to 1.0.</li>
+            initialized to 1.0.
+            </li>
             <li>Let <var>compression gain</var> be a foating point number,
-            initialized to 1.0.</li>
+            initialized to 1.0.
+            </li>
           </ul>
           <p>
-            Now, the actual algorithm to determine the reduction gain, in linear
-          gain value.
+            Now, the actual algorithm to determine the reduction gain, in
+            linear gain value.
           </p>
-            <ol>
-              <li>
-                Let <var>input signal</var> be the the block of audio to be
-                processed.
-              </li>
-              <li>
-                Let <var>treshold</var>, <var>knee</var>, <var>ratio</var>,
-                <var>attack</var> and <var>release</var> have the value of the
-                <a>AudioParam</a> of the same name, sampled at the time of
-                processing (those are <a>k-rate</a> parameters).
-              </li>
-              <li>
-                Set <var>attack</var> and <var>release</var> to the division
-                between themselves and the sample-rate for this
-                <a>BaseAudioContext</a>.
-              </li>
-              <li>
-                For each samples <var>input</var> of <var>input signal</var>,
-              execute the following steps:
+          <ol>
+            <li>Let <var>input signal</var> be the the block of audio to be
+            processed.
+            </li>
+            <li>Let <var>treshold</var>, <var>knee</var>, <var>ratio</var>,
+            <var>attack</var> and <var>release</var> have the value of the <a>
+              AudioParam</a> of the same name, sampled at the time of
+              processing (those are <a>k-rate</a> parameters).
+            </li>
+            <li>Set <var>attack</var> and <var>release</var> to the division
+            between themselves and the sample-rate for this
+            <a>BaseAudioContext</a>.
+            </li>
+            <li>For each samples <var>input</var> of <var>input signal</var>,
+            execute the following steps:
               <ol>
                 <li>Let <var>envelope rate</var> be the result of computing the
-                <a href=envelope-rate>envelope rate</a> </li>
-                <li>
-                Let <var>attenuation</var> be the value of putting the absolute
-                value of <var>input</var> <a href=compression-curve>through the compression curve</a>.
+                <a href="envelope-rate">envelope rate</a>
                 </li>
-                <li> Add the difference between <var>attenuation</var> and
-                <var>detector average</var> to <var>detector average</var>, and
-                clamp <var>detector average</var> to 1.0.</li>
+                <li>Let <var>attenuation</var> be the value of putting the
+                absolute value of <var>input</var> <a href="compression-curve">
+                  through the compression curve</a>.
+                </li>
+                <li>Add the difference between <var>attenuation</var> and <var>
+                  detector average</var> to <var>detector average</var>, and
+                  clamp <var>detector average</var> to 1.0.
+                </li>
                 <li>TODO If attacking, add <var>detector average</var> minus
-                <var>compressor gain</var> to <var>compressor gain</var></li>
+                <var>compressor gain</var> to <var>compressor gain</var>
+                </li>
                 <li>Else, multiply <var>compressor gain</var> by <var>enveloppe
-                  rate</var></li>
+                rate</var>
+                </li>
                 <li>Compute <var>final gain</var> to be <var>compressor
-                  gain</var> multiplied by the return value of <a>computing the
-                  makeup gain</a>.
+                gain</var> multiplied by the return value of <a>computing the
+                makeup gain</a>.
+                </li>
                 <li>Compute <var>metering gain</var> to be <var>final
                 gain</var>, <a href="#todo">converted to decibel</a>.
+                </li>
               </ol>
-              </li>
-            </ol>
-            <p>
+            </li>
+          </ol>
+          <p>
             <dfn>Computing the makeup gain</dfn> means executing the following
             steps:
-            <ol>
-              <li>XXX todo</li>
-            </ol>
-            </p>
-            <p>
-              Computing the <dfn>enveloppe rate</dfn> means executing the
-              following steps:
-              <ol>
-                <li>XXX TODO</li>
-              </ol>
-            </p>
-            <p>
-              Putting a sample through a <dfn>compression curve</dfn> means executing the
-            following steps:
-            <ol>
-              <li>XXX TODO</li>
-            </ol>
-            </p>
           </p>
+          <ol>
+            <li>XXX todo
+            </li>
+          </ol>
+          <p>
+            Computing the <dfn>enveloppe rate</dfn> means executing the
+            following steps:
+          </p>
+          <ol>
+            <li>XXX TODO
+            </li>
+          </ol>
+          <p>
+            Putting a sample through a <dfn>compression curve</dfn> means
+            executing the following steps:
+          </p>
+          <ol>
+            <li>XXX TODO
+            </li>
+          </ol>
         </section>
       </section>
       <section>
--- a/index.html
+++ b/index.html
@@ -8935,210 +8935,6 @@ function calculateNormalizationScale(buffer)
             </dd>
           </dl>
         </section>
-        <section>
-          <h2>
-            Processing
-          </h2>
-          <p>
-            Dynamics compression can be implemented in a variety of ways. The
-            <a>DynamicsCompressorNode</a> implements a dynamics processor that
-            has the following characteristics:
-          </p>
-          <ul>
-            <li>Fixed look-ahead (this means that an
-            <a>DynamicsCompressorNode</a> adds a fixed latency to the signal
-            chain).
-            </li>
-            <li>Configurable attack speed, release speed, threshold, knee
-            hardness and ratio.
-            </li>
-            <li>Side-chaining is not supported
-            </li>
-            <li>The gain reduction is reported <em>via</em> the
-            <code>reduction</code> property on the
-            <a>DynamicsCompressorNode</a>
-            </li>
-            <li>The compression curve is has three parts:
-              <ul>
-                <li>The first part is the identify: \(f(x) = x\).
-                </li>
-                <li>The second part is the soft-knee portion, which MUST be a
-                monotonically increasing function.
-                </li>
-                <li>The third part is a linear function of equation \(f(x) =
-                \frac{1}{ratio} \cdot x \).
-                </li>
-              </ul>
-              This curve MUST be continuous.
-            </li>
-          </ul>
-          <p>
-            Graphically, such a curve would look something like this:
-          </p>
-          <figure>
-            <img alt="Graphical representation of a compression curve" src=
-            "images/compression-curve.svg" style="width: 50%">
-            <figcaption>
-              A typical compression curve, showing the knee portion (soft or
-              hard) as well as the threshold.
-            </figcaption>
-          </figure>
-          <p>
-            Internally, the <a>DynamicsCompressorNode</a> is described with a
-            combination of other <a>AudioNode</a>s, as well as a special
-            algorithm, to compute the gain reduction value.
-          </p>
-          <p>
-            The following <a>AudioNode</a> graph is used internally,
-            <code>input</code> and <code>output</code> respectively being the
-            input and output <a>AudioNode</a>, <code>context</code> the
-            <a>BaseAudioContext</a> for this <a>DynamicsCompressorNode</a>, and
-            new class, <code>CompressionProcessor</code>, that instantiate a
-            special object that behaves like an <a>AudioNode</a>, described
-            below:
-          </p>
-          <pre>
-              var delay = new DelayNode(context, {delayTime: 0.006});
-              var gain = new GainNode(context);
-              var compression = new CompressionProcessor();
-
-              input.connect(delay).connect(gain).connect(output);
-              input.connect(compression).connect(gain.gain);</pre>
-          <div class="note">
-            This implements the pre-delay and the application of the reduction
-            gain.
-          </div>
-          <p>
-            The following algorithm describes the
-            <code>CompressionProcessor</code> object to be applied to the input
-            signal to produce the gain reduction value. First, a couple
-            variables that persist accross invocation of this algorithm, tied
-            to this particular <code>CompressionProcessor</code> instance.
-          </p>
-          <ul>
-            <li>Let <var>detector average</var> be a floating point number,
-            initialized to 1.0.
-            </li>
-            <li>Let <var>compression gain</var> be a foating point number,
-            initialized to 1.0.
-            </li>
-          </ul>
-          <p>
-            Now, the actual algorithm to determine the reduction gain, in
-            linear gain value.
-          </p>
-          <ol>
-            <li>Let <var>input signal</var> be the the block of audio to be
-            processed.
-            </li>
-            <li>Let <var>treshold</var>, <var>knee</var> have the value of the
-            <a>AudioParam</a> of the same name, <a>converted to linear unit</a> sampled at the time of
-              processing of this block (as <a>k-rate</a> parameters).
-            </li>
-            <li>Let <var>ratio</var> have the value of the <code>ratio</code>
-            <a>Audioparam</a>, sampled at the time of processing of this block
-            (as a <a>k-rate</a> parameter).</li>
-            <li>Let <var>attack</var> and <var>release</var> have the value of the <a>
-              AudioParam</a> of the same name, sampled at the time of
-              processing (those are <a>k-rate</a> parameters), converted to
-              sample-frames, based on the sample-rate of the
-              <a>BaseAudioContext</a> this <a>DynamicsCompressorNode</a> is
-              <a href=#associated>associated</a> with.
-            </li>
-            <li>Set <var>attack</var> and <var>release</var> to the division
-            between themselves and the sample-rate for this
-            <a>BaseAudioContext</a>.
-            </li>
-            <li>For each samples <var>input</var> of <var>input signal</var>,
-            execute the following steps:
-              <ol>
-                <li>Let <var>envelope rate</var> be the result of computing the
-                <a href="envelope-rate">envelope rate</a>
-                </li>
-                <li>Let <var>attenuation</var> be the value of putting the
-                absolute value of <var>input</var> <a href="compression-curve">
-                  through the compression curve</a>.
-                </li>
-                <li>Add the difference between <var>attenuation</var> and <var>
-                  detector average</var> to <var>detector average</var>, and
-                  clamp <var>detector average</var> to 1.0.
-                </li>
-                <li>TODO If attacking, add <var>detector average</var> minus
-                <var>compressor gain</var> to <var>compressor gain</var>
-                </li>
-                <li>Else, multiply <var>compressor gain</var> by <var>enveloppe
-                rate</var>
-                </li>
-                <li>Compute <var>final gain</var> to be <var>compressor
-                gain</var> multiplied by the return value of <a>computing the
-                makeup gain</a>.
-                </li>
-                <li>Compute <var>metering gain</var> to be <var>final
-                gain</var>, <a href="#todo">converted to decibel</a>.
-                </li>
-              </ol>
-            </li>
-          </ol>
-          <p>
-            <dfn>Computing the makeup gain</dfn> means executing the following
-            steps:
-          </p>
-          <ol>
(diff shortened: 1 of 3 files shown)


padenot:
Preview: https://rawgit.com/padenot/web-audio-api/4fa027223c5972b72c0d587c5eb736115d558a66/index.html#processing

svgeesus:
This seems fairly clearly explained, to me. (A couple of typos - treashold -> threshold, Enveloppe -> Envelope).
We should probably log an actual EnvelopeFollower node as a post-v1 feature request, as it is a useful building block.

index.html
```diff
@@ -9222,6 +9222,240 @@ <h3 id="example-2">
...
+                </li>
+                <li>Compute <var>metering gain</var> to be <var>final
+                gain</var>, <a href="#linear-to-decibel">converted to
+                decibel</a>.
+                </li>
+              </ol>
+            </li>
+          </ol>
+          <p>
+            The makeup gain is a fixed gain stage that only depends on ratio,
```
rtoy: Maybe add `<dfn>makeup gain</dfn>` here?  And links to this from other places?

index.html
```diff
@@ -9222,6 +9222,240 @@ <h3 id="example-2">
...
+          </p>
+          <ol>
+            <li>Let <var>full range gain</var> be the value returned by putting
+            1.0 through the <a>compression curve</a>.
+            </li>
+            <li>Let <var>full range makeup gain</var> the inverse of <var>full
+            range gain</var>.
+            </li>
+            <li>Return the result of computing <var>full range makeup
+            gain</var><sup>0.6</sup>.
```
rtoy: This is pretty hard to read to notice that it's the 0.6 power. (I initially read it as footnote number 0.6!).  Maybe just say "take the 0.6 power of full range makeup gain"?

index.html
```diff
@@ -9222,6 +9222,240 @@ <h3 id="example-2">
...
+            ratio of the <var>compressor gain</var> and the <var>detector
+            average</var> through a curve. User-agents are allowed to choose
+            the shape of the attack and release curve. However, they MUST
+            respect the following constraints:
+          </p>
+          <ul>
+            <li>The input to the curve MUST be the ratio of the <var>compressor
+            gain</var> and the <var>detector average</var>.
+              <div class="note">
+                When attacking, this number inferior or equal to 1, when
```
rtoy: "inferior or equal to" -> "lis ess than or equal to"?

index.html
```diff
@@ -9222,6 +9222,240 @@ <h3 id="example-2">
...
+                releasing, this number is strictly superior to 1.
```
rtoy: "superior" -> "greater than"?

index.html
```diff
@@ -9222,6 +9222,240 @@ <h3 id="example-2">
...
+            </li>
+          </ul>
+          <div class="note">
+            It is allowed, for example, to have a compressor that performs an
+            <em>adaptative release</em>, that is, releasing faster the harder
+            the compression, or to have curves for attack and release that are
+            not of the same shape.
+          </div>
+          <p>
+            Putting a the absolute value of a sample through a <dfn>compression
```
rtoy: You use the phrase "putting foo through a curve" in many places above.  Maybe add a `<dfn>` for this and reference it?

index.html
```diff
@@ -9222,6 +9222,240 @@ <h3 id="example-2">
...
+            curve</dfn> means computing the absolute value of this sample to a
+            function, and returning the computed value. This function MUST
+            respects the following characteristics:
+          </p>
+          <ol>
+            <li>This function is the identify up to the value of the linear
```
rtoy: Typo: "identify" -> "identity"?

index.html
```diff
@@ -9222,6 +9222,240 @@ <h3 id="example-2">
...
+            <code>threshold</code> (i.e., \(f(sample) = sample\)).
```
rtoy: Maybe just use f(x) = x?  Easier to read as a mathematical formula.

index.html
```diff
@@ -9222,6 +9222,240 @@ <h3 id="example-2">
...
+            </li>
+            <li>From the <code>threshold</code> up to the <code>threshold +
+            knee</code>, User-Agents can choose the curve shape. The whole
+            function MUST be continuous (except as noted in the note below) and
```
rtoy: "except as noted in the note below" -> "except as noted below"?

index.html
```diff
@@ -9222,6 +9222,240 @@ <h3 id="example-2">
...
+            monotonically increasing.
+              <div class="note">
+                As such, if the <code>knee</code> is 0, the
+                <a>DynamicsCompressorNode</a> is a hard-knee compressor, and
+                the function has a discontinuity at <code>threshold</code>.
```
rtoy: This seems to contradict the requirement that the curve MUST be continuously differentiable.

index.html
```diff
@@ -9222,6 +9222,240 @@ <h3 id="example-2">
...
+              </div>
+            </li>
+            <li>This function returned is linear, based on the ratio, after the
+            <code>threshold</code> and the soft knee (i.e., \(f(sample) =
```
rtoy: Copy the formula from earlier so that it has the same form (1/ratio * x).

index.html
```diff
@@ -9222,6 +9222,240 @@ <h3 id="example-2">
...
+              <ul>
+                <li>The first part is the identify: \(f(x) = x\).
+                </li>
+                <li>The second part is the soft-knee portion, which MUST be a
+                monotonically increasing function.
+                </li>
+                <li>The third part is a linear function of equation \(f(x) =
+                \frac{1}{ratio} \cdot x \).
+                </li>
+              </ul>This curve MUST be continuously differentiable, and
```
padenot: I think we need to write that it is allowed to have a function that is not continuously differentiable iff the knee is 0, which kind of makes sense, because the purpose of a hard knee is to have an hard edge.

index.html
```diff
@@ -9222,6 +9222,240 @@ <h3 id="example-2">
...
+            </li>
+            <li>The compression curve is has three parts:
+              <ul>
+                <li>The first part is the identify: \(f(x) = x\).
+                </li>
+                <li>The second part is the soft-knee portion, which MUST be a
+                monotonically increasing function.
+                </li>
+                <li>The third part is a linear function of equation \(f(x) =
+                \frac{1}{ratio} \cdot x \).
```
padenot: I would have done that in proper LaTeX, but the layout is weird in mathjax.

index.html
```diff
@@ -9222,6 +9222,240 @@ <h3 id="example-2">
...
+            <a href="#associated">associated</a> with.
+            </li>
+            <li>Set <var>attack</var> and <var>release</var> to the
+            multiplication between themselves and the sample-rate for this <a>
+              BaseAudioContext</a>.
+            </li>
+            <li>For each samples <var>input</var> of <var>input signal</var>,
+            execute the following steps:
+              <ol>
+                <li>Let <var>envelope rate</var> be the result of computing the
```
padenot: `[[foo]]` refer to internal slots of objects (it's the ECMAScript notation). `<var></var>` is recommended for algorithm variables.

padenot: Also I'm switching the names around a bit so that it's less confusing.

index.html
```diff
@@ -9222,6 +9222,240 @@ <h3 id="example-2">
...
+                <a href="envelope-rate">envelope rate</a>
+                </li>
+                <li>Let <var>attenuation</var> be the value of putting the
```
padenot: Yes it's defined below:

> Putting a the absolute value of a sample through a <dfn>compression curve</dfn> means computing the absolute value of this sample to a function, and returning the computed value. This function MUST respects the following characteristics:

index.html
```diff
@@ -9222,6 +9222,240 @@ <h3 id="example-2">
...
+                absolute value of <var>input</var> <a href="compression-curve">
+                  through the compression curve</a>.
+                </li>
+                <li>Add the difference between <var>attenuation</var> and <var>
+                  detector average</var>, multiplied by the <var>enveloppe
+                  rate</var> to <var>detector average</var>, and clamp
```
padenot: I'll be spelling out this computation in mode details, so that the precedence is very clear.

index.html
```diff
@@ -9222,6 +9222,240 @@ <h3 id="example-2">
...
+                  <var>detector average</var> to 1.0.
+                </li>
+                <li>If attacking, add <var>detector average</var> minus
+                <var>compressor gain</var>, multiplied by <var>enveloppe
+                rate</var> to <var>compressor gain</var>, and clamp
```
padenot: Spelled it out in more detail.

index.html
```diff
@@ -9222,6 +9222,240 @@ <h3 id="example-2">
...
+                <a href="envelope-rate">envelope rate</a>
+                </li>
+                <li>Let <var>attenuation</var> be the value of putting the
+                absolute value of <var>input</var> <a href="compression-curve">
+                  through the compression curve</a>.
+                </li>
+                <li>Add the difference between <var>attenuation</var> and <var>
+                  detector average</var>, multiplied by the <var>enveloppe
+                  rate</var> to <var>detector average</var>, and clamp
+                  <var>detector average</var> to 1.0.
```
padenot: I made it clearer.

index.html
```diff
@@ -9222,6 +9222,240 @@ <h3 id="example-2">
...
+                </li>
+                <li>If attacking, add <var>detector average</var> minus
+                <var>compressor gain</var>, multiplied by <var>enveloppe
+                rate</var> to <var>compressor gain</var>, and clamp
+                <var>compressor gain</var> to 1.0.
```
padenot: Made this one clearer as well.

index.html
padenot: Oops I forgot to send this one back to the main thread so that authors can read the value back and know how hard the compressor is being hit.

index.html
padenot: I trust you on the how to express this in English, I learned maths in french :-)

index.html
```diff
@@ -9222,6 +9222,240 @@ <h3 id="example-2">
...
+            special object that behaves like an <a>AudioNode</a>, described
+            below:
+          </p>
+          <pre>
+              var delay = new DelayNode(context, {delayTime: 0.006});
+              var gain = new GainNode(context);
+              var compression = new EnveloppeFollower();
+
+              input.connect(delay).connect(gain).connect(output);
+              input.connect(compression).connect(gain.gain);</pre>
```
padenot: I added a picture as well.

index.html
```diff
@@ -9222,6 +9222,240 @@ <h3 id="example-2">
...
+          <p>
+            Now, the actual algorithm to determine the reduction gain, in
+            linear gain value.
+          </p>
+          <ol>
+            <li>Let <var>input signal</var> be the the block of audio to be
+            processed.
+            </li>
+            <li>Let <var>threshold</var>, <var>knee</var> have the value of the
+            <a>AudioParam</a> of the same name, <a>converted to linear unit</a>
```
padenot: Added that.

index.html
```diff
@@ -9222,6 +9222,240 @@ <h3 id="example-2">
...
+                <var>compressor gain</var> to 1.0.
+                </li>
+                <li>Else, multiply <var>compressor gain</var> by <var>enveloppe
+                rate</var>
+                </li>
+                <li>Compute <var>final gain</var> to be <var>compressor
+                gain</var> multiplied by the return value of <a>computing the
+                makeup gain</a>.
+                </li>
+                <li>Compute <var>metering gain</var> to be <var>final
```
padenot: I've added a thing on this, it's actually important.

index.html
padenot: Fixed in a different way.

index.html
padenot: I think the MUST applies to the whole sentence: it MUST (be continuously differentiable except when the knee is 0, and then that is the case there is a discontinuity at the knee).

I've changed the prose a bit.

index.html
```diff
@@ -9222,6 +9222,240 @@ <h3 id="example-2">
...
+            parameters).
+            </li>
+            <li>Let <var>ratio</var> have the value of the <code>ratio</code>
+            <a>Audioparam</a>, sampled at the time of processing of this block
+            (as a <a>k-rate</a> parameter).
+            </li>
+            <li>Let <var>attack</var> and <var>release</var> have the value of
+            the <a>AudioParam</a> of the same name, sampled at the time of
+            processing (those are <a>k-rate</a> parameters), converted to
+            sample-frames, based on the sample-rate of the
```
padenot: I clarified that. It's kept in float.

index.html
```diff
@@ -9222,6 +9222,240 @@ <h3 id="example-2">
...
+            <a>BaseAudioContext</a> this <a>DynamicsCompressorNode</a> is
+            <a href="#associated">associated</a> with.
+            </li>
+            <li>Set <var>attack</var> and <var>release</var> to the
+            multiplication between themselves and the sample-rate for this <a>
+              BaseAudioContext</a>.
```
padenot: This is a mistake, this is kept in float.

joeberkovitz:
Latest preview for review: https://rawgit.com/padenot/web-audio-api/ec6b3b2c1cbab235d57b07aa97501452254d3462/index.html

index.html
```diff
@@ -9222,6 +9225,256 @@ <h3 id="example-2">
...
+                <li>Let <var>attenuation</var> be the value of applying the
+                <a href="compression-curve">compression curve</a> to the
+                absolute value of <var>input</var> .
+                </li>
+                <li>Set <var>change-rate</var> to the difference between <var>
+                  attenuation</var> and <var>detector average</var>. Multiply
+                  <var>change-rate</var> by the <var>envelope rate</var>. Add
+                  <var>change-rate</var> to <var>detector average</var>, and
+                  clamp <var>detector average</var> to a maximum of 1.0.
+                </li>
```
joeberkovitz: I think there's a mistake here. In the algorithm, change-rate is not multiplied by the envelope rate, but by a different variable named `rate`, which in turn is derived from the  attenuation in dB divided by the baked-in parameter `sat_release_frames`.

index.html
```diff
@@ -9222,6 +9225,256 @@ <h3 id="example-2">
...
+                <li>If attacking, set <var>attack</var> to <var>detector
+                average</var> minus <var>compressor gain</var>. Set
+                <var>compressor gain</var> to the multiplication of
+                <var>attack</var> by <var>envelope rate</var>. Clamp
+                <var>compressor gain</var> to a maximum of 1.0.
+                </li>
+                <li>Else, multiply <var>compressor gain</var> by <var>envelope
```
joeberkovitz: Mistake: compressor gain is not set to the product of attack and envelope rate, it is increased by that amount. Could reword as follows:

Multiply _attack_ by _envelope rate_. Add attack to _compressor gain_.

index.html
```diff
@@ -9222,6 +9225,256 @@ <h3 id="example-2">
...
+                rate</var>
+                </li>
```
joeberkovitz: The clamping to 1.0 is needed in the Else clause also.

index.html
```diff
@@ -9222,6 +9225,256 @@ <h3 id="example-2">
...
+            </li>
+            <li>Return the result of taking 0.6 power of <var>full range makeup
+            gain</var>.
+            </li>
+          </ol>
+          <p>
+            <dfn id="#envelope-rate">Computing the envelope rate</dfn> is done
+            by applying a function to the ratio of the <var>compressor
+            gain</var> and the <var>detector average</var>. User-agents are
+            allowed to choose the shape of the attack and release curve.
```
joeberkovitz: The sudden introduction of "curves" here is confusing. I think we need a sentence or two explaining that the envelope rate is a scale factor by which some gain quantity is multiplied, in each sample frame. These successive multiplications, in turn, generate the envelope curve.

index.html
```diff
@@ -9222,6 +9225,256 @@ <h3 id="example-2">
...
+            <li>The attack curve MUST be a continuous, monotonically increasing
+            function in the range \([0, 1]\).
+            </li>
+            <li>The release curve MUST be a continuous, monotonically
+            decreasing function that is always greater to 1.
+            </li>
+          </ul>
+          <div class="note">
+            It is allowed, for example, to have a compressor that performs an
+            <em>adaptative release</em>, that is, releasing faster the harder
```
joeberkovitz: adaptative -> adaptive

index.html
```diff
@@ -9222,6 +9225,256 @@ <h3 id="example-2">
...
+              <div class="note">
+                When attacking, this number less than or equal to 1, when
+                releasing, this number is strictly greater to 1.
+              </div>
+            </li>
+            <li>The attack curve MUST be a continuous, monotonically increasing
+            function in the range \([0, 1]\).
+            </li>
+            <li>The release curve MUST be a continuous, monotonically
+            decreasing function that is always greater to 1.
```
joeberkovitz: greater than 1

index.html
```diff
@@ -9222,6 +9225,256 @@ <h3 id="example-2">
...
+          <ul>
+            <li>The input to the curve MUST be the ratio of the <var>compressor
+            gain</var> and the <var>detector average</var>.
+              <div class="note">
+                When attacking, this number less than or equal to 1, when
+                releasing, this number is strictly greater to 1.
+              </div>
+            </li>
+            <li>The attack curve MUST be a continuous, monotonically increasing
+            function in the range \([0, 1]\).
```
joeberkovitz: Given the above eplanation that I am suggesting, these requirements follow directly from the constraints on the value of the envelope rate.

index.html
```diff
@@ -9222,6 +9225,256 @@ <h3 id="example-2">
...
+          </ol>
+          <p>
+            <dfn id="#envelope-rate">Computing the envelope rate</dfn> is done
+            by applying a function to the ratio of the <var>compressor
+            gain</var> and the <var>detector average</var>. User-agents are
+            allowed to choose the shape of the attack and release curve.
+            However, they MUST respect the following constraints:
+          </p>
+          <ul>
+            <li>The input to the curve MUST be the ratio of the <var>compressor
```
joeberkovitz: Rather than the "input to the curve", this is more clearly described as the envelope rate itself (which in turn determines the curve). So... "the envelope rate MUST be calculated from the ratio of..."

index.html
rtoy: If you just say continuous and piece-wise differentiable instead of continuously differentiable, everything should work out.  The hard knee is continuous and piece-wise differentiable.

index.html
```diff
@@ -9222,6 +9225,256 @@ <h3 id="example-2">
...
+              </ul>This curve MUST be continuously differentiable, and
+              corresponds to a target output level, based on the input level.
+            </li>
+          </ul>
+          <p>
+            Graphically, such a curve would look something like this:
+          </p>
+          <figure>
+            <img alt="Graphical representation of a compression curve" src=
+            "images/compression-curve.svg" style="width: 50%">
```
rtoy: In the diagram, you draw ratio as if it's the angle of the line after the knee.  It's not; it's the reciprocal of the slope.  Not exactly sure how to show that.

index.html
```diff
@@ -9222,6 +9225,256 @@ <h3 id="example-2">
...
+              var gain = new GainNode(context);
+              var compression = new EnvelopeFollower();
+
+              input.connect(delay).connect(gain).connect(output);
+              input.connect(compression).connect(gain.gain);</pre>
+          <figure>
+            <img src="images/dynamicscompressor-internal-graph.svg" alt=
+            "Schema of
+            the internal graph used by the DynamicCompressorNode">
+            <figcaption>
```
rtoy: Thanks for the diagram!  Looks really nice.  One nit:  Kind of hard to tell that the envelope follower output connects to the gain attribute of the gain node.  Not exactly sure how to best illustrate this, though.

index.html
```diff
@@ -9144,7 +9146,8 @@ <h3 id="example-2">
               A read-only decibel value for metering purposes, representing the
               current amount of gain reduction that the compressor is applying
               to the signal. If fed no signal the value will be 0 (no gain
-              reduction).
+              reduction). When this attribute is read, return the value of the
+              private slot [[\reduction]].
```
padenot: Is it not what I've done?

index.html
```diff
@@ -9222,6 +9225,256 @@ <h3 id="example-2">
...
+            gain</var> and the <var>detector average</var>. User-agents are
+            allowed to choose the shape of the attack and release curve.
+            However, they MUST respect the following constraints:
+          </p>
+          <ul>
+            <li>The input to the curve MUST be the ratio of the <var>compressor
+            gain</var> and the <var>detector average</var>.
+              <div class="note">
+                When attacking, this number less than or equal to 1, when
+                releasing, this number is strictly greater to 1.
```
rtoy: "greater to" -> "greater than"

index.html
```diff
@@ -9222,6 +9225,256 @@ <h3 id="example-2">
...
+          <div class="note">
+            It is allowed, for example, to have a compressor that performs an
+            <em>adaptative release</em>, that is, releasing faster the harder
+            the compression, or to have curves for attack and release that are
+            not of the same shape.
+          </div>
+          <p>
+            Applying a <dfn>compression curve</dfn> to a value means computing
+            the value of this sample when passed to a function, and returning
+            the computed value. This function MUST respects the following
```
rtoy: "respects" -> "respect"

index.html
```diff
@@ -9222,6 +9225,256 @@ <h3 id="example-2">
...
+          </p>
+          <ol>
+            <li>This function is the identity up to the value of the linear
+            <code>threshold</code> (i.e., \(f(x) = x\)).
+            </li>
+            <li>From the <code>threshold</code> up to the <code>threshold +
+            knee</code>, User-Agents can choose the curve shape. The whole
+            function MUST be monotonically increasing and continuous except
+            when the <code>knee</code> parameter is 0, in which case a
+            discontinuity happens at the treshold.
```
rtoy: In the diagram with a hard knee, the curve is still continuous, but not differentiable; there is no discontinuity.  I think it's enough here to say that the curve in monotonic increasing and continuous.  No exception needs to be made when the knee is 0.

index.html
```diff
@@ -9222,6 +9225,256 @@ <h3 id="example-2">
...
+            </li>
+          </ol>
+          <p>
+            Converting a value \(v\) in <dfn id="linear-to-decibel">linear gain
+            unit to decibel means</dfn> executing the following steps:
+          </p>
+          <ol>
+            <li>If <var>v</var> is equal to zero, return -1000.
+            </li>
+            <li>Else, return \( 20 \times \log_{10}{v} \)
```
rtoy: `\times` not really needed, but ok.  (Might need `\,` or something if you do remove the times, so that there's a tiny space between 20 and log.)

index.html
joeberkovitz: IIRC, what you've done is to say [[reduction]] is ignored when the signal is present, and zero is returned instead.

index.html
padenot: This is now well defined in the algorithm above.

index.html
padenot: Yes, the idea here is that it's allowed to do a normal release, or adaptive release, etc. I'll rephrase that.

index.html
padenot: Yes, this is somewhat of a language abuse. In music software, it's often represented like that, but we need to be more precise here.

index.html
padenot: I tried to put a little "Gain" label, maybe we can do something better.

index.html
```diff
@@ -9248,18 +9248,20 @@ <h3 id="example-2">
...
-            <li>The compression curve is has three parts:
+            <li>The compression curve has three parts:
               <ul>
-                <li>The first part is the identify: \(f(x) = x\).
+                <li>The first part is the identity: \(f(x) = x\).
                 </li>
                 <li>The second part is the soft-knee portion, which MUST be a
                 monotonically increasing function.
                 </li>
                 <li>The third part is a linear function of equation \(f(x) =
```
rtoy: Nit:  "linear function of equation" -> "the linear function:"

The ":" to make to match what you say in line 9253 above.

index.html
```diff
@@ -9323,8 +9325,8 @@ <h3 id="example-2">
             </li>
           </ul>
           <p>
-            Now, the actual algorithm to determine the reduction gain, in
-            linear gain value.
+            Now, the actual algorithm to determine <var>reduction gain</var>, in
+            linear gain value, for each sample of input.
```
rtoy: This isn't a complete sentence.  Maybe just replace the final period with a colon?

index.html
```diff
@@ -9350,28 +9352,36 @@ <h3 id="example-2">
...
               <ol>
                 <li>Let <var>envelope rate</var> be the result of <a href=
-                "envelope-rate">computing the envelope rate</a>
+                "#envelope-rate">computing the envelope rate</a>
                 </li>
-                <li>Let <var>attenuation</var> be the value of applying the
-                <a href="compression-curve">compression curve</a> to the
-                absolute value of <var>input</var> .
+                <li>If the absolute value of <var>input</var> is less than
+                0.0001, let <var>attenuation</var> be 1.0. Else, let
```
rtoy: Is this what Chrome and Firefox do today?

index.html
```diff
@@ -9411,34 +9421,44 @@ <h3 id="example-2">
             <dfn id="#envelope-rate">Computing the envelope rate</dfn> is done
             by applying a function to the ratio of the <var>compressor
             gain</var> and the <var>detector average</var>. User-agents are
-            allowed to choose the shape of the attack and release curve.
+            allowed to choose the shape the enveloppe function.
```
rtoy: "enveloppe" -> "envelope", here and elsewhere. (Is this the French spelling?)

Mr21: Exactly :fr:

index.html
```diff
@@ -9085,7 +9085,9 @@ <h3 id="example-2">
             <p>
               Let <var>node</var> be a new <a>DynamicsCompressorNode</a>
               object. <a href="#audionode-constructor-init">Initialize</a>
-              <var>node</var>, and return <var>node</var>.
+              <var>node</var>. Let [[\reduction]] be a private slot on this
```
joeberkovitz: It looks as though you changed the name of the slot everywhere except here.

index.html
```diff
@@ -9222,6 +9225,274 @@ <h3 id="example-2">
...
+                </li>
+                <li>If the absolute value of <var>input</var> is less than
+                0.0001, let <var>attenuation</var> be 1.0. Else, let
+                <var>attenuation</var> be the value of applying the <a
+                href="#compression-curve">compression curve</a> to the absolute
+                value of <var>input</var>, divided by the absolute value of
+                <var>input</var>.
+                </li>
+                <li>Let <var>decibels per frame</var> be the result of dividing
+                <var>attenuationDB</var> by <var>releaseRate</var>.</li>
```
joeberkovitz: _attenuationDB_ does not appear to be defined anywhere

joeberkovitz: I think you meant to divide attenuation by the quantity known in the C++ code as `sat_release_frames`, which is the number of frames in a release time interval. Note that `releaseRate` is not computed until the following step (and it's computed from the work done in this step!)

index.html
```diff
@@ -9222,6 +9225,274 @@ <h3 id="example-2">
...
+                <li>Set <var>relaseRate</var> be the result of
```
joeberkovitz: relaseRate -> releaseRate

index.html
```diff
@@ -9222,6 +9225,274 @@ <h3 id="example-2">
...
+                <a href="#db-to-linear">converting <var>decibels per frames
+                </var>to linear gain</a>, and substract one.</li>
+                <li>Set <var>change-rate</var> to the difference between <var>
+                  attenuation</var> and <var>detector average</var>.
```
joeberkovitz: This where _releaseRate_ is supposed to be multiplied by the difference: the new change rate is (attenuation - detector average) * release rate.

joeberkovitz: That is, unless we're attacking.

index.html
```diff
@@ -9222,6 +9225,274 @@ <h3 id="example-2">
...
+                </li>
+                <li>Apply the <a href="#detector-curve">detector curve</a> to
+                <var>change-rate</var></li>
+                <li>Add <var>change-rate</var> to <var>detector average</var>,
+                and clamp <var>detector average</var> to a maximum of 1.0.  </li>
+                <li>If attacking, set <var>attack</var> to <var>detector
+                average</var> minus <var>compressor gain</var>. Set
+                <var>compressor gain</var> to the multiplication of
+                <var>attack</var> by <var>envelope rate</var>. Clamp
```
joeberkovitz: The sentence "Set compressor gain to the multiplication of..." appears to be left over from before. It should be replaced by the new language in line 9382 about multiplying attack by envelope rate and adding attack to compressor gain.

index.html
padenot: Yes

padenot: I think it's good enough to avoid the division by zero.

index.html
```diff
@@ -9222,6 +9225,274 @@ <h3 id="example-2">
...
+            (as a <a>k-rate</a> parameter).
+            </li>
+            <li>Let <var>attack</var> and <var>release</var> have the value of
+            the <a>AudioParam</a> of the same name, sampled at the time of
+            processing (those are <a>k-rate</a> parameters), mutiplied by the
+            sample-rate of the <a>BaseAudioContext</a> this
+            <a>DynamicsCompressorNode</a> is <a href=
+            "#associated">associated</a> with.
+            </li>
+            <li>For each samples <var>input</var> of <var>input signal</var>,
```
rtoy: Nit: "each samples" -> "each sample"

index.html
```diff
@@ -9222,6 +9225,274 @@ <h3 id="example-2">
...
+            execute the following steps:
+              <ol>
+                <li>Let <var>envelope rate</var> be the result of <a href=
+                "#envelope-rate">computing the envelope rate</a>
```
rtoy: Nit: End with a period, like you do elsewhere.

index.html
```diff
@@ -9222,6 +9225,274 @@ <h3 id="example-2">
...
+                </li>
+                <li>If the absolute value of <var>input</var> is less than
+                0.0001, let <var>attenuation</var> be 1.0. Else, let
+                <var>attenuation</var> be the value of applying the <a
+                href="#compression-curve">compression curve</a> to the absolute
+                value of <var>input</var>, divided by the absolute value of
```
rtoy: This phrasing is a bit confusing because I interpreted this as dividing the absolute value of the input by the absolute value of the input.  A closer reading shows that you really mean compression curve value divided by input absolute value.  Maybe rephrase or use two sentences?

index.html
```diff
@@ -9222,6 +9225,274 @@ <h3 id="example-2">
...
+            </li>
+            <li>Return the result of taking 0.6 power of <var>full range makeup
+            gain</var>.
+            </li>
+          </ol>
+          <p>
+            <dfn id="#envelope-rate">Computing the envelope rate</dfn> is done
+            by applying a function to the ratio of the <var>compressor
+            gain</var> and the <var>detector average</var>. User-agents are
+            allowed to choose the shape the enveloppe function.
```
rtoy: "envelope"

index.html
```diff
@@ -9222,6 +9225,274 @@ <h3 id="example-2">
...
+            However, they MUST respect the following constraints:
+          </p>
+          <ul>
+            <li>The enveloppe rate MUST be the calculated from the ratio of the
```
rtoy: "envelope"

Also are you referring to the "envelope rate" variable?

index.html
```diff
@@ -9222,6 +9225,274 @@ <h3 id="example-2">
...
+            </li>
+            <li>From the <code>threshold</code> up to the <code>threshold +
+            knee</code>, User-Agents can choose the curve shape. The whole
+            function MUST be monotonically increasing and continuous.
+              <div class="note">
+                If the <code>knee</code> is 0, the
+                <a>DynamicsCompressorNode</a> is called a hard-knee compressor.
+              </div>
+            </li>
+            <li>This function returned is linear, based on the ratio, after the
```
rtoy: "This function returned is linear" -> "This function is linear"

index.html
padenot: I've used two sentences and an intermediate variable so make this extra-clear.

index.html
padenot: I've clarified this.

index.html
padenot: I've tried something else, have a look.

padenot:
The last few commits refactor the algorithm so that things are in the right order, change the diagram, address the other comments, and tidy.

PR #1281, opened 2017-08-02
Create AudioBuffer only when offline rendering begins.

### LINKED ISSUE #1277: startRendering behavior changed
When adding support for the construction of an `OfflineAudioContext` with a dictionary, the behavior of `startRendering()` changed somewhat.  Previously, it said http://rawgit.com/WebAudio/web-audio-api/d99de23/index.html#widl-OfflineAudioContext-startRendering-Promise-AudioBuffer.  In the second list, this says that the `AudioBuffer` is called when `startRendering()` is called.  In the current version of the spec, the buffer is created when `OfflineAudioContext` is constructed.

I think it actually makes sense to create the `AudioBuffer` when `startRendering() is called. Thus, we should put back the old behavior.

Note also that the old behavior isn't well-defined. It's possible the `OfflineAudioContext` is called with valid values for the number of channels, length, and sample rate, but creating the `AudioBuffer` when `startRendering()` is called may fail. (Most likely out of memory).  We don't specify what should happen in this case.  Presumably `startRendering()` will reject the promise with some error code.

joeberkovitz:
Fixes #1277

### DIFF
--- a/index.html
+++ b/index.html
@@ -2249,26 +2249,29 @@ function setupRoutingGraph() {
               promise with <code>InvalidStateError</code>, and abort these
               steps.
               </li>
+              <li>Let <var>promise</var> be a new promise.
+              </li>
               <li>Create a new <code>AudioBuffer</code>, with a number of
               channels, length and sample rate equal respectively to the <code>
                 numberOfChannels</code>, <code>length</code> and
                 <code>sampleRate</code> values passed to this instance's
                 constructor in the <code>contextOptions</code> parameter.
-                Assign this buffer to <a><var>[[rendered buffer]]</var></a>. An
-                exception may be thrown during the constructor call, causing
-                the remaining steps in this process to be skipped.
+                Assign this buffer to <a><var>[[rendered buffer]]</var></a>.
               </li>
-              <li>Let <var>promise</var> be a new promise.
+              <li>If an exception was thrown during the preceding
+              <code>AudioBuffer</code> constructor call, reject
+              <var>promise</var> with this exception.
               </li>
-              <li>Start to render the audio graph on another thread.
+              <li>Otherwise, in the case that the buffer was successfully
+              constructed, <a>begin offline rendering</a>.
               </li>
               <li>Return <var>promise</var>.
               </li>
             </ol>
             <p>
-              When rendering an audio graph on another thread, the following
-              steps MUST happen on a <a>rendering thread</a> that is created
-              for the occasion.
+              To <dfn>begin offline rendering</dfn>, the following steps MUST
+              happen on a <a>rendering thread</a> that is created for the
+              occasion.
             </p>
             <ol>
               <li>Given the current connections and scheduled changes, start
--- a/index.html
+++ b/index.html
@@ -2150,14 +2150,6 @@ function setupRoutingGraph() {
               <a>channelCount</a> set to
               <code>contextOptions.numberOfChannels</code>.
               </li>
-              <li>
-                <p>
-                  Let <dfn>[[rendered buffer]]</dfn> be an internal slot
-                  initialized to <code>undefined</code>. When rendering begins,
-                  this slot will hold an <a>AudioBuffer</a> containing rendered
-                  audio data for the offline context.
-                </p>
-              </li>
             </ol>
             <dl class="parameters">
               <dt>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2153,18 +2153,9 @@ <h2 id="OfflineAudioContext">
...
-    length: contextOptions.length,
-    sampleRate: contextOptions.sampleRate
-})
-                </pre>
-                <p>
-                  This buffer holds the rendered audio data for the offline
-                  context.
+                  initialized to <code>undefined</code>. When rendering begins,
+                  this slot will hold an <a>AudioBuffer</a> containing rendered
+                  audio data for the offline context.
```
rtoy: Do we really need this at all?  The construction of the context doesn't need to say anything because you don't need one until `startRendering()` is called.

joeberkovitz: I'm not sure (since I didn't add the text regarding the slot), but I imagined that the reason it exists is to supply the buffer on the control thread side when the Promise succeeds. So I just left it as-is.  Let's discuss this today on the call but I don't have any particular leaning; happy to do whatever works best.

joeberkovitz:
Change the AudioBuffer construction language to just allocate the buffer for storage, and indicate the AudioBuffer is constructed in the control thread when the promise is resolved.

index.html
padenot: Let's remove all the steps here.

index.html
```diff
@@ -2271,6 +2262,15 @@ <h2 id="OfflineAudioContext">
               for the occasion.
             </p>
             <ol>
+              <li>Create a new <code>AudioBuffer</code>, with a number of
```
padenot: Initialize a slot here, and the store the `AudioBuffer` in the slot.

We need to add language to the `AudioBuffer` ctor to allow throwing OOM errors and returning undefined or null or something.

joeberkovitz:
@padenot @rtoy PTAL. I moved the construction of the buffer into the control thread so that OOM exceptions could be handled synchronously up front to a Promise being created. Note that https://tc39.github.io/ecma262/#sec-createbytedatablock behavior already implies that the AudioBuffer ctor will throw such exceptions.

index.html
```diff
@@ -2258,6 +2249,15 @@ <h2 id="OfflineAudioContext">
...
               </li>
+              <li>Create a new <code>AudioBuffer</code>, with a number of
+              channels, length and sample rate equal respectively to the <code>
+                numberOfChannels</code>, <code>length</code> and
+                <code>sampleRate</code> values passed to this instance's
+                constructor in the <code>contextOptions</code> parameter.
+                Assign this buffer to <a><var>[[rendered buffer]]</var></a>. An
+                exception may be thrown during the constructor call, causing
+                the remaining steps in this process to be skipped.
+              </li>
```
rtoy: Shouldn't this say that startRendering will reject the promise?

I think this also means we should create the buffer after we've created a new promise so we have a promise to reject.

joeberkovitz: In this instance, `startRendering()` throws an exception without even returning a Promise, which I think is appropriate. Other methods that return Promises can throw immediately, for instance if you pass invalid args to them...

rtoy: As a rather naive JS user, my expectation was that since `startRendering()` returns a promise, any errors or issues that occur should signal such things by rejecting the promise.

Passing invalid args to methods returning promises is different.  In essence, the method never actually had a chance to get started.  The error comes from parsing the args, the the guts of the method.

hoch: Then it takes more work to handle three cases:

```js
try {
  oac.startRendering().then(resolutionHandler, rejectionHandler);
} catch (e) {
  errorHandler(e);
}
```

I believe it would be easier if all kinds of rejection/error can be handled by a single handler callback.

joeberkovitz: OK, I'll have it return a rejected Promise in this case.

joeberkovitz:
@rtoy @hoch PTAL again, now rejects Promise...

hoch:
Non-reviewer LGTM.

index.html
rtoy: Yeah, I don't think we need to say anything about [[rendered buffer]] anymore. So, remove all of step 4.

But this would require updating `startRendering` to define the slot.

joeberkovitz:
@rtoy please see latest changes restricting internal slot to startRendering()

PR #1285, opened 2017-08-09
Fix #1162: MediaStream to continue when context is closed

### LINKED ISSUE #1162: What happens to the `HTMLMediaElement`s or `MediaStream`s when the context is closed
I think it should be the same as with `suspend`, i.e., nothing, everything continues as usual.

rtoy:
When a context is closed any MediaStreams or HTMLMediaElements are
effectively disconnected from the context and continue as if they were
never connected.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1970,6 +1970,13 @@ function setupRoutingGraph() {
                 </ol>
               </li>
             </ol>
+            <p>
+              When an <a>AudioContext</a> is closed, any
+              <code>MediaStream</code>s and <code>HTMLMediaElement</code>s that
+              were connected to an <a>AudioContext</a> are effectively
+              disconnected from the <a>AudioContext</a>. These will continue to
+              behave as if they were never connected to an <a>AudioContext</a>.
+            </p>
             <p class="note">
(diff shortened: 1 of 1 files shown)


rtoy:
PTAL. Not sure about the most effective way to state what we want.

padenot:
This means that if `AudioContext.close` is called, the `HTMLMediaElement` will start outputing audio to the speakers again, if it was captured with `AudioContext.createMediaElementSource`. The alternative is to have the frames dropped for ever.

If authors want more flexibility, they can use `HTMLMediaelement.captureStream` instead. This allows controlling everything manually.

Also it's a bit unclear what connect and disconnect mean, but I think agreeing on the behaviour is more important.

fwiw, <https://w3c.github.io/mediacapture-fromelement/#dom-htmlmediaelement-capturestream()> has language elements we could have a look at when drafting the spec for `createMediaElementSource`.

WofWca:
Not an expert, but tbh this makes `AudioContext.createMediaElementSource()` look straight up inferior to `HTMLMediaElement.captureStream()`.

PR #1306, opened 2017-09-01
Fix #833: use contiguous idl for audiocontext

### LINKED ISSUE #833: Respec warning about using Contiguous IDL
The spec now produces the warning

> Defining WebIDL in `dl` elements is deprecated. Please use Contiguous IDL instead: https://www.w3.org/respec/guide.html#contiguous-idl

We should probably fix this.  Is this also the cause of our mathjax displayed formulas no longer getting rendered as math? (See #831).

rtoy:
DO NOT MERGE to gh-pages branch!!!!


index.html
```diff
@@ -1635,9 +2193,625 @@ <h2 id="AudioContext">
...
+    MediaStreamAudioSourceNode      createMediaStreamSource (MediaStream mediaStream);
+    MediaStreamTrackAudioSourceNode createMediaStreamTrackSource (MediaStreamTrack mediaStreamTrack);
+    MediaStreamAudioDestinationNode createMediaStreamDestination ();
+};              
+        </pre>
+        <section>
+          <h2>
+            Constructors
+          </h2>
+          <dl class="methods" data-dfn-for="AudioContext" data-link-for=
```
rtoy: @marcoscaceres What's the proper way to define constructors? I just used class="methods", but I think that might be wrong. Maybe data-dfn-for and data-link-for are also wrong.

marcoscaceres: Unfortunately, I've not linked up constructors properly yet in ReSpec. This is what we do in `PaymentRequest` API. If you do the same, it will auto link correctly once I fix that:

```HTML
        <p>
          The <code><dfn data-lt=
          "PaymentRequest.PaymentRequest()">PaymentRequest(<var>methodData</var>,
          <var>details</var>, <var>options</var>)</dfn></code> constructor MUST
          act as follows:
        </p>
```

The key part is that `PaymentRequest.PaymentRequest()` is the actual constructor, because it's just a function.

rtoy: Yeah, I knew constructors weren't linked.  And from the PaymentRequest, I see that constructors aren't defined using any special `<dl class="foo">` construct, so perhaps I should just rip this out.

You're not going to add something like `<dl class="constructor">` or something to match the attribute and methods dl classes?

rtoy:
Closing.  This branch has been manually merged to the 833-use-contiguous-idl branch.

PR #1331, opened 2017-09-20
Fix #1328: Remove local copies of respec-w3c-common

### LINKED ISSUE #1328: Remove the local copies of respec-w3c-common
Now that #833 is fixed, we don't need to keep our own copy of the respec code.  Remove them.

rtoy:
Remove these files; they're not needed anymore after the conversion to
contiguous idl because we can use the respec version directly from the
w3c servers.


padenot:
Hrm this is very useful when working in planes.

rtoy:
Do you want to grab another more recent copy, then? I'm ok if you want to add it back in.

PR #1338, opened 2017-09-20
Clarify behavior of AudioParam.value setter w/r/t exceptions.

### LINKED ISSUE #1296: AudioParam value setter vs setValueAtTime
In https://webaudio.github.io/web-audio-api/#widl-AudioParam-value, we have

> Setting this attribute has the effect of assigning the requested value to the [[current value]] slot, and calling setValueAtTime() with the current AudioContext's currentTime and [[current value]]

What does "effect of...calling setValueAtTime()" imply here?  If this means that we do call `setValueAtTime`, do we also throw errors whenever `setValueAtTime` would throw?  In particular, if we try to set the value with the setter during the time where a `setValueCurveAtTime` is running, does the value setter throw an error?

If we do throw an error, this would be incompatible with how it used to work. (Chrome hasn't implemented the value setter semantics yet.)

joeberkovitz:
Fixes #1296

***
[Preview](https://s3.amazonaws.com/pr-preview/WebAudio/web-audio-api/1296-audioparam-setter-behavior.html) | [Diff](https://s3.amazonaws.com/pr-preview/WebAudio/web-audio-api/47294ae...eeea7e7.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -5004,13 +5004,12 @@ interface AudioDestinationNode : AudioNode {
           </li>
         </ul>
         <p class="note">
-          <code>AudioParam</code>s are read only and must be set using the
-          <a>automation methods</a> or the <a href=
-          "#widl-AudioParam-value">value</a> attribute. Using the <a href=
-          "#widl-AudioParam-value">value</a> attribute is equivalent to calling
-          <code>setValueAtTime()</code> with the current
-          <code>AudioContext</code>'s <code>currentTime</code> and the
-          requested value.
+          <code>AudioParam</code> attributes are read only, with the exception
+          of the <a href="#dom-audioparam-value">value</a> attribute. Setting
+          the <a href="#dom-audioparam-value">value</a> attribute is
+          equivalent to calling the <a href="#dfn-automation-method">automation
+          method</a> <a href=
+          "#dom-audioparam-setvalueattime">setValueAtTime()</a>.
         </p>
         <p>
           Each <a>AudioParam</a> has an internal slot <var>[[current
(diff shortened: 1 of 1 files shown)


joeberkovitz:
Note that I removed some details from the up-front "Note" paragraph, since they felt redundant with the more detailed documentation for `value`.

index.html
```diff
@@ -5004,13 +5004,12 @@ <h2 id="AudioParam">
...
-          <code>AudioParam</code>s are read only and must be set using the
-          <a>automation methods</a> or the <a href=
-          "#widl-AudioParam-value">value</a> attribute. Using the <a href=
-          "#widl-AudioParam-value">value</a> attribute is equivalent to calling
-          <code>setValueAtTime()</code> with the current
-          <code>AudioContext</code>'s <code>currentTime</code> and the
-          requested value.
+          <code>AudioParam</code> attributes are read only, with the exception
+          of the <a href="#dom-audioparam-value">value</a> attribute. Setting
+          the <a href="#dom-audioparam-value">value</a> attribute is
```
rtoy: Maybe leave out the "Setting the value..." sentence? It's not really a note any more in light of the more explicit explanation below.

index.html
```diff
@@ -5082,9 +5081,13 @@ <h2 id="AudioParam">
               <p>
                 Setting this attribute has the effect of assigning the
                 requested value to the <var>[[current value]]</var> slot, and
-                calling <code>setValueAtTime()</code> with the current
-                <code>AudioContext</code>'s <code>currentTime</code> and
-                <var>[[current value]]</var>.
+                calling the <a href=
+                "#dom-audioparam-setvalueattime">setValueAtTime()</a> method
```
rtoy: Nit: I think we should avoid this `href="#dom-audioparam..."` stuff. With contiguous idl, I think the right way is `<a data-link-for="AudioParam">setValueAtTime</a>()`.  (Not sure if the parens should be inside our outside that a tag.

index.html
joeberkovitz: Makes sense to me - I'll remove it. This duplication already existed so wasn't sure what to do.

index.html
joeberkovitz: cool, I didn't know the right way.

joeberkovitz:
@rtoy please review again

PR #1339, opened 2017-09-20
Have ABSN acquire its buffer content when the buffer is set and start() called.

### LINKED ISSUE #1315: Acquiring the contents in AudioBufferSourceNode start?
https://webaudio.github.io/web-audio-api/#acquire-the-content describes how to acquire the contents of the AudioBuffer when `AudioBufferSourceNode.start()` is called.

However, it's perfectly valid to do, say, `start(currentTime + 100)` and then sometime later set the `buffer` attribute of the source node.

Should this be updated to say that the contents are also acquired when the buffer is set if `start()` has already been called but hasn't actually stopped? Are there other corner cases to consider?

joeberkovitz:
Fix #1315

***
[Preview](https://s3.amazonaws.com/pr-preview/WebAudio/web-audio-api/1315-acquire-when-starting-absn.html) | [Diff](https://s3.amazonaws.com/pr-preview/WebAudio/web-audio-api/47294ae...83b6e90.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -7145,6 +7145,12 @@ interface AudioBuffer {
           node's <code>buffer</code>. If the operation fails, nothing is
           played.
           </li>
+          <li>When the <code>buffer</code> of an <a>AudioBufferSourceNode</a>
+          is set and <code>AudioBufferSourceNode.start</code> has been
+          previously called, the setter <a href="#acquire-the-content">acquires
+          the content</a> of the <a>AudioBuffer</a>. If the operation fails,
+          nothing is played.
+          </li>
           <li>When a <a>ConvolverNode</a>'s <code>buffer</code> is set to an
           <a>AudioBuffer</a> while the node is connected to an output node, or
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -7958,7 +7969,7 @@ <h3 id="playback-AudioBufferSourceNode">
 
   // Determine loop endpoints as applicable
   let actualLoopStart, actualLoopEnd;
-  if (loop) {
+  if (loop &amp;& buffer != null) {
```
rtoy: Isn't it just `&amp;` to get `&` instead of `&amp;&`? If so, I guess line 7373 below needs to be fixed too.

Mr21: True, you will need `&amp;&amp;` to get `&&`.

joeberkovitz: I used `&amp;&amp;` (give me some credit :-) but html5tidy "corrects" this to `&amp;&`. So checks fail if the correct construct is used.

PR #1340, opened 2017-09-21
Fix broken “#widl-“ links

### LINKED ISSUE #1335: Fix/replace old widl-foo links
With the conversion to contiguous idl, many of the old links of the form `<a href="#widl-...">foo</a>` may be incorrect.  At the very least they should be replaced by something like `<a data-link-for="bar">foo</a>`

joeberkovitz:
Fixes #1335

***
[Preview](https://s3.amazonaws.com/pr-preview/WebAudio/web-audio-api/1335-fix-widl-links.html) | [Diff](https://s3.amazonaws.com/pr-preview/WebAudio/web-audio-api/7fec459...e969d97.html)


index.html
```diff
@@ -879,26 +878,26 @@ <h2 id="BaseAudioContext">
...
               </p>
               <p>
-                In the time coordinate system of <a href=
-                "#widl-BaseAudioContext-currentTime"><code>currentTime</code></a>,
-                the value of zero corresponds to the first sample-frame in the
-                first block processed by the graph. Elapsed time in this system
-                corresponds to elapsed time in the audio stream generated by
-                the <a><code>BaseAudioContext</code></a>, which may not be
+                In the time coordinate system of <a data-link-for=
+                "BaseAudioContext"><code>currentTime</code></a>, the value of
```
rtoy: I think that `<code>` should be outside the a tag. Or might be possibly dropped completely?

But I see we do a lot of that elsewhere, so maybe ignore this.

index.html
```diff
@@ -1496,9 +1495,9 @@ <h2 id="BaseAudioContext">
                     <span role="img" aria-label="False">✘</span>
                   </td>
                   <td class="prmDesc">
-                    A sequence of cosine parameters. See its <a href=
-                    "#widl-PeriodicWaveOptions-real">constructor
-                    counterpart</a> for a more detailed description.
+                    A sequence of cosine parameters. See its <a data-link-for=
+                    "PeriodicWaveOptions">constructor counterpart</a> for a
```
rtoy: This link doesn't work.

I'm ok if you want to fix this broken link in a later PR.

index.html
```diff
@@ -1515,9 +1514,9 @@ <h2 id="BaseAudioContext">
                     <span role="img" aria-label="False">✘</span>
                   </td>
                   <td class="prmDesc">
-                    A sequence of sine parameters. See its <a href=
-                    "#widl-PeriodicWaveOptions-imag">constructor
-                    counterpart</a> for a more detailed description.
+                    A sequence of sine parameters. See its <a data-link-for=
+                    "PeriodicWaveOptions">constructor counterpart</a> for a
```
rtoy: Ditto

rtoy:
Latest validation result:  https://validator.w3.org/nu/?doc=http%3A%2F%2Frawgit.com%2FWebAudio%2Fweb-audio-api%2Febccd9a%2Findex.html

There are only 4 warnings now.  I wonder why it doesn't catch the periodic wave constructor thing?  That makes me wonder what other links aren't correctly linked. :-(

joeberkovitz:
I have manually reviewed each of the changed links and found the errors you mentioned, plus a couple others.  All of these cases were cases in which the original link text was somehow incorrect. So I think we're OK now -- take another look if you like.

PR #1347, opened 2017-09-22
Use [Exposed].

### LINKED ISSUE #1297: Use [Exposed]
Please annotate all interface with [Exposed] so it's clear on which global they can be found. I suspect most would just be Window.

Needed for https://github.com/heycam/webidl/issues/365.

padenot:
This fixes #1297.

***
[Preview](https://s3.amazonaws.com/pr-preview/padenot/web-audio-api/1297-exposed.html) | [Diff](https://s3.amazonaws.com/pr-preview/WebAudio/web-audio-api/7fec459...padenot:e8dfd00.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -828,6 +828,7 @@ callback DecodeErrorCallback = void (DOMException error);
 
 callback DecodeSuccessCallback = void (AudioBuffer decodedData);
 
+[Exposed=Window]
 interface BaseAudioContext : EventTarget {
     readonly        attribute AudioDestinationNode destination;
     readonly        attribute float                sampleRate;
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2172,8 +2174,9 @@ <h2 id="AudioContext">
           </tbody>
         </table>
         <pre class="idl">
-[Constructor,
- Constructor(optional AudioContextOptions contextOptions)]
+[Exposed=Window,
+ Constructor,
```
rtoy: Not for this PR, but should we get rid of this Constructor? Since the next constructor has a optional dictionary, aren't they equivalent?

padenot: Yes.

PR #1358, opened 2017-09-25
Fix #1304: add node property tables

### LINKED ISSUE #1304: Describe node properties neatly
Consider, for example, http://localhost:7100/#the-pannernode-interface.  We have a list of default values for numberOfInputs/Outputs, and channel stuff.

Wouldn't it be nicer if we just used a table?  Say,

Node Parameter | Value | Notes
-- | - | -
numberOfInputs | 1 |
numberOfOutputs | 1 |
channelCount | 2 | Has channelCountConstraints
channeCountMode | "clamped-max" | Has constraints
channelInterpretation | "speakers" | 
tail-time reference | Yes if HRTF, otherwise no.|

Well, maybe we don't want tail-time reference in this table.

rtoy:
Adds a table for each node to specify the node parameters like number of inputs/outputs, channel count stuff, and time-time reference.

(Maybe the heading of "Property" is not right?)

I think this makes that intro section much easier to find, and read, and it looks much better than the original information stuffed in a `<pre>` section.

The table properties are kind of random and could be refined, but this is a good start.

***
[Preview](https://s3.amazonaws.com/pr-preview/rtoy/web-audio-api/1304-add-node-property-tables.html) | [Diff](https://s3.amazonaws.com/pr-preview/WebAudio/web-audio-api/304c606...rtoy:2896cb2.html)


rtoy:
PTAL.  This PR basically adds a table for every node that replaces the `<pre>` section that contained the `numberOfInputs`, `numberOfOutputs`, and channel information with the same information in table form.

Also add a entry for a tail-time reference with a note on why.

Every node also  explicitly says the count and mode, even though some nodes didn't have that previously.  Did this just for completeness, even though there are defaults specified by `AudioNode`.

Finally, for source nodes, I left out the table entries for the channel stuff since they're not really relevant, even though every node has a value for them.

mdjp:
This one looks good to me, makes it much clearer.

rtoy:
Thanks @mdjp !

PR #1359, opened 2017-09-25
Fix #1299: add audio param tables

### LINKED ISSUE #1299: Better exposure of a-rate/k-rate parameters and nominal range
Currently, we just have random text describing whether an AudioParam is a-rate or k-rate and what the nominal range is for the parameter.

Wouldn't it be nice if we had a table that described this?  Then all entries would look the same and it would make it easier to find this information for each AudioParam

rtoy:
For each AudioParam, add a simple table giving the values that the AudioParam object would have by default.  Also mention whether the param is a-rate or k-rate.  While not necessary, I think it makes it easier to see what the AudioParam values should be and how the nominal range is really the min and max values of the parameter.

And this makes the description of the AudioParam consistent everywhere.

***
[Preview](https://s3.amazonaws.com/pr-preview/rtoy/web-audio-api/1299-add-audio-param-tables.html) | [Diff](https://s3.amazonaws.com/pr-preview/WebAudio/web-audio-api/23c9dcc...rtoy:2a372d9.html)


rtoy:
PTAL. Don't think it's controversial; it just formalizes the default, min, and max values for each AudioParam in a readable table, and also adds a note on whether the parameter is a-rate or k-rate.  Removes a bunch of text and makes it all uniform.

mdjp:
2.17.1 Attributes 

These all still include the old description

eg This parameter is a-rate when used with a PannerNode that has a panningModel set to "equalpower" , k-rate otherwise. Its nominal range is (−∞,∞). 

Same with 
2.23.2 Attributes (release)
2.24.2 Gain

I'm assuming this is unintentional duplication

rtoy:
Thanks for catching those errors. I thought I had fixed them all. :-(

mdjp:
Looks good to me now @rtoy

rtoy:
Thanks for the quick review @mdjp

padenot:
Changing to -FLT_MIN and FLT_MIN is a breaking change. It would. have been good to open an issue about it. I'm not opposed to it though, I think it makes more sense.

rtoy:
Oops.  Sorry about that. I think allowing infinity for the limit is a bad idea because it probably very quickly causes audio to be infinity or, more likely, NaN.  Of course, FLT_MAX will also probably quickly produce that too, but that's a different issue.

PR #1364, opened 2017-09-26
Fix #1361: Clarify nominal range for simple parameters

### LINKED ISSUE #1361: Clarify nominal range definition
In https://webaudio.github.io/web-audio-api/#dfn-nominal-range we have:

> The nominal range for a computedValue are the lower and higher values this parameter can effectively have.

It would be helpful if also mentioned that these lower and higher values for simple parameters are the minValue and maxValue attributes of the AudioParam.  Only in passing in the description of the minValue and maxValue of the AudioParam attributes do we mention this connection to the nominal range.

rtoy:
Basically define a new term `simple nominal range` that says the
AudioParam `minValue` and `maxValue` form the valid range, and use
that in the description of the full `nominal range` term that includes
both simple and compound parameters.

***
[Preview](https://s3.amazonaws.com/pr-preview/rtoy/web-audio-api/1361-clarify-nominal-range-with-min-max.html) | [Diff](https://s3.amazonaws.com/pr-preview/WebAudio/web-audio-api/b7ca241...rtoy:5fde671.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -4945,8 +4945,8 @@ interface AudioDestinationNode : AudioNode {
           Each <code>AudioParam</code> includes <a data-link-for=
           "AudioParam"><code>minValue</code></a> and <a data-link-for=
           "AudioParam"><code>maxValue</code></a> attributes that together form
-          the <a>nominal range</a> for the parameter. In effect, value of the
-          parameter is clamped to the range \([\mathrm{minValue},
+          the <dfn>simple nominal range</dfn> for the parameter. In effect,
+          value of the parameter is clamped to the range \([\mathrm{minValue},
           \mathrm{maxValue}]\). See the section <a href=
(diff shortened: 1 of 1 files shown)


rtoy:
Not sure if `simple nominal range` is the best term, but it is short and pretty descriptive.

rtoy:
Right now, nominal range is circular.  What this is trying to do is add a 'simple nominal range' item that says for a simple parameter, the nominal range is the [`minValue`, `maxValue`] from the attributes of the AudioParam.  See https://s3.amazonaws.com/pr-preview/rtoy/web-audio-api/1361-clarify-nominal-range-with-min-max.html#dfn-simple-nominal-range.  Without this, you have to somehow read the paragraph and remember it when reading the definition of nominal range.

PR #1386, opened 2017-09-28
AudioWorkletNodeOptions.outputChannelCount

### LINKED ISSUE #1325: [audioworklet] Number of channels in output buffers
Consider the following:

```js
// In AudioWorkletProcessor class definition.
process(inputs, outputs, parameters) {
  for (let i = 0; i < inputs.length; ++i) {
    let inputBuffer = inputs[i];
    let inputChannelCount = inputBuffer.channel;
  }
  for (let o = 0; o < outputs.length; ++o) {
    let outputBuffer = outputs[o];
    let outputChannelCount = outputBuffer.channel;
  }
}
```

In the example above, what you get from `inputChannelCount` is determined by the channel count computing algorithm. ([spec](https://webaudio.github.io/web-audio-api/#channel-up-mixing-and-down-mixing)) However, what you get for `outputChannelCount` is not well-defined.

The simplest approach is making the matching input/output buffers have the same channel count. For example, if `inputChannelCount` gives you _n_, you can expected `outputChannelCount` to be _n_.

Although there are more elegant solutions for this, I am against creating AudioBuffer objects (filling in `outputs` every render quantum) in the `process` method because that will generate garbages fast and furious.

hoch:
- Fixes #1325.

***
[Preview](https://s3.amazonaws.com/pr-preview/hoch/web-audio-api/1325-output-channel-count.html) | [Diff](https://s3.amazonaws.com/pr-preview/WebAudio/web-audio-api/8d6d807...hoch:9215fc9.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -9109,6 +9109,7 @@ interface AudioWorkletNode : AudioNode {
 dictionary AudioWorkletNodeOptions : AudioNodeOptions {
              unsigned long             numberOfInputs = 1;
              unsigned long             numberOfOutputs = 1;
+             sequence&lt;unsigned long&gt;    outputChannelCount;
              record&lt;DOMString, double&gt;  parameterData;
 };
             </pre>
@@ -9137,6 +9138,20 @@ dictionary AudioWorkletNodeOptions : AudioNodeOptions {
                   <a data-link-for="AudioNode">numberOfOutputs</a> attribute.
                 </dd>
                 <dt>
+                  <code><dfn>outputChannelCount</dfn></code> of type
+                  <span class="idlAttrType"><code>sequence&lt;unsigned
+                  long&gt;</code></span>
+                </dt>
+                <dd>
+                  This is used to configure the number of channels in each
+                  output. <code>IndexSizeError</code> MUST be thrown if the
+                  length of sequence does not match <a data-link-for=
+                  "AudioWorkletNodeOptions">numberOfOutputs</a>. A
+                  <code>NotSupportedError</code> exception MUST be thrown if a
+                  channel count is not in the valid range of <a data-link-for=
+                  "AudioNode">channelCount</a>.
+                </dd>
+                <dt>
                   <code><dfn>parameterData</dfn></code> of type <span class=
                   "idlAttrType"><code>record&lt;DOMString,
                   double&gt;</code></span>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -9150,6 +9165,51 @@ <h2 id="AudioWorkletNodeOptions">
...
+              <h3>
+                Configuring Channels with <a>AudioWorkletNodeOptions</a>
+              </h3>
+              <p>
+                With a combination of <a data-link-for=
+                "AudioWorkletNodeOptions">numberOfInputs</a>, <a data-link-for=
+                "AudioWorkletNodeOptions">numberOfOutputs</a> and
+                <a data-link-for=
+                "AudioWorkletNodeOptions">outputChannelCount</a>, various
+                channel configurations can be achieved.
```
rtoy: Can we use a table somehow?

index.html
```diff
@@ -9267,7 +9327,8 @@ <h2 id="defining-a-valid-audioworkletprocessor">
                   <code>Float32Array</code> of audio samples for the
                   <code>m</code>th channel of <code>n</code>th input. While the
                   number of inputs is fixed at the construction, the number of
-                  channels can be changed dynamically.
+                  channels can be changed dynamically based on
+                  <a>computedNumberOfChannels</a>
```
rtoy: Need period to end the sentence.

index.html
```diff
@@ -9267,7 +9327,8 @@ <h2 id="defining-a-valid-audioworkletprocessor">
                   <code>Float32Array</code> of audio samples for the
                   <code>m</code>th channel of <code>n</code>th input. While the
                   number of inputs is fixed at the construction, the number of
```
rtoy: Not part of the PR, but noticed that "is fixed at the construction" -> "is fixed at construction"

index.html
```diff
@@ -9284,9 +9345,9 @@ <h2 id="defining-a-valid-audioworkletprocessor">
                 The output audio buffer that is to be consumed by the user
                 agent. <code>outputs[n][m]</code> is a
                 <code>Float32Array</code> object containing the audio samples
-                for <code>m</code>th channel of <code>n</code>th output. While
-                the number of outputs is fixed at the construction, the number
-                of channels can be changed dynamically.
+                for <code>m</code>th channel of <code>n</code>th output. The
+                number of channels in the output can be dynamically changed
+                only when the node has single output.
```
rtoy: A link to the section describing this would be helpful.

index.html
hoch: Hmm. I actually prefer this way. I think a list is better than using `<table>` element. We only have 3 cases with a single description for each.

index.html
hoch: Done.

index.html
hoch: Done.

index.html
hoch: Rephrased it with a link.

index.html
rtoy: Ok

index.html
```diff
@@ -9137,6 +9138,20 @@ <h2 id="AudioWorkletNodeOptions">
                   <a data-link-for="AudioNode">numberOfOutputs</a> attribute.
                 </dd>
                 <dt>
+                  <code><dfn>outputChannelCount</dfn></code> of type
+                  <span class="idlAttrType"><code>sequence&lt;unsigned
+                  long&gt;</code></span>
+                </dt>
+                <dd>
+                  This is used to configure the number of channels in each
+                  output. <code>IndexSizeError</code> MUST be thrown if the
```
rtoy: I think you still need to say a little bit more on what this means.  Something like "outputChannelCount[n]" specifies the number of channels for output number n". Or something.

hoch:
@joeberkovitz Could you take a look?

hoch:
@padenot Could you take a look?

index.html
```diff
@@ -9137,6 +9138,24 @@ <h2 id="AudioWorkletNodeOptions">
...
+                  each output. For example, <code>outputChannelCount: [n,
+                  m]</code> specifies the number of channels in the first
+                  output to <code>n</code> and the second output to
+                  <code>m</code> respectively. <code>IndexSizeError</code> MUST
+                  be thrown if the length of sequence does not match
+                  <a data-link-for=
+                  "AudioWorkletNodeOptions">numberOfOutputs</a>. A
+                  <code>NotSupportedError</code> exception MUST be thrown if a
+                  channel count is not in the valid range of <a data-link-for=
+                  "AudioNode">channelCount</a>.
```
joeberkovitz: This seems confusing -- first of all, `channelCount` is not a range, it's a single number. Also, it's not clear if you mean the option dictionary value `AudioNodeOptions.channelCount`, or the read-only attribute `AudioNode.channelCount` (I'm guessing you mean the options value, but then you have to account for what happens if this was omitted, in which case there is no defined channel count for up/downmixing).

index.html
```diff
@@ -9150,6 +9169,51 @@ <h2 id="AudioWorkletNodeOptions">
...
+                channel configurations can be achieved.
+              </p>
+              <ol>
+                <li>
+                  <a data-link-for="AudioWorkletNodeOptions">numberOfInputs</a>
+                  = 0, <a data-link-for=
+                  "AudioWorkletNodeOptions">numberOfOutputs</a> = 0
+                  <ul>
+                    <li>
+                      <code>NotSupportedError</code> MUST be thrown.
```
joeberkovitz: must be thrown... by the constructor

index.html
hoch: What I meant was `AudioNode.channelCount` here. We have a defined range for that property and the same rules should be applied to each value in `AudioWorkletNodeOptions.outputChannelCount` array. I thought `data-link-for` was quite clear, but it's not visible in the actual page. So I edited the text to clarify what `channelCount` means here.

Also the behavior on unspecified (omitted) `outputChannelCount` is described in the following section.

index.html
hoch: Done.

PR #1397, opened 2017-10-06
Fix #1385 AnalyserNode.channelCount should probably be 2

### LINKED ISSUE #1385: AnalyserNode.channelCount should probably be 2
Currently, the spec says the `channelCount` is 1 with a `channelCountMode` of "max".  Since the mode is "max", the count has no bearing on the computed number of channels.

In addition, the default value for `channelCount` is 2 for an AudioNode.

Changing the spec to say 2 instead of 1 would be better; there's no real reason for the value to be 1. But perhaps it's too late for that.

mdjp:
***
[Preview](https://s3.amazonaws.com/pr-preview/WebAudio/web-audio-api/1385-AnalyserNode.channelCount-should-be-2.html) | [Diff](https://s3.amazonaws.com/pr-preview/WebAudio/web-audio-api/23c9dcc...fe35220.html)


index.html
```diff
@@ -18233,12 +18233,12 @@ <h2 id="requirements">
         Corporation); Mandyam, Giridhar (Qualcomm Innovation Center, Inc);
         Noble, Jer (Apple, Inc.); O'Callahan, Robert(Mozilla Foundation);
         Onumonu, Anthony (British Broadcasting Corporation); Paradis, Matthew
-        (British Broadcasting Corporation); Raman, T.V. (Google, Inc.);
-        Schepers, Doug (W3C/MIT); Shires, Glen (Google, Inc.); Smith, Michael
-        (W3C/Keio); Thereaux, Olivier (British Broadcasting Corporation); Toy,
-        Raymond (Google, Inc.); Verdie, Jean-Charles (MStar Semiconductor,
-        Inc.); Wilson, Chris (Google,Inc.) - Specification Co-editor; ZERGAOUI,
-        Mohamed (INNOVIMAX)
+        (British Broadcasting Corporation) - WG Chair; Raman, T.V. (Google,
```
rtoy: I'm happy to take this change.  Just not in this PR.

rtoy:
We probably want to add `.support` to `.gitignore`; it's generated if you build and run tidy locally.  I'll do that in a separate PR.

index.html
mdjp: Ha, its one of those things that was in the back of my head to do so thought i'd get it sorted. You are of course right though, will sort it.

rtoy:
Arrgh. I merged #1358 which conflicts with this.

I'll fix this for you.

PR #1401, opened 2017-10-06
Fix #1395: ABSN loop limited by duration

### LINKED ISSUE #1395: ABSN loop duration parameter
Setting the loop attribute to true causes playback of the region of the buffer defined by the endpoints loopStart and loopEnd to continue indefinitely, once any part of the looped region has been played. While loop remains true, looped playback will continue until stop() is called, or the scheduled stop time has been reached.

From #1387 We need to clearly state that looped playback will continue untill stop OR the initial duration has been reached.

rtoy:
Add some text that explain the condition where looping stops.  This
includes when a `duration` value is specified when calling `start`.

***
[Preview](https://s3.amazonaws.com/pr-preview/rtoy/web-audio-api/1395-absn-looping-with-duration.html) | [Diff](https://s3.amazonaws.com/pr-preview/WebAudio/web-audio-api/ea86376...rtoy:a0d5d01.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -7791,7 +7791,7 @@ interface AudioBufferSourceNode : AudioScheduledSourceNode {
                 </tr>
                 <tr>
                   <td class="prmName">
-                    <dfn>duration</dfn>
+                    duration
                   </td>
                   <td class="prmType">
                     <a><code>double</code></a>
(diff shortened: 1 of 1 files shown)


rtoy:
Thanks!

PR #1407, opened 2017-10-09
Fix #1402: Add links to looping section

### LINKED ISSUE #1402: LInkify loop/loopStart/loopEnd in Looping section
In [Looping](https://webaudio.github.io/web-audio-api/#looping-AudioBufferSourceNode), there are many highlighted items referring to `loop`, `loopStart`, `loopEnd`, and so on.  Since the previous section is for the `AudioBufferSourceOptions` dictionary that defines members of the same name, it can be confusing what these actually refer to in the looping section.

We should linkify these to point to the attributes of the `AudioBufferSourceNode` itself, and not the members of the `AudioBufferSourceOptions` dictionary.

rtoy:
Add links to `loop`, `loopStart`, and `loopEnd` to clarify what is actually being discussed here.  The preceding section is the `AudioBufferSourceOptions` dictionary, which also as members with the same name.  Adding links makes it clear that we're referring to the attributes of the `AudioBufferSourceNode` and not the dictionary.

***
[Preview](https://s3.amazonaws.com/pr-preview/rtoy/web-audio-api/1402-linkify-looping-section.html) | [Diff](https://s3.amazonaws.com/pr-preview/WebAudio/web-audio-api/ac422a9...rtoy:b9047c4.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -8000,37 +8000,44 @@ dictionary AudioBufferSourceOptions {
             normative requirements.</em>
           </p>
           <p>
-            Setting the <code>loop</code> attribute to true causes playback of
-            the region of the buffer defined by the endpoints
-            <code>loopStart</code> and <code>loopEnd</code> to continue
-            indefinitely, once any part of the looped region has been played.
-            While <code>loop</code> remains true, looped playback will continue
-            until <code>stop()</code> is called, or the scheduled stop time has
-            been reached.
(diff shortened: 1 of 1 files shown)


rtoy:
I think it's pretty clear that they're referring to the attributes, but we might as well.

mdjp:
Very true, but happy with the additional link. This looks good to me now.

PR #1418, opened 2017-10-13
Remove confusing misuse of “mutually perpendicular” in AudioListener ctor args.

### LINKED ISSUE #1298: Term "linear independent" confused with "mutually perpendicular"
The spec at https://www.w3.org/TR/webaudio/#widl-SpatialListener-forwardX says

```
These values are expected to be linearly independent (at right angles to each other),
and unpredictable behavior may result if they are not.
```

The term of linear independence does not mean being at right angles. Two vectors `u` and `v` are linearly independent if one is not a scalar multiple of the other, i.e. if there does not exist a scalar `k` such that `v=k*u`. This is much weaker than being at a right angle.

Two vectors that are at right angles at each other are also linearly independent, but two vectors that are linearly independent are not necessarily also mutually perpendicular to each other. For example, vectors `u=(0,1,0)` and `v=(0,1,1)` are linearly independent, since there does not exist a `k` such that `u=k*v`, but the angle between these two vectors is only 45 degrees.

Does the doc want to say that unpredictable behavior may occur if the two vectors are not mutually perpendicular, or that unpredictable behavior may occur if the two vectors are not linearly independent?

----

Sidelining a bit, whichever the desired answer is to the above question, it is not required for unpredictable behavior to exist in either case, but it would be possible to apply a fixed process to always complete a 3D frame from given two input vectors:

If the two vectors are not mutually perpendicular, it is possible to well define the result by perpendicularizing one over the other using [Gram-Schmidt process](https://en.wikipedia.org/wiki/Gram%E2%80%93Schmidt_process). That behaves well, and e.g. perpendicularizing the up vector at the reference of the forward vector (forward direction takes precedence over up), would be predictable, as long as it is documented well.

If the two vectors are linearly dependent and as such, the set `forward,up,cross(forward,up)` is degenerate, it is still possible to well define the result, by documenting steps that would be taken in the degenerate case to "complete" the set to span a 3D frame. For example, if `forward=(0,0,0)`, it could be documented that a fixed direction `forward=(0,0,-1)` is then assumed (which is the documented default `forward` direction). If, after that, the direction `up` is linearly dependent to `forward`, then a default direction `up=(0,1,0)` could be used, and if after that the vectors are still linearly dependent (in the case `forward=(0,1,0), up=(0,0,0)`), it could be stated that a second default direction `up=(1,0,0)` is used.

joeberkovitz:
Fixes #1298.

Also:
- Make variable naming in spatialization algorithm consistent with attribute names.
- Add a case to detect linearly dependent forward/up listener vectors in spatialization algorithm.
- Fix assorted typos in Listener/Panner descriptions.

***
[Preview](https://s3.amazonaws.com/pr-preview/WebAudio/web-audio-api/1298-linear-independent.html) | [Diff](https://s3.amazonaws.com/pr-preview/WebAudio/web-audio-api/f4af47d...81690f6.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -11464,7 +11464,7 @@ interface PannerNode : AudioNode {
             <dd>
               <p>
                 Sets the x coordinate position of the audio source in a 3D
-                Cartesian system. \infty)\).
+                Cartesian system.
               </p>
               <div class="audioparam-info">
                 <table>
@@ -12107,11 +12107,9 @@ dictionary PannerOptions : AudioNodeOptions {
           the listener. In simple human terms, the <code>forward</code> vector
           represents which direction the person's nose is pointing. The
           <code>up</code> vector represents the direction the top of a person's
-          head is pointing. These values are expected to be linearly
-          independent (at right angles to each other), and unpredictable
-          behavior may result if they are not. For normative requirements of
-          how these values are to be interpreted, see the
-          <a>Spatialization/Panning</a> section.
+          head is pointing. These two vectors are expected to be linearly
+          independent. For normative requirements of how these values are to be
+          interpreted, see the <a>Spatialization/Panning</a> section.
         </p>
         <pre class="idl">
 [Exposed=Window]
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -20535,20 +20532,32 @@ <h3 id="Spatialization-background">
...
+  // Determine a unit vector orthogonal to listener's right, forward
+  let listenerRightNorm = listenerRight.normalize();
+  let listenerForwardNorm = listenerForward.normalize();
+  let up = listenerRightNorm.cross(listenerForwardNorm);
+
   let upProjection = sourceListener.dot(up);
   let projectedSource = sourceListener.diff(up.scale(upProjection)).normalize();
 
-  azimuth = 180 * Math.acos(projectedSource.dot(listenerRight)) / PI;
+  azimuth = 180 * Math.acos(projectedSource.dot(listenerRightNorm)) / PI;
```
rtoy: While we're here: "PI" -> "Math.PI"?

PR #1421, opened 2017-10-15
Fetch MathJax from cdnjs

pioug:
Looking for some improvements to make the page loads faster. 

I found that fetching assets from W3.org is really slow (from Singapore at least):
image

If we use cdnjs instead:

image

***
[Preview](https://s3.amazonaws.com/pr-preview/pioug/web-audio-api/gh-pages.html) | [Diff](https://s3.amazonaws.com/pr-preview/WebAudio/web-audio-api/5cd4fbf...pioug:870598b.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -9,22 +9,22 @@
     <script src='https://www.w3.org/Tools/respec/respec-w3c-common' class=
     'remove'></script>
     <link rel="preload" href=
-    "https://www.w3.org/scripts/MathJax/2.6.1/MathJax.js?config=TeX-AMS-MML_HTMLorMML"
+    "https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.6.1/MathJax.js?config=TeX-AMS-MML_HTMLorMML"
     as="script">
     <link rel="preload" href=
-    "https://www.w3.org/scripts/MathJax/2.6.1/jax/output/HTML-CSS/jax.js?rev=2.6.1"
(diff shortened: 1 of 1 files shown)


joeberkovitz:
I have no idea but it does involve relying on an outside open source project hosted by a non-W3C infrastructure company (Cloudflare); either the project or the company could conceivably change its URL or even go away.

I think the best answer to making spec page loads faster would be for the W3C to use a CDN for its static resources rather than for different specs to go to different CDNs on a one-off basis, and to assign that CDN a stable domain name that W3C commits to maintaining. I will ask about this but in the meantime I'd leave the spec as-is until we find out more.

dontcallmedom:
There is indeed right now a rule forbidding TR documents to use non-w3.org resources - for sake of making our specs work in the long term, no matter what may happen to this or that external service provider.

I have started a conversation within the staff to figure out what kind of trade-off we can make here. Using a w3.org CDN would be interesting, but may not be trivial.

joeberkovitz:
Closing since W3C needs to treat this as a systemic issue and can't merge this PR.

chrisguttandin:
Sorry for commenting on a closed issue, but would it be an option to add a Service Worker to cache the assets? I could give it a try if you like.

joeberkovitz:
@chrisguttandin This issue is about a request to edge-cache assets using a CDN, for which ServiceWorkers would have no effect. The current spec already caches all static assets locally for 14 days, as I've just verified. So the problem only affects initial page loads after this local cache has expired, which isn't so terrible.

chrisguttandin:
Ok, I understand why a ServiceWorker wouldn't be useful in that situation. Thanks for your explanation.

PR #1423, opened 2017-10-18
Ask for tests for normative changes in README.md

foolip:
See https://github.com/foolip/testing-in-standards/blob/master/policy.md
for context. This phrasing matches that used for many other specs'
CONTRIBUTING.md files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,3 +10,11 @@ This repository contains the latest editor's draft of the [W3C AudioWG](https://
 
 You can preview the current version of the master branch [here](https://webaudio.github.com/web-audio-api/).
 
+# Tests
+
+For normative changes, a corresponding
+[web-platform-tests](https://github.com/w3c/web-platform-tests) PR is highly appreciated. Typically,
+both PRs will be merged at the same time. Note that a test change that contradicts the spec should
+not be merged before the corresponding spec change. If testing is not practical, please explain why
+and if appropriate [file an issue](https://github.com/w3c/web-platform-tests/issues/new) to follow
+up later. Add the `type:untestable` or `type:missing-coverage` label as appropriate.


foolip:
There wasn't a CONTRIBUTING.md, so I went with README.md.

Some additional context is that in https://foolip.github.io/day-to-day/ one can see that there's a lot of activity on the Web Audio spec, but not much on shared tests. I know there's been a lot of recent IDL changes so that the numbers aren't very meaningful, but anyhow that's why I came here.

I'm not wedded to any particular way of phrasing this or where, or indeed to the details of the working mode at all. If there'a anything that needs to be done to make interop testing of Web Audio easier, that'd be great to learn about.

rtoy:
We are working on tests. Chrome has converted (almost) all of the tests to use testharness and will be moving tests to wpt soon.  Firefox is also moving their tests to wpt too.

foolip:
> We are working on tests. Chrome has converted (almost) all of the tests to use testharness and will be moving tests to wpt soon. Firefox is also moving their tests to wpt too.

Sounds perfect! I know that in Blink there are *-expected.wav files, a different type of test. If you find that you need something like that to effectively use web-platform-tests, then please file an issue on the wpt repo with the infra label. Can't promise that anyone will fix it, but at the very least I'd like to understand what testing pain points exist in all parts of the web platform.

rtoy:
On Wed, Oct 18, 2017 at 8:21 AM, Philip Jägenstedt <notifications@github.com
> wrote:

> We are working on tests. Chrome has converted (almost) all of the tests to
> use testharness and will be moving tests to wpt soon. Firefox is also
> moving their tests to wpt too.
>
> Sounds perfect! I know that in Blink there are *-expected.wav files, a
> different type of test. If
>
​Not exactly sure how generally useful those tests are, and we've been kind
of moving away from that when possible.  Except (I think) for our codec
tests, these are used as references and comparisons are done based on max
diff and/or SNR between that reference and the test result.  Unlike how
they used to be done by comparing (bitwise) the *-actual.wav against the
*-expected.wav.​

​This was a giant pain to maintain if different platforms or different OS
versions produced different results.​

-- 
Ray

foolip:
OK, sounds like we shouldn't be in a hurry to add something similar to wpt, but if it becomes a blocker to sharing tests at any point I'd love to know about it.

Issue #1139, opened 2017-01-12
Behaviour when navigating away from a document using the Web Audio API


padenot:
(This issue is in reponse to a [Gecko bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1325215#c22), comment 22 onwards, where Boris mentions that what happens on navigation is not defined (comment 28), and Domenic explain what happens with document navigation and promises (comment 34)).

For now, this is not specced. What happens if you're calling `AudioContext.decodeAudioData`, or `AudioContext.resume`, etc., and immediately navigate away from the page? Should the sound stop when navigating away from the page ? Should `OfflineAudioContext`'s rendering stop as well?

In general, I think we want to stop everything (more or less what is being done by implementations):
- Reject any promise in flight (i.e. a promise that has been created but has not been resolved or rejected). We could also drop them on the floor. 
- Do something like `AudioContext.close` (without involving the promise, i.e., do the same operation): stop the sound, stop the clock, allow freeing resources, etc. This would require splitting up the text for `close` into its own abstract operation and calling that from the steps to to execute when unloading a document, and from the `close` method itself.

Practically speaking, we need to insert a bunch of checks for an "active document" throughout, register all promises used on private slots on the `BaseAudioContext` so we can access them at any time, and add a section that explains the "unloading document cleanup steps".

There is already a difference in behaviour between Gecko and Blink on this: Blink rejects the promise, Gecko does not (they are just dropped). We need to decide what to do, and I can write the text once we decide what to do.

domenic:
Thanks for opening this; it's great to get interop on these cases.

To clarify one thing I didn't spell out in detail, the reason you need to

> insert a bunch of checks for an "active document" throughout

is to take care of code like this:

```html
<iframe></iframe>
<script>
const OldAudioContext = frames[0].AudioContext;
iframe.onload = () => {
  // Can I still use OldAudioContext to play sound? The spec currently allows it, I believe.
  const ac = new OldAudioContext();
  const source = ac.createBufferSource();
  source.buffer = dogBarkingBuffer;
  source.connect(context.destination);
  source.start(0);
};

iframe.src = "http://example.com/";
```

Similarly, you'd need a check in the methods (creatBufferSource/connect/start) since I could have done some of those steps before navigation and some of them after.

padenot:
Just a note to self: this needs to prevent creating new nodes on the context in a synchronous manner. As soon as the document goes away, node creation is not allowed anymore. 

[...]

PR #1425, opened 2017-10-19
Spec document unloading

padenot:
This does the following:
- Keep all promises around, so we can deal with them on unload
- Spec what to do when unloading a document: very similar to calling `close()` on all audiocontext
- This stops the rendering of AudioContext, stops decoding of decodeAudioData tasks, stop the rendering of OfflineAudioContext.
- A consequence of this is that there won't be any audio output anymore.
- Another consequence of this (because the rendering loop does not running anymore) is that any control message sent by Web Audio API objects are never going to be executed. We don't have to check for "active document" everywhere.

This deals with #1139.

***
[Preview](https://s3.amazonaws.com/pr-preview/padenot/web-audio-api/unload.html) | [Diff](https://s3.amazonaws.com/pr-preview/WebAudio/web-audio-api/8b7d14d...padenot:47a8e9d.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -2243,7 +2243,10 @@ interface AudioContext : BaseAudioContext {
             </dt>
             <dd>
               <p>
-                If the <a>relevant settings object</a>'s <a>responsible
+                If the <a href=
+                "https://html.spec.whatwg.org/#concept-current-everything">current
+                settings object</a>'s <a href=
+                "https://html.spec.whatwg.org/#responsible-document">responsible
                 document</a> is NOT <a href=
                 "https://html.spec.whatwg.org/multipage/browsers.html#fully-active">
                 fully active</a>, throw an <code>InvalidStateError</code> and
@@ -2485,8 +2488,8 @@ interface AudioContext : BaseAudioContext {
                     This will stop rendering.
                   </div>
                 </li>
-                <li>If this <a>control message</a> is being run because the
-                document is being unloaded, abort this algorithm
+                <li>If this <a>control message</a> is being run in reaction to
+                the document is being unloaded, abort this algorithm
                 </li>
                 <div class="note">
                   We don't have to notify the control thread in this case.
@@ -3024,8 +3027,11 @@ interface OfflineAudioContext : BaseAudioContext {
             </dt>
             <dd>
               <p>
-                If the <a>relevant settings object</a>'s <a>responsible +
-                document</a> is NOT <a href=
+                If the <a href=
+                "https://html.spec.whatwg.org/#concept-current-everything">current
+                settings object</a>'s <a href=
+                "https://html.spec.whatwg.org/#responsible-document">responsible
+                + document</a> is NOT <a href=
                 "https://html.spec.whatwg.org/multipage/browsers.html#fully-active">
                 fully active</a>, throw an <code>InvalidStateError</code> and
                 abort these steps.
@@ -3264,11 +3270,9 @@ interface OfflineAudioContext : BaseAudioContext {
                 <li>Otherwise, in the case that the buffer was successfully
                 constructed, <a>begin offline rendering</a>.
                 </li>
-                <li>Append <var>promises</var> to <a>pendingPromises</a>.
+                <li>Append <var>promise</var> to <a>pendingPromises</a>.
                 </li>
                 <li>Return <var>promise</var>.
-                </li>
-                <li>Append <em>promise</em> to <a>pendingPromises</a>.
                 </li>
               </ol>
               <p>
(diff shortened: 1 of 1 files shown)


padenot:
@domenic, does this makes sense? In particular the last point.

index.html
```diff
@@ -2237,6 +2243,13 @@ <h2 id="AudioContext">
             </dt>
             <dd>
               <p>
+                If the <a>relevant settings object</a>'s <a>responsible
```
rtoy: Don't quite follow "relevant settings object" and "responsible document".

Plus these `<a>` tags don't seem to link up anything, at least in preview.

index.html
```diff
@@ -2463,7 +2481,16 @@ <h2 id="AudioContext">
                 </li>
                 <li>Set the <a>rendering thread state</a> to
                 <code>suspended</code>.
+                  <div class="note">
+                    This will stop rendering.
+                  </div>
+                </li>
+                <li>If this <a>control message</a> is being run because the
```
rtoy: Phrasing seems odd. Do you mean if the control message is being run *when* the document is being unloaded?

index.html
```diff
@@ -2995,6 +3024,13 @@ <h2 id="OfflineAudioContext">
             </dt>
             <dd>
               <p>
+                If the <a>relevant settings object</a>'s <a>responsible +
```
rtoy: Same comment as above for "relevant settings object" and "responsible document".

index.html
```diff
@@ -3228,8 +3264,12 @@ <h2 id="OfflineAudioContext">
                 <li>Otherwise, in the case that the buffer was successfully
                 constructed, <a>begin offline rendering</a>.
                 </li>
+                <li>Append <var>promises</var> to <a>pendingPromises</a>.
```
rtoy: "Promises"? Isn't there only one for `startRendering`?

index.html
```diff
@@ -19633,6 +19673,26 @@ <h2 id="MediaStreamTrackAudioSourceNode">
           </li>
         </ul>
       </section>
+      <section>
+        <h2>
+          Unloading a document
+        </h2>
+        <p>
+          When unloading a document with <a>BaseAudioContext</a>s, the
+          following steps MUST be executed:
```
domenic: This should explicitly say that these are "Unloading document cleanup steps" and link to https://html.spec.whatwg.org/#unloading-document-cleanup-steps

index.html
```diff
@@ -19633,6 +19673,26 @@ <h2 id="MediaStreamTrackAudioSourceNode">
...
+        </p>
+        <ul>
+          <li>Reject all the promises of <a>pendingPromises</a>, for each
+          <a>AudioContext</a> and <a>OfflineAudioContext</a> of this document.
```
domenic: Reject them with what?

index.html
domenic: Relevant settings objects are always "of" something. Since the object is not created yet here, the concept isn't really applicable. You want "current settings object" here. See https://html.spec.whatwg.org/#realms-settings-objects-global-objects in particular

> Note that in constructors, where there is no this value yet, the current concept is the appropriate default.

index.html
padenot: I rephrased with "in reaction to".

index.html
padenot: yes, typo.

padenot:
Oops, closed by mistake. This has been rebased and has all the comments addressed.

joeberkovitz:
@domenic Hi Domenic, just pinging to make sure you've seen Paul's recent updates which he believes will address your feedback on #1139. We'd love to get an updated review from you as soon as it's convenient, since your input is important and this is our last open issue prior to CR. Thank you!

index.html
```diff
@@ -2995,6 +3027,16 @@ <h2 id="OfflineAudioContext">
             </dt>
             <dd>
               <p>
+                If the <a href=
+                "https://html.spec.whatwg.org/#concept-current-everything">current
+                settings object</a>'s <a href=
+                "https://html.spec.whatwg.org/#responsible-document">responsible
+                + document</a> is NOT <a href=
```
domenic: Stray "+" here

index.html
```diff
@@ -19633,6 +19677,29 @@ <h2 id="AudioWorklet-Examples">
...
+        <p>
+          Additional <a href=
+          "https://html.spec.whatwg.org/#unloading-document-cleanup-steps">unloading
+          document cleanup steps</a> are defined for document that uses
+          <a>BaseAudioContext</a>:
+        </p>
+        <ul>
+          <li>Reject all the promises of <a>pendingPromises</a> with
+          <code>InvalidStateError</code>, for each <a>AudioContext</a> and <a>
+            OfflineAudioContext</a> of this document.
```
domenic: "of this document" is kind of imprecise.

I think you'll want: for each AudioContext and OfflineAudioContext whose relevant global object is the same as the document's associated Window.

joeberkovitz:
@padenot it seems that this could be merged any time with the couple of minor changes suggested by @domenic...

padenot:
Yes, I'll be addressing the comments and merging.

PR #1444, opened 2017-11-08
Serialize AudioWorkletNode options dictionary to audio thread side.

### LINKED ISSUE #1442: [audioworklet] node options not passed to processor
It looks from https://webaudio.github.io/web-audio-api/#instantiation-of-AudioWorkletNode-and-AudioWorkletProcessor as though the `options` passed to the node are not communicated to the processor, although this is clearly implied by the text in https://webaudio.github.io/web-audio-api/#AudioWorklet-Sequence

One of the above has to be wrong.

joeberkovitz:
Fix #1442

***
[Preview](https://s3.amazonaws.com/pr-preview/WebAudio/web-audio-api/1442-bring-back-awp-options.html) | [Diff](https://s3.amazonaws.com/pr-preview/WebAudio/web-audio-api/b027c50...86eaf1c.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -18734,23 +18734,38 @@ interface AudioWorkletProcessor {
                 <code>AudioWorkletProcessor</code>
               </dt>
               <dd>
-                Initializes the <a>AudioWorkletProcessor</a> using the contents
-                of the <code>options</code> dictionary that was passed to the
-                constructor for <a>AudioWorkletNode</a>. The contents of this
-                dictionary will have been serialized and deserialized according
-                to the algorithm for <a href=
-                "#instantiation-of-AudioWorkletNode-and-AudioWorkletProcessor">instantiating
-                an AudioWorkletProcessor</a>.
+                When the constructor for <a>AudioWorkletProcessor</a> is
+                invoked, the following steps are performed:
+                <ol>
+                  <li>Compare the value of NewTarget with the <a href=
+                  "https://tc39.github.io/ecma262/#active-function-object">active
+                  function object</a>; if the two are equal, throw a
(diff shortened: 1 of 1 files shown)


joeberkovitz:
The plain-vanilla `StructuredSerialize()` operation is used to serialize the options dictionary. I felt this was probably better than trying to autodetect transferables in the options dictionary, which might or might not result in a predictable behavior. This also handles SharedArrayBuffers, which `StructuredSerializeWithTransfer()` may not.

If developers want to actually transfer (not just clone) transferables to the AWP, they can use `postMessage()` to do this with full control over the details.

hoch:
Don't we have to wait until Bikeshed transtion is finished? Or can we do this in parallel?

joeberkovitz:
@hoch Not going to merge this until Bikeshed work is done — I will reapply by hand at that point, should be easy (same goes for other commit)

rtoy:
So far all the changes have been relatively small and mostly self-contained, so I think we can do it before or  after bikeshed.  If before, we just have to fix the merge conflicts in bikeshed.

In some ways I'd prefer we do the fix now so that the information is out there.  (This isn't so important, but moving the worklet to BaseAudioContext is a big change.)

joeberkovitz:
@rtoy if this looks OK to you I don't mind merging it now, it's up to you -- I thought you might want me to wait.

rtoy:
Depends on when bikeshed is done.  If it's soon, then waiting seems best.  I won't know where we are until next week probably.

joeberkovitz:
Fixes #1447 as well.

hoch:
Non-editor LGTM. Thanks for working on this!

Issue #1441, opened 2017-11-06
[audioworklet] Text of processor registration steps is corrupted


joeberkovitz:
The text of step 9 of the `registerProcessor()` algorithm (see https://webaudio.github.io/web-audio-api/#methods-13) is corrupted. It's not clear how to read this.

hoch:
Hmm. I don't know how or when this happened. I'll fix it today.

joeberkovitz:
@hoch I already made this fix in #1445

Issue #1436, opened 2017-10-31
[audioworklet] Enforcing SecureContext on audioWorklet.addModule()


hoch:
I don't think we have a broader consensus on this, but it might be worth to discuss it in our WG. Here are related issues from TAG and CSS Houdini TF:
https://github.com/w3ctag/design-principles/pull/75
https://github.com/w3c/css-houdini-drafts/pull/379

In short, the argument is that HTTPS should be enforced on `audioWorklet.addModule()`.

I generally agree with this idea because running JS code on the higher priority thread (audio rendering thread) should take extra caution. We might want to see this as a feature rather than a restriction.

In terms of the spec work, this work needs to be done in the Worklet - not in its variants (Paint/Audio/Animation...). However, if we find a potential issue on it we want to raise a flag before it's too late.

joeberkovitz:
We're closing since we've discussed and it falls to Worklet to decide how this goes... our spec doesn't have to change,

hoch:
Houdini WG issue: https://github.com/w3c/css-houdini-drafts/issues/505

annevk:
I think this should be reopened since it does directly affect your specification. E.g., we shold not expose `self.audioWorklet` on insecure contexts.

joeberkovitz:
@annevk Is there a clear writeup of what paint worklets are doing re: this issue? It'd be easier than reverse engineering the intent from Chromium diffs. My guess is that we add `[SecureContext]` to the `audioWorklet` attribute definition in `AudioContext` but I didn't see a flat-out statement in the discussion minutes or bug threads.

Reopening to make sure we track this.

annevk:
From https://groups.google.com/a/chromium.org/d/msg/blink-dev/Jex3idOld48/VeOxMUNDBAAJ:

> Ship CSS Paint in secure contexts only, and also block future types of worklets (e.g. audio worklet) on secure contexts.

I can state that this also goes for Mozilla.

annevk:
(You didn't hit reopen btw.)

hoch:
> E.g., we shold not expose self.audioWorklet on insecure contexts.

@annevk Now we have `AudioWorklet` interface (which extends Worklet) and it should have the attribute of `[SecureContext]`. I believe that's good enough.

@joeberkovitz We can add this change to PR #1445 since it's really tiny.

annevk:
> I believe that's good enough.

[...]

PR #1445, opened 2017-11-08
Move worklet under BaseAudioContext.

### LINKED ISSUE #1435: [audioworklet] Initialization of `sampleRate` in AudioWorkletGlobalScope
Consider this example:
```js
class MyProcessor extends AudioWorkletProcessor {
  static get parameterDescriptors() {
    return [{
      name: 'frequency',
      defaultValue: 440,
      minValue: 0,
      maxValue: sampleRate * 0.5
    }];
  }

  constructor() {
    super();
  }
  process() {}
}
```

The value of `sampleRate` in this example is evaluated when `audioWorklet.addModule(url)` gets called. So it is possible that there is no valid/active BaseAudioContext is running. Thus, `maxValue` of AudioParam becomes undefined.

Interestingly enough, having `sampleRate` and `currentTime` in AudioWorkletProcessor does not really solve the issue as well, so this could be a structural problem caused by the orthogonality of AudioWorklet and BaseAudioContext.

The spec also seem to be missing when `sampleRate` and `currentTime` are initialized and updated. I propose they are initialized to 0 when AudioWorkletGlobalScope is created, and `currentTime` is updated before every render quantum.

joeberkovitz:
Fixes #1435

***
<a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/pull/1445.html" title="Last updated on Nov 28, 2017, 4:13 PM GMT">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/b027c50...a7e220d.html" title="Last updated on Nov 28, 2017, 4:13 PM GMT">Diff</a>


joeberkovitz:
This is a WIP proposal for the fix to #1435 - please take a look. I have probably missed some things that need changing but tried to hit the main points (including the diagram).

index.html
```diff
@@ -18016,8 +18037,8 @@ <h2 id="AudioWorklet-concepts">
           <pre class="example" title=
           "Importing a script and creating AudioWorkletNode">
 // The main global scope
-window.audioWorklet.addModule("bypass.js").then(function () {
-  var context = new AudioContext();
+var context = new AudioContext();
```
hoch: Perhaps we should use `let` instead of `var`. L18042 as well.

index.html
```diff
@@ -18016,8 +18037,8 @@ <h2 id="AudioWorklet-concepts">
...
+context.worklet.addModule("bypass.js").then(function () {
```
hoch: Let's use single quote for the path string.

index.html
```diff
@@ -19208,8 +19206,8 @@ <h2 id="AudioWorklet-Examples">
               <a><code>AudioWorkletProcessor</code></a>.
             </p>
             <pre class="example" title="BitCrusher - Global Scope">
-window.audioWorklet.addModule('bitcrusher.js').then(function () {
-  let context = new AudioContext();
+let context = new AudioContext();
+context.Worklet.addModule('bitcrusher.js').then(function () {
```
hoch: s/function ()/() =>/

index.html
```diff
@@ -17971,6 +17982,13 @@ <h2 id="AudioWorklet-concepts">
             </figcaption>
           </figure>
           <p>
+            Each <a>BaseAudioContext</a> possesses exactly one
+            <a>AudioWorklet</a>. Scripts imported using this
+            <a>AudioWorklet</a> are run in one or more
```
hoch: "one or more"

Is this correct?

```js
contextA.worklet.addModule('a.js');
contextB.worklet.addModule('b.js');
```

A processor defined in `a.js` will not be available for `contextB`. I guess AudioWorklet in L17987 means a "general AudioWorklet", but I find a bit confusing. I think we should clarify there is one-to-one mapping between a context and a worklet global scope.

index.html
```diff
@@ -877,6 +877,7 @@ <h2 id="BaseAudioContext">
     readonly        attribute double               currentTime;
     readonly        attribute AudioListener        listener;
     readonly        attribute AudioContextState    state;
+    readonly        attribute Worklet              worklet;
```
hoch: Not sure if it's necessary, but it seems we're missing `[SameObject]` attribute.

index.html
joeberkovitz: Actually I think it’s 1:N for contexts and scopes since we may want to ensure some indeterminacy wrt where rendering work is done to avoid erroneous caching. 

AudioWorklet here should really say “Worklet” since there’s no AudioWorklet interface (applies elsewhere also).

index.html
joeberkovitz: When does one use that? What does it mean?

index.html
hoch: > Actually I think it’s 1:N for contexts and scopes since we may want to ensure some indeterminacy wrt where rendering work is done to avoid erroneous caching.

I am confused. Is this something we discussed in person?

> AudioWorklet here should really say “Worklet” since there’s no AudioWorklet interface (applies elsewhere also).

I agree, but that actually opens up another editorial issue. We have a whole section named "the AudioWorklet interface".

joeberkovitz: In a recent meeting the group did discuss the benefits of having more than one scope per context to avoid developers unintentionally saving state in global variables and counting on it being there. Apparently PaintWorklets always use 2 scopes and randomly switch between them. However, we do need to make sure that any given AWP always runs in the same scope, in order to be able to reliably maintain its state -- think filters, delay lines, etc. So... what is this indeterminacy buying us really?

OTOH it's also possible that in the future we might support more than one rendering thread per AudioContext; that was discussed quite a long time ago. This would also result in >1 AWGS per context.

I'd appreciate hearing from other WG folks on these points since we should try to get this right. Certainly at the moment in practice impls will do a 1:1 relationship, it's just a question of futureproofing the spec.

index.html
hoch: https://heycam.github.io/webidl/#SameObject

Once it's constructed/assigned by UA, the reference of `worklet` would not be changed. So `worklet` getter guarantees that you get the identical reference every time.

index.html
hoch: I think 1:N mapping between the main and the worklet global scope only makes sense for PaintWorklet case. PaintWorklet is more about supporting "state-less function call" whereas AudioWorkletProcessor is a complete class structure with various states inside. We need to support "stateful" programming model, thus maintaining a consistent global scope is imperative.

index.html
```diff
@@ -17997,28 +18021,32 @@ <h2 id="AudioWorklet-concepts">
...
   }
 });
 </pre>
           <pre class="example" title=
           "Importing a script and creating AudioWorkletNode">
 // The main global scope
-window.audioWorklet.addModule("bypass.js").then(function () {
-  var context = new AudioContext();
-  var bypass = new AudioWorkletNode(context, "Bypass");
+var context = new AudioContext();
```
hoch: s/var/let/

index.html
```diff
@@ -19156,25 +19168,25 @@ <h2 id="AudioWorklet-Sequence">
...
-            <li>As part of the global scope's initialization, the set of
-            imported scripts is run, including the one that was previously
-            imported.
+            <li>The imported script is run in the newly created global scope.
             </li>
             <li>As part of running the imported script, an
             <a>AudioWorkletProcessor</a> is registered under the key
             <code>"Custom1"</code> within the <a>AudioWorkletGlobalScope</a>.
+            This populates maps both in the global scope and in the
+            <a>AudioContext</a>.
```
hoch: s/AudioContext/BaseAudioContext/

Here we have some cross-thread operations going on. Not sure if this captures the correct "order" of sequence with the proper spec language. ("queue a task" and all?)

index.html
joeberkovitz: +1

index.html
joeberkovitz: This is the non-normative section. I added some normative "queue a task" language in the newly added step 11 of the normative description of `registerProcessor()`; see here:

https://s3.amazonaws.com/pr-preview/WebAudio/web-audio-api/1435-context-owns-worklet.html#dom-audioworkletglobalscope-registerprocessor

joeberkovitz: BTW I believe the usage of `AudioContext` is correct since this is a non-normative example and states it's not a general description but a specific narrative of one use case.

hoch: Still LGTM! Thanks for this change! 👍

joeberkovitz:
Fixes #1441 also

rtoy:
Nit: In BaseAudioContext attributes section, the attributes are listed alphabetically, so `audioWorklet` needs to be first.  (This is a hold-over from the original respec version that automatically alphabetized things, and we kept it that way after contiguous idl.)

index.html
```diff
@@ -18202,11 +18209,16 @@ <h2 id="AudioWorkletGlobalScope">
...
                   </li>
-                  <li>d the key-value pair (<em>name</em> -
+                  <li>Add the key-value pair (<em>name</em> -
                   <em>definition</em>) to the <a>node name to processor
                   definition map</a> of the associated
                   <a>AudioWorkletGlobalScope</a>.
                   </li>
+                  <li>Queue a task to the <a>control thread</a> to add the
+                  key-value pair (<em>name</em> - <em>descriptors</em>) to the
+                  <a>node name to parameter descriptors map</a> of the
```
rtoy: This link isn't working. I think you want "descriptor", not "descriptors".

joeberkovitz:
@rtoy done. anything else?

index.html
```diff
@@ -19156,25 +19168,25 @@ <h2 id="AudioWorklet-Sequence">
...
             <li>An <a>AudioContext</a> is created.
             </li>
-            <li>An <a>AudioWorkletGlobalScope</a> is created in association
-            with the context's audio rendering thread. This is the global scope
-            in which <a>AudioWorkletProcessor</a> class definitions will be
-            evaluated.
+            <li>In the main scope, <code>context.audioWorklet</code> is
+            requested to add a script module.
+            </li>
+            <li>Since none exists yet, a new <a>AudioWorkletGlobalScope</a> is
```
rtoy: Although not normative text, we should explain what happens on the second call, when the global scope exists.

index.html
```diff
@@ -19156,25 +19168,25 @@ <h2 id="AudioWorklet-Sequence">
...
+            class definitions will be evaluated.
             </li>
-            <li>As part of the global scope's initialization, the set of
-            imported scripts is run, including the one that was previously
-            imported.
+            <li>The imported script is run in the newly created global scope.
             </li>
             <li>As part of running the imported script, an
             <a>AudioWorkletProcessor</a> is registered under the key
             <code>"Custom1"</code> within the <a>AudioWorkletGlobalScope</a>.
```
rtoy: Didn't notice this before, but explicitly saying "Custom1" is the key here is really confusing because I was reading this to be that the key really is "Custom1".  In fact, I think what you're saying here is that the processor is registered under the user-specified key, **not** actually "Custom1".

This can be addressed in a different PR.

index.html
```diff
@@ -19156,25 +19168,25 @@ <h2 id="AudioWorklet-Sequence">
...
+            This populates maps both in the global scope and in the
+            <a>AudioContext</a>.
+            </li>
+            <li>The promise for the addModule() call is resolved.
             </li>
             <li>In the main scope, an <a>AudioWorkletNode</a> is created using
             the key <code>"Custom1"</code> along with an <code>opts</code>
```
rtoy: Missed this before too, but "opts" dictionary should probably be a link to something so I can actually look up exactly what this is.  This can be addressed in a different PR.

index.html
joeberkovitz: These are variables, not links to a definition, and the _descriptors_ variable is already defined above. I would have used a `<var>` tag but `<em>` was already used by existing language in this section.

index.html
joeberkovitz: +1

index.html
joeberkovitz: Actually +1 on this fix because the diagram had changed to use different variable names (originally, it said Custom1).

index.html
joeberkovitz: Same as above

index.html
rtoy: Sorry, I meant `<a>node name to parameter descriptors map</a>` is not linked. I think it should be "descriptor".

index.html
```diff
@@ -19208,8 +19220,8 @@ <h3>
               <a><code>AudioWorkletProcessor</code></a>.
             </p>
             <pre class="example" title="BitCrusher - Global Scope">
-window.audioWorklet.addModule('bitcrusher.js').then(function () {
-  let context = new AudioContext();
+let context = new AudioContext();
+context.Worklet.addModule('bitcrusher.js').then(() =&gt; {
```
hoch: s/Worklet/audioWorklet/

index.html
```diff
@@ -17942,6 +17954,11 @@ <h3>
         <h2 id="AudioWorklet">
           The <dfn>AudioWorklet</dfn> Interface
         </h2>
+        <pre class="idl">
+[Exposed=Window]
```
hoch: `[Exposed=Window, SecureContext]`

joeberkovitz:
Fixes #1436

PR #1488, opened 2018-02-07
Fix #1464: Optimize AudioParams for AudioWorklets

### LINKED ISSUE #1464: [audioworklet] Optimizing AudioParams for AudioWorklet
As currently spec'ed, an `AudioWorklet` gets a `Float32Array` of length 128 containing all of the parameter values for the current rendering quantum.  This is perfect if the `AudioParam` is actually changing during this period.

However, in many cases, the `AudioParam` value is actually fixed.

The worklet then either has to check that the AudioParam is constant
or just assume that it is changing and do the necessary computations.

For example, consider a `BiquadFitler`.  The `AudioParam`'s have a
complicated relationship between the values and the filter
coefficients, which is pretty expensive to have to compute for every
sample, even if the filter isn't changing.  Checking that the
parameter is constant isn't free either, and there are 4 `AudioParams`
for a biquad.

Proposal: If the `AudioParam` is known to be constant for the render
quantum, pass an array of length 1 containing just the one fixed
value.  Otherwise, pass a 128-element array for all values. This is a
relatively simple change and allows some potentially important
optimizations for the worklet.

rtoy:
State that the `Float32Array` for an `AudioParam` of an `AudioWorklet` may have a length of 1 if the `AudioParam` is constant over the current render quantum.

***
<a href="https://pr-preview.s3.amazonaws.com/rtoy/web-audio-api/pull/1488.html" title="Last updated on Mar 27, 2018, 12:41 AM GMT (1dffc02)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1488/8101b38...rtoy:1dffc02.html" title="Last updated on Mar 27, 2018, 12:41 AM GMT (1dffc02)">Diff</a>


rtoy:
Ping.

rtoy:
Updated for the bikeshed version of the spec.

PTAL.

PR #1493, opened 2018-02-13
Fix #1491: Add currentFrame to worklet global scope

### LINKED ISSUE #1491: Add currentFrame to AudioWorkletGlobalScope
The `AudioWorkletGlobalScope` has `currentTime` to let the processor know the current time.  It would also be very nice to have `currentFrame` that is the frame counter where `currentTime = currentFrame/sampleRate`.

For algorithms that need precise timing and complex computations wrt to time, using frames reduces round-off errors.

In Chrome, currentTime used to be accumulated each render quantum, causing a small round-off each time.  This slight drift was noticeable after even a short time.  Switching Chrome's internal implementation to count frames got rid a huge class of round-off errors.  

Doing the same in Chrome's AudioParam implementation greatly reduced the small, but subtle issues in timing.  Before fixing this, @hoch  had an example where his synthesizer would suddenly stop producing output because of round-off in the audio param implementation.  Changing the implementation to work with frames instead of time fixed this and a bunch of other issues.

rtoy:
Add `currentFrame` attribute to AudioWorkletGlobalScope and describe
it and its relationship to `currentTime`.

***
<a href="https://pr-preview.s3.amazonaws.com/rtoy/web-audio-api/pull/1493.html" title="Last updated on Mar 28, 2018, 6:10 PM GMT (24a9583)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1493/34fc87c...rtoy:24a9583.html" title="Last updated on Mar 28, 2018, 6:10 PM GMT (24a9583)">Diff</a>


index.html
```diff
@@ -18193,6 +18193,7 @@ <h2 id="AudioWorkletGlobalScope">
 [Global=(Worklet, AudioWorklet), Exposed=AudioWorklet]
 interface AudioWorkletGlobalScope : WorkletGlobalScope {
     void registerProcessor (DOMString name, VoidFunction processorCtor);
+    readonly attribute double currentFrame;
```
hoch: Perhaps `unsigned long long`?

index.html
```diff
@@ -18193,6 +18193,7 @@ <h2 id="AudioWorkletGlobalScope">
...
     readonly attribute double currentTime;
```
hoch: I am not sure about keeping both. However, we'll have to define `currentFrame` in the spec if we remove the current time here.

index.html
hoch: The range would be [0, 18446744073709551615]. I think this is good enough?

index.html
sletz: JACK audio use integers for timing values, see: https://github.com/jackaudio/headers/blob/master/types.h

It could be helpful to check what other audio systems (like CoreAudio) do:

CoreAudio here: https://developer.apple.com/documentation/coreaudio/audiotimestamp

So "unsigned long long" possibly makes more sense ?

rtoy: Changed to unsigned long long.  We are now in an interesting situation where the current frame has higher resolution than current time.  But I'm not worried since even at 384 kHz, the max frame is like 1.5 million years from now.

hoch:
About keeping both properties here - I'll leave it up to @padenot.

index.html
```diff
@@ -18208,11 +18209,26 @@ <h3>
...
+                </p>
+              </dd>
+              <dt>
+                <code><dfn>currentFrame</dfn></code> of type <span class=
+                "idlAttrType"><code>unsigned long long</code></span>, readonly
+              </dt>
+              <dd>
+                The current frame number of the context for the block of audio
+                being processed. This is very closely related to
+                <a>currentTime</a>.
```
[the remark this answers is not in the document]
rtoy: Yeah, the reason I left currentTime in is because no where in the spec do we talk about currentFrame. By leaving currentTime in, and relating the two, we don't have to define currentFrame except here.  I suppose we could relate currentFrame to context.currentTime elsewhere and remove it from the global scope.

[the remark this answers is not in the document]
rtoy: @padenot: Anything that needs to be done here?

rtoy:
PTAL.

The original text was really messed up; this cleans up the text defining currentFrame.

Also added a note to explain why this exists.

index.bs
```diff
@@ -9345,6 +9346,19 @@ interface AudioWorkletGlobalScope : WorkletGlobalScope {
 Attributes</h5>
 
 <dl dfn-type=attribute dfn-for="AudioWorkletGlobalScope">
+	: <dfn>currentFrame</dfn>
+	::
+		The current frame of the block of audio being
+		processed. This must be equal to {{AudioWorkletGlobalScope/currentTime}}
+		multiplied by {{AudioWorkletGlobalScope/sampleRate}}.
+
+		Note: Implementations typically count time by frames
```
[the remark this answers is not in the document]
rtoy: Yes, of course. Let me look to see what changes would be needed.  If it's not too much, then this is definitely the way to go.  I was just afraid we'd miss of places and have a spec that was just gratuitously incomplete or incorrect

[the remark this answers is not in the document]
rtoy: Think we just need to add an internal `[[current frame]]` to https://webaudio.github.io/web-audio-api/#rendering-loop, saying `currentTime` is `[[current frame]]` divided by the sample rate.

Will do.

rtoy:
Please take another look.  `currentTime` is defined in terms of an internal `current frame` and that's what `currentFrame` is set to.

PR #1509, opened 2018-02-23
Revise AudioWorkletNode IDL: `AudioWorkletProcessorState` and `onprocessorerror`

### LINKED ISSUE #1451: [audioworklet] Clarify transition between AudioWorkletProcessorState
https://webaudio.github.io/web-audio-api/#idl-def-audioworkletprocessorstate

Currently the spec only defines the meaning of each state. However, it would be useful to clarify the transition between them. This is a typical transition between states.

`pending` -> `running` -> `stopped`

We might want to clarify then this is irreversible; a `running` processor cannot go to the `pending` state again, and a `stopped` processor cannot to back to `pending` or `running` states.

`pending` -> `error` (construction failure)
`pending` -> `running` -> `error` (process callback failure)

Scenarios for the `error` state can be summed up to two cases above. Similarly, once a processor reaches to the `error` state, there is no way to recover.

hoch:
Fixes #1451.

***
<a href="https://pr-preview.s3.amazonaws.com/hoch/web-audio-api/pull/1509.html" title="Last updated on Feb 23, 2018, 8:49 PM GMT (1a90eb5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1509/e79e859...hoch:1a90eb5.html" title="Last updated on Feb 23, 2018, 8:49 PM GMT (1a90eb5)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -18578,17 +18578,16 @@ interface AudioWorkletNode : AudioNode {
             <dl class="attributes" data-dfn-for="AudioWorkletNode"
             data-link-for="AudioWorkletNode">
               <dt>
-                <code><dfn>onprocessorerror</dfn></code> of type
-                <span class=
+                <code><dfn>onprocessorerror</dfn></code> of type <span class=
                 "idlAttrType"><a><code>EventHandler</code></a></span>
               </dt>
               <dd>
                 When an exception is thrown from the processor's
                 <code>constructor</code>, <code>process</code> method, or any
                 user-defined class method, the processor will queue a task on
-                the control thread to fire <a>onprocessorerror</a> event
-                to the node. Note that once an exception is thrown, the
-                processor will output silence throughout its lifetime.
+                the control thread to fire <a>onprocessorerror</a> event to the
+                node. Note that once an exception is thrown, the processor will
+                output silence throughout its lifetime.
               </dd>
               <dt>
                 <code><dfn>parameters</dfn></code> of type <span class=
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -18639,14 +18578,16 @@ <h3>
...
               <dd>
-                Any state change on the processor will queue a task on the
-                control thread to fire <a>onprocessorstatechange</a> event to
-                the node.
+                When an exception is thrown from the processor's
+                <code>constructor</code>, <code>process</code> method, or any
+                user-defined class method, the processor will queue a task on
+                the control thread to fire <a>onprocessorerror</a> event to the
+                node. Note that once an exception is thrown, the processor will
+                output silence throughout its lifetime.
```
rtoy: Is any information about what caused the error included in the event?  That would be great if possible so that the developer has some clue on what happened.

hoch: This should work similarly to `Worker.onerror`.
https://developer.mozilla.org/en-US/docs/Web/API/WorkerGlobalScope/onerror

The EventHandler will fire the callback with `error` object as an argument.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error

The error message can vary between implementations, but the error type must be consistent and it should be specified in the relevant section. (i.e. AWP.constructor or AWP.process)

rtoy:
Closing this PR, which has been replaced by PR #1517 that converts this to bikeshed.

PR #1517, opened 2018-03-19
Fix #1451: AudioWorkletNode IDL: `AudioWorkletProcessorState` and `onprocessorerror`

### LINKED ISSUE #1451: [audioworklet] Clarify transition between AudioWorkletProcessorState

rtoy:
This is a merge of the changes from PR #1509 to a new branch, converting those html changes to bikeshed.

No other editorial changes.

***
<a href="https://pr-preview.s3.amazonaws.com/rtoy/web-audio-api/pull/1517.html" title="Last updated on Mar 27, 2018, 1:06 AM GMT (22fe711)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1517/3cb3931...rtoy:22fe711.html" title="Last updated on Mar 27, 2018, 1:06 AM GMT (22fe711)">Diff</a>


rtoy:
PTAL. I just merged PR #1509 to here so that I can modify it myself.

index.bs
```diff
@@ -3081,16 +3081,12 @@ There are several types of references:
 	<em>playing</em> reference to themselves while they are currently
 	playing.
 
-3. An <dfn>active reference</dfn> for {{AudioWorkletNode}}s
-	whose {{AudioWorkletNode/processorState}}
-	property is set to <code>running</code>.
-
-4. A <em>connection</em> reference which occurs if another
+3. A <em>connection</em> reference which occurs if another
```
hoch: The change in 3084, 3089, 3099 and 3105 do not seem to be related to the AudioWorklet change. Is this intended?

rtoy: It's just a change in the numbers.  Since the original numbered them in numerical order, and this change removed one of the entries, I decided to keep numerical order, without a gap.

Should probably just make them all the same number so we never have to renumber.  It is markdown after all.  But that's for a different PR.

PR #1521, opened 2018-03-20
Fix #1269: Add selectable rate for each AudioParam

### LINKED ISSUE #1269: Allow downgrading of AudioParam automation rate
We are doing a lot of automation control in our Web Audio API based DAW. There can be entire user-configurable processing graphs that ultimately control one or more AudioParam objects at the end of the chain, something like this:

audionodes-screenshot

[Live demo of this app](https://audionodes.com/online/)

-----

We simply call these things Nodes, and the purple color indicates an automation branch (in this screenshot, ultimately the `gain` AudioParam of a GainNode is being automated).

To maintain consistent performance and latency, we are using Web Audio API for our automation everywhere. For example, those values you see (`0.25` and `0.5`) are actually AudioBufferSourceNode instances outputting the given value constantly (as the ConstantSourceNode is not yet widely available), and the multiplier node (the one that says `*`) simply performs a multiplication using a GainNode (one signal is routed directly into the GainNode, while the other signal is routed into the `gain` AudioParam), and the result is finally connected to the `gain` AudioParam.

What we are doing here would be called sample-accurate automation, as this automation is applied to the final `gain` param at a-rate, which is essentially equal to the sample rate.

Now a lot of other, "native desktop" DAW suites offer disabling sample-accurate automation to squeeze out some more performance in heavy projects. I'm wondering if this is something the Web Audio API plans to offer, or already offers in some "hidden" way.

I'm aware there are _k-rate_ AudioParam objects, which are processed in sample _blocks_. This would be nice, but I've been checking out #49, and it seems you don't intend to add a k/a-rate switch (although that issue is 4 years old). This is understandable. However, what about processing AudioParam automation at a lowered sample rate? Is this something that would be feasible? We really don't need to automate a gain at 44.1 kHz, this is pure overkill for most tasks.

rtoy:
For each `AudioParam` add an `automationRate` attribute that can be
set to either "a-rate" or "k-rate".  But some nodes have constraints
on the possible values so enumerate those constraints.

For each `AudioParam`, specify the default initial value of the
`automationRate`.

Also define an "effective automation rate" for a `PannerNode` which
basically means the `AudioParam` is "k-rate" if the `panningModel` is
"HRTF", independent of the actual value of the `automationRate`.

***
<a href="https://pr-preview.s3.amazonaws.com/rtoy/web-audio-api/pull/1521.html" title="Last updated on Apr 3, 2018, 7:50 PM GMT (cba19cf)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1521/fc74f42...rtoy:cba19cf.html" title="Last updated on Apr 3, 2018, 7:50 PM GMT (cba19cf)">Diff</a>


index.bs
```diff
@@ -3189,6 +3189,10 @@ the value of a <dfn id="k-rate">k-rate</dfn> parameter MUST be sampled at the ti
 very first sample-frame, and that value MUST be used for the entire
 block. <dfn id="a-rate">a-rate</dfn> parameters MUST be sampled for
 each sample-frame of the block.
+For many nodes, the rate of the {{AudioParam}} can be selected by
```
hoch: many? some? Not sure.

index.bs
```diff
@@ -3279,6 +3321,50 @@ interface AudioParam {
...
+	: <dfn>automationRate</dfn>
+	::
+		The automation rate for the {{AudioParam}}.  The
+		default value depends on the actual {{AudioParam}};
+		see the description of the {{AudioParam}} for the
+		default value.
+
+		Some nodes have additional <dfn dfn>automation rate constraints</dfn> as follows:
+
+		: {{AudioListener}}
```
hoch: Is this the right place to mention these? Shouldn't we describe them in each node's section?

index.bs
```diff
@@ -5097,10 +5185,12 @@ interface AudioListener {
 <h4 id="AudioListener-attributes">
 Attributes</h4>
 
-For all of the following {{AudioParam}}s, the {{AudioParam}}
-rate is specified by the <dfn>listener AudioParam rate</dfn>
-which is <a>a-rate</a> when any connected {{PannerNode}} is
-<a>a-rate</a> and is <a>k-rate</a> otherwise.
+For all of the following {{AudioParam}}s, the {{AudioParam}} rate is
```
hoch: I had to read this paragraph several times to understand what it means, and I am still not entirely sure.

index.bs
```diff
@@ -5116,7 +5206,7 @@ which is <a>a-rate</a> when any connected {{PannerNode}} is
 			min-notes: Approximately -3.4028235e38
 			max: <a>most-positive-single-float</a>
 			max-notes: Approximately 3.4028235e38
-			rate:
+			rate: "{{AutomationRate/a-rate}}"
```
hoch: Where is this in the rendered page? I can't find these.

index.bs
rtoy: Don't know how best to phrase this.

index.bs
rtoy: I was only following what we already have for other constraints:  https://webaudio.github.io/web-audio-api/#audionode-channelcount-constraints, https://webaudio.github.io/web-audio-api/#audionode-channelcountmode-constraints

index.bs
rtoy: The original was kind of wrong and didn't really spell out what the listener rate was.  Now we are more explicit and say the listener is a-rate if any panner node has an audio param with an effective automation rate of a-rate.  "effective rate" means the k-rate if the panner is HRTF.  Otherwise, it is the automationRate for the param.

index.bs
rtoy: Yeah, the online converter doesn't know that it needs to include some macros, so these don't show up anywhere.  You'll have to run bikeshed locally for this to work.

The alternative is for me to generate and check-in index.html so it can be viewed.  I'll do that.

rtoy:
HTML preview, with everything: https://rawgit.com/rtoy/web-audio-api/1269-selectable-audio-param-rate/index.html

index.bs
padenot: Something like:

> Depending on the `AudioParam` its rate can be controlled via setting the attribute, or it can be fixed, see each `AudioParam`s for details.

index.bs
rtoy: Oh, each `AudioParam` table does mention if there are constraints on the selectable rate and links to here.

index.bs
rtoy: After much discussion off-line, we've agreed this is wrong and should be removed.  Instead, in the `PannerNode`, we should say that if the `panningModel` is "HRTF", the actual automation rate for the `PannerNode` and `AudioListener` is ignored and use "k-rate".  Otherwise, the actual selected automation rate is used.

index.bs
```diff
@@ -3284,13 +3288,55 @@ The following rules will apply when calling these methods:
...
+<pre class="idl">
+[Exposed=Window]
+enum AutomationRate {
+	"a-rate",
+	"k-rate"
+};
+</pre>
+
+<div class="enum-description">
+<table class="simple" dfn-type=enum-value dfn-for="AutomomationRate">
```
hoch: Typo: "Automomation"

index.bs
```diff
@@ -3284,13 +3288,55 @@ The following rules will apply when calling these methods:
 Note: {{AudioParam}} attributes are read only, with the exception
 of the {{AudioParam/value}} attribute.
 
+The automation rate of an {{AudioParam}} can be selected setting the
+{{AudioParam/automationRate}} attribute with one of the following
+values.  However, some {{AudioParam}}s have constraints on whether the
+automation rate can be changed.
+
+<pre class="idl">
+[Exposed=Window]
```
hoch: Do we need `Exposed` for enum?

index.bs
```diff
@@ -3308,6 +3354,45 @@ interface AudioParam {
...
 
 <dl dfn-type=attribute dfn-for="AudioParam">
+	: <dfn>automationRate</dfn>
+	::
+		The automation rate for the {{AudioParam}}.  The
+		default value depends on the actual {{AudioParam}};
+		see the description of the {{AudioParam}} for the
+		default value.
+
+		Some nodes have additional <dfn dfn>automation rate constraints</dfn> as follows:
```
hoch: Typo: `dfn dfn`

index.bs
rtoy: Oops. Probably too much cut'n'paste.  I don't think it's needed; we don't use it anywhere else.

index.bs
rtoy: No, I think that's right.  Without the second "dfn", I think bikeshed thinks it's part of `AudioParam` idl, but it's really just a plain dfn unassociated with anything. (I think that's what Tab told be about other parts in the spec where we have "dfn dfn".

PR #1523, opened 2018-03-20
Have travis install and run bikeshed

rtoy:


rtoy:
WIP. Currently Travis does run the online version and compared the expected
output with the actual output. But the online version doesn't know about
our macro files so those aren't rendered as expected.

PR #1530, opened 2018-03-26
Update CONTRIBUTING.md

ArnaudBienner:
Add information on how to check your changes, and preview the results.

PR #1525, opened 2018-03-23
Fix typo in bypass code example

ArnaudBienner:
I wanted to generate index.html too, but this didn't work.
I believe it could be useful to add a note about how to do this on the contribute page.
Looking at the travis info, I thought the following would work:
curl https://api.csswg.org/bikeshed/ -F file=@index.bs -F force=1 -F output=err -o err
but it failed.
With bikeshed installed locally, it didn't worked either.

***
<a href="https://pr-preview.s3.amazonaws.com/ArnaudBienner/web-audio-api/pull/1525.html" title="Last updated on Mar 23, 2018, 4:04 PM GMT (6707efc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1525/5db4a60...ArnaudBienner:6707efc.html" title="Last updated on Mar 23, 2018, 4:04 PM GMT (6707efc)">Diff</a>

PR #1531, opened 2018-03-26
Use -E instead of -r for sed command, since it is recommended for por…

ArnaudBienner:
…tability

See https://www.gnu.org/software/sed/manual/sed.html#Command_002dLine-Options

@rtoy This is the issue I had I mentioned in #1525.
The command failed because -r wasn't supported for me (on macOS). Strangely, the sed command didn't failed because of unrecognized option when piped to diff.
Looking at the sed documentation, seems like -E is the preferred option.
I've submitted another pull request #1530 to complete CONTRIBUTING.md, with the sed command using -E, but I believe it would be nice to have this script updated as well, since contributors might be likely to look at .travis.yml directly (like I did).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ script:
   # If expected-errs.txt ever needs to be updated, run the following
   # commands, but save the output from sed to expected-errs.txt and
   # check in the file.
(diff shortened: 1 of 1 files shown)


padenot:
It works for me with GNU sed (on an ubuntu machine) and BSD sed (on an OSX machine).

PR #1541, opened 2018-03-28
Update Processing Model for AudioWorklet

### LINKED ISSUE #1511: Reconcile the processing algorithm with the worklet event loop
For now, the AudioWorklet text uses things that need an event loop to work (`postMessage`, `onmessage`, `new Promises`, that kind of thing), but it's not defined how it all works.

It must be defined, otherwise it can't be implemented properly.

For example, does `onmessage` execute before or after the call to `process`? On which thread? When does resolved promises execute?

This was discussed but forgotten about in <https://github.com/WebAudio/web-audio-api/issues/1193#issuecomment-292279108>.

hoch:
- Fixes #1511
- Revise "Processing Model" section by updating the algorithm with AudioWorkletProcessor and microtask checkpoint.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -10400,8 +10400,8 @@ queue</dfn>, that is a list of <dfn lt="control message">control
 messages</dfn> that are operations running on the <a>control
 thread</a>.
 
-<dfn id="queuing" lt="queue a control message">Queuing a control message</dfn> means adding the
-message to the end of the <a>control message queue</a> of an
+<dfn id="queuing" lt="queue a control message">Queuing a control message</dfn>
+means adding the message to the end of the <a>control message queue</a> of an
 {{AudioContext}}.
 
 <a>Control messages</a> in a <a>control message queue</a> are ordered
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -10467,177 +10467,195 @@ asynchronous sections can't be reordered.
...
 
-The algorithm for rendering a block of audio from an
-{{AudioContext}} <em>G</em> with a <a href="#control-message-queue">control message queue</a> <em>Q</em> is as
-follows.
+The algorithm for rendering a block of audio from an {{AudioContext}} <em>G</em>
+with a <a href="#control-message-queue">control message queue</a> <em>Q</em> is
+comprised of 3 steps:
+	1. Process the control message queue.
+	2. Process a render quantum.
+	3. Perform a microtask check point.
```
rtoy: Maybe just use "1" as the item number so it never has to be renumbered if something is added to or removed from the list?

index.bs
```diff
@@ -10467,177 +10467,195 @@ asynchronous sections can't be reordered.
...
 
-If the algorithm returns true then it MUST be executed again in the
-future, to render the next block of audio. Else, the <a>rendering
-thread</a> yields and the processing stops. The <a>control thread</a>
-can restart executing this algoritm if needed.
+The order of these steps cannot be altered or interrupted. If the second step
+returns true then it MUST be executed again in the future, to render the
```
rtoy: This is messy.  Step 2 says "process a render quantum".  There's no concept of returning anything here and for all nodes except a worklet, there's nothing to look at.

index.bs
```diff
@@ -10467,177 +10467,195 @@ asynchronous sections can't be reordered.
...
+	callback passes in a buffer that has to be filled with the audio that will be
+	output. The size of the buffer is often larger than a rendering quantum. In
+	this case, multiple invocations of the rendering algorithm will be called in a
+	rapid succession, in the same callback, before returning. After some time, the
+	underlying audio system will call the callback again, and the algorithm will
+	be executed again. This is an implementation detail that should not be
+	observable, apart from the latency implications.
+
+	{{OfflineAudioContext}} will execute the algorithm continuously, until
+	{{OfflineAudioContext/length}} frames have been rendered.
```
rtoy: Did you really change this in any way other than wrapping lines?  (The preview diff shows no changes.)

Please don't do that, if so, and revert this change.

index.bs
```diff
@@ -10467,177 +10467,195 @@ asynchronous sections can't be reordered.
...
 </div>
 
+The system level audio callback is also queued as a task in the
```
rtoy: Doesn't this mean we now have to define "system level audio callback"?

index.bs
```diff
@@ -10467,177 +10467,195 @@ asynchronous sections can't be reordered.
...
+<a href="#control-message-queue">control message queue</a>. UA must perform the
```
rtoy: "UA must" -> "The UA must".  And maybe "MUST"?

index.bs
```diff
@@ -10467,177 +10467,195 @@ asynchronous sections can't be reordered.
...
+following sequence to process render quanta to fulfill such task by filling up
+the requested buffer size. Along with the primary message queue, the
```
rtoy: "requested buffer size"?  Is that defined anywhere?

index.bs
```diff
@@ -10467,177 +10467,195 @@ asynchronous sections can't be reordered.
...
+<a>rendering thread</a> has another task queue for microtasks for any microtask
+operation such as resolution of {{Promise}} in the
```
rtoy: "{{Promise}}" -> "{{Promise}}s"

index.bs
```diff
@@ -10467,177 +10467,195 @@ asynchronous sections can't be reordered.
...
 
-	6. For each {{DelayNode}} in a cycle, <a href="#available">make
-		available for reading</a> a block of audio from the {{DelayNode}}
-		buffer, available for reading.
+				1. If this {{AudioParam}} has any {{AudioNode}} connected to it,
+					<a href="#SummingJunction">sum</a> the buffers
+					<a href="#available-for-reading"> made available for reading</a> by
+					all {{AudioNode}} connected to this {{AudioParam}},
+					<a href="#down-mix">down mix</a> the resulting buffer down to mono
+					channel, and call this buffer the <dfn id="input-audioparam-buffer">
```
rtoy: Either leave as "mono" or change "mono channel" to "a mono channel".

index.bs
```diff
@@ -10467,177 +10467,195 @@ asynchronous sections can't be reordered.
...
 
-		2. If this {{AudioNode}} has any {{AudioNode}}s connected
-			to its input, <a href="#SummingJunction">sum</a> the buffers
-			<a href="#available">made available for reading</a> by all
-			{{AudioNode}}s connected to this {{AudioNode}}. The
-			resulting buffer is called the <dfn>input buffer</dfn>.
-			<a href="#channel-up-mixing-and-down-mixing">Up or down-mix</a>
-			it to match if number of input channels of this
-			{{AudioNode}}.
+		  5. If this {{AudioNode}} is an {{AudioWorkletNode}}, invoke associated
```
rtoy: I think this should be indented one more level?  That is part of the "For each AudioNode" loop.

index.bs
```diff
@@ -10467,177 +10467,195 @@ asynchronous sections can't be reordered.
...
-		5. Else, <a>process</a> the <a>input buffer</a>,
-			and <a href="#available">make available for reading</a> the resulting buffer.
+		6. <a>Atomically</a> increment {{BaseAudioContext/currentTime}} by 128 /
+			{{BaseAudioContext/sampleRate}}.
 
-	9. <a>Atomically</a> increment {{BaseAudioContext/currentTime}} by 128 / {{BaseAudioContext/sampleRate}}.
+		7. Return true.
 
-	10. Return true.
+	3. <a href="https://html.spec.whatwg.org/multipage/webappapis.html#perform-a-microtask-checkpoint">
```
rtoy: Can this be a bikeshed link?  It would be nice not to have to have an explicit href link here and let bikeshed take care of keeping the link up-to-date.

index.bs
```diff
@@ -10467,177 +10467,195 @@ asynchronous sections can't be reordered.
...
+		Perform a microtask checkpoint</a>.
 </div>
 
-<dfn id="mute">Muting</dfn> an {{AudioNode}} means that its
-output MUST be silence for the rendering of this audio block.
+<dfn id="mute">Muting</dfn> an {{AudioNode}} means that its output MUST be
```
rtoy: Did you just wrap these and following lines?  The preview diff shows no change in text here.  If you did, please undo this.

index.bs
hoch: It's only 3 steps and we should be explicit about it. Numbering all of them into `1` does not really help anything but the editorial convenience.

index.bs
hoch: Did you want something like this? https://github.com/whatwg/html/issues/998

If so, that should not definitely be in this PR.

rtoy: I thought it might already exist.  I was actually thinking that maybe something like `[[html#perform-a-microtask-checkpoint]]` might work (with an appropriate value for `html`.  But you can leave it as is.

index.bs
rtoy: Except when it churns numbers later on.  But you can leave it as is.

index.bs
rtoy: Ok.  But maybe we should add a link (and dfn) for "system level audio callback".  In a different PR.

index.bs
rtoy: It's in a non-normative note, so we don't have to do anything, I think.

index.bs
```diff
@@ -10562,14 +10562,18 @@ Operations that happen <dfn>atomically</dfn> on a
...
-can restart executing this algoritm if needed.
+The algorithm for rendering a block of audio from an {{AudioContext}} <em>G</em>
+with a <a href="#control-message-queue">control message queue</a> <em>Q</em> is
+comprised of 3 steps and the order of these steps cannot be altered or
+interrupted:
+	1. Process the control message queue.
+	2. Process a render quantum.
+	3. Perform a microtask check point.
+
+If the second step returns true then it MUST be executed again in the future, to
```
rtoy: This is kind of inconsistent with what line 10567 says about not being altered or interrupted.

index.bs
```diff
@@ -10588,124 +10592,144 @@ can restart executing this algoritm if needed.
...
-				this {{AudioParam}} for this block.
+			3. If this {{AudioNode}} is a <a>source node</a>,
+				[=Computing a block of audio|compute a block of audio=], and
+				<a href="#available-for-reading">make it available for reading</a>.
 
-			3. <a>Queue a control message</a> to set the
-				{{[[current value]]}} slot of this {{AudioParam}}
-				to the last value computed in the preceding step.
+			4. If this {{AudioNode}} is an {{AudioWorkletNode}}, invoke associated
+				{{AudioWorkletProcessor}}'s {{process}} method to
```
rtoy: You need to fix this bikeshed error about `{{process}}` not existing, which is true. It's not defined by IDL.

index.bs
hoch: I revised the structure to address this.

index.bs

index.bs
hoch: I did what I could. PTAL at l.9812 in the latest patch.

index.bs
```diff
@@ -10574,14 +10574,20 @@ Operations that happen <dfn>atomically</dfn> on a
...
-follows.
-
-If the algorithm returns true then it MUST be executed again in the
-future, to render the next block of audio. Else, the <a>rendering
-thread</a> yields and the processing stops. The <a>control thread</a>
-can restart executing this algoritm if needed.
+The algorithm for rendering a block of audio from an {{AudioContext}} <em>G</em>
+with a <a href="#control-message-queue">control message queue</a> <em>Q</em> is
+comprised of multiple steps and the order of these steps cannot be altered or
+interrupted:
```
rtoy: I think I'd just says "comprised of multiple steps as follows and explained in further detail below in <link to algorithm.>:"  These steps really lack detail but you fill them in in the algorithm below, so linking would be great.

index.bs
```diff
@@ -10574,14 +10574,20 @@ Operations that happen <dfn>atomically</dfn> on a
...
+	1. Prepare the rendering.
+	2. Process the control message queue.
+	3. Process a render quantum.
+	4. Perform a microtask check point.
+	5. Return the result of the rendering.
+
+If the step 5 returns true then it MUST be executed again in the future, to
+render the subsequent block of audio. Otherwise, the <a>rendering thread</a>
+yields and the processing stops. The <a>control thread</a> can restart executing
+these steps if needed.
```
rtoy: If you do link to the algorithm, then this paragraph doesn't really provide an additional detail and isn't needed, right?

index.bs
```diff
@@ -10600,129 +10606,162 @@ can restart executing this algoritm if needed.
...
 <div algorithm="rendering a graph">
-	1. Set the internal slot <dfn attribute
-		for="BaseAudioContext">[[current frame]]</dfn> of the
-		{{BaseAudioContext}} to 0.
-	1. Let <em>Q<sub>rendering</sub></em> be an empty <a>control message
-		queue</a>. <a>Atomically</a> <a>swap</a>
-		<em>Q<sub>rendering</sub></em> and <em>Q</em>
+  1. Prepare the rendering.
+
+		1. Let <em>Q<sub>rendering</sub></em> be an empty <a>control message
```
rtoy: I think bikeshed style is to use `<var>` here instead of `<em>`, here and for all the variables in the algorithm. below.

index.bs
```diff
@@ -10600,129 +10606,162 @@ can restart executing this algoritm if needed.
...
+			queue</a>. <a>Atomically</a> <a>swap</a>
+			<em>Q<sub>rendering</sub></em> and <em>Q</em>.
```
rtoy: Not your fault, but Q seems undefined here.

index.bs
```diff
@@ -10600,129 +10606,162 @@ can restart executing this algoritm if needed.
...
+			2. Set {{BaseAudioContext/currentTime}} to {{[[current frame]]}} divided
+				by {{BaseAudioContext/sampleRate}}.
+
+		7. Set <var>RenderResult</var> to true.
+
+	4. <a href="https://html.spec.whatwg.org/multipage/webappapis.html#perform-a-microtask-checkpoint">
+		Perform a microtask checkpoint</a>.
+
+	5. Return <var>RenderResult</var>.
+	</div>
```
rtoy: This div closes the algorithm div right?  Then I think it should be in column 1, not indented.

index.bs

index.bs

index.bs
hoch: I didn't touch this and want to fix this in a follow up PR if needed.

index.bs
hoch: I didn't touch this and want to fix this in a follow up PR if needed.

index.bs

PR #1550, opened 2018-03-30
Fix #1490: Update bitcrusher example for constant AudioParam values

### LINKED ISSUE #1490: Update bitcrusher example for optimized AudioParams
If (when?) #1464 is fixed, we need to update the bitcrusher demo because it uses `AudioParam`'s.

Similarly if #1480 is fixed, we need to handle that as well.

rtoy:
***
<a href="https://pr-preview.s3.amazonaws.com/rtoy/web-audio-api/pull/1550.html" title="Last updated on Mar 30, 2018, 11:18 PM GMT (728c808)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1550/fc74f42...rtoy:728c808.html" title="Last updated on Mar 30, 2018, 11:18 PM GMT (728c808)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -10174,36 +10174,43 @@ registerProcessor('BitCrusher', class extends AudioWorkletProcessor {
 	process (inputs, outputs, parameters) {
 		let input = inputs[0];
 		let output = outputs[0];
+		let bitDepthLength = parameters.bitDepth.length;
+		let reductionLength = parameters.frequencyReduction.length;
 
-		// Handle the case where the AudioParam has a constant
-		// value, indicated by the fact that the array has length
-		// 1.  This can be done in multiple ways.
+		if (bitDepthLength > 1) {
+			// General case
+			for (let channel = 0; channel &lt; output.length; ++channel) {
+				for (let i = 0; i &lt; output[channel].length; ++i) {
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -10174,8 +10174,22 @@ registerProcessor('BitCrusher', class extends AudioWorkletProcessor {
 	process (inputs, outputs, parameters) {
 		let input = inputs[0];
 		let output = outputs[0];
+
```
hoch: Nothing wrong with this code, but how it's done defeats the purpose of this feature completely. Not sure if this is "exemplary".

rtoy: I can certainly write a bunch of extra code to handle the 4 possible scenarios if you want.

hoch: Using modulo operation can do the trick, I believe.

```js
...
let step = Math.pow(0.5, bitDepth[i % bitDepth.length]);
this._phase += frequencyReduction[i % frequencyReduction.length];
...
```

rtoy:
I won't change the indentation here, but I think since it's in a `<pre>` block, we don't need to use tabs here and can indent with whatever we want and wrap however we want.

PR #1556, opened 2018-04-06
Fix #1553: Add automationRate to AudioParamDescriptor dict

### LINKED ISSUE #1553: AudioWorklet examples should include automationRate
In https://webaudio.github.io/web-audio-api/#defining-a-valid-audioworkletprocessor, there's an example with an AudioParam.  With the introduction of `AudioParam.automationRate` (in #1269 and #1521), this example and others should probably include that.

Also, we need to specify what the default automation rate is for an `AudioParam`s for an `AudioWorkletProcessor`.  For backward compatibility, it should probably be "a-rate".

rtoy:
Also update example to set the automationRate to "k-rate".

***
<a href="https://pr-preview.s3.amazonaws.com/rtoy/web-audio-api/pull/1556.html" title="Last updated on Apr 6, 2018, 3:21 PM GMT (d9157d3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1556/55b069b...rtoy:d9157d3.html" title="Last updated on Apr 6, 2018, 3:21 PM GMT (d9157d3)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3383,13 +3383,13 @@ Attributes</h4>
 		::
 			If the {{PannerNode/panningModel}} is
 			"{{PanningModelType/HRTF}}", the setting of
-			the {{automationRate}} for any
+			the {{AudioParam/automationRate}} for any
 			{{AudioParam}} of the {{PannerNode}} is ignored.
 			Likewise, the setting of the
-			{{automationRate}} for any {{AudioParam}}
+			{{AudioParam/automationRate}} for any {{AudioParam}}
 			of the {{AudioListener}} is ignored.  In this
 			case, the {{AudioParam}} behaves as if the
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -9980,13 +9981,18 @@ dictionary AudioParamDescriptor {
 	float defaultValue = 0;
 	float minValue = -3.4028235e38;
 	float maxValue = 3.4028235e38;
+	AutomationRate automationRate = "a-rate";
```
hoch: So by default, it's `a-rate`?

rtoy: Chrome is a-rate now, so for compatibility, use "a-rate" as the default.

PR #1586, opened 2018-04-27
Fix #1580: connect/disconnect must throw on closed contexts

### LINKED ISSUE #1580: clarify behaviour of (dis)connect() when the corresponding context is closed
Please consider the following example:

```js
audioContext = new AudioContext();
gainNode = audioContext.createGain();

audioContext
    .close()
    .then(() => {
        gainNode.connect(audioContext.destination);
    });
```

This will throw a somewhat cryptic error in Chrome and Opera (`TypeError: Failed to execute 'connect' on 'AudioNode': No function was found that matched the signature provided.`) but it will be executed without any complaints in Firefox and Safari. I don't know much about the implementation details but I guess this is because the resources of the context are already released in Chrome and Opera.

I would propose that calling connect() or disconnect() on an AudioNode which belongs to a closed context should throw an InvalidStateError. This behaviour would be similar with the behaviour of the factory methods.

rtoy:
Update the description of the "closed" state to say you can't connect
or disconnect if the context is closed.  Also change "will" to MUST
for the sentence about creating new nodes.

***
<a href="https://pr-preview.s3.amazonaws.com/rtoy/web-audio-api/pull/1586.html" title="Last updated on Apr 27, 2018, 7:07 PM GMT (6d353cc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1586/d5a5d1c...rtoy:6d353cc.html" title="Last updated on Apr 27, 2018, 7:07 PM GMT (6d353cc)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -683,12 +683,16 @@ enum AudioContextState {
 			<td>
 				This context has been released, and can no longer be used to
 				process audio. All system audio resources have been released.
-				<span class="synchronous">Attempts to create new Nodes on the
-				AudioContext will throw {{InvalidStateError}}</span>.
-				(AudioBuffers may still be created,
-				through {{BaseAudioContext/createBuffer()}},
-				{{BaseAudioContext/decodeAudioData()}},
-				or the {{AudioBuffer}} constructor.)
+				<span class="synchronous">Attempts to create new {{AudioNode}}s on the
+				{{BaseAudioContext}} MUST throw {{InvalidStateError}}</span>.
+				<span class="synchronous">{{AudioNode/connect()|Connecting}} or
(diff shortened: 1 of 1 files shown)


rtoy:
Thanks @mdjp, but I think we want to hold off on committing this until #1580 is clarified a bit more.  There's some opposition on whether this should be done (with even more error cases) or just not producing any errors, including creating new nodes.

rtoy:
Based on https://github.com/WebAudio/web-audio-api/issues/1580#issuecomment-395485491 this is incorrect.

Closing.

PR #1594, opened 2018-05-01
Fix AudioWorkletProcessor process function examples: they should retu…

ArnaudBienner:
…rn true so processing will continue

As discussed in #1593.

@hoch I used the same wording you used in [1] which LGTM.

[1]: https://developers.google.com/web/updates/2017/12/audio-worklet#audioworkletprocessorprocess_method

***
<a href="https://pr-preview.s3.amazonaws.com/ArnaudBienner/web-audio-api/pull/1594.html" title="Last updated on May 1, 2018, 10:04 PM GMT (ccbca13)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1594/7776603...ArnaudBienner:ccbca13.html" title="Last updated on May 1, 2018, 10:04 PM GMT (ccbca13)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -9376,6 +9376,9 @@ class BypassProcessor extends AudioWorkletProcessor {
 		let input = inputs[0];
 		let output = outputs[0];
 		output[0].set(input[0]);
+
+		// To keep this processor alive.
+		return true;
 	}
 });
 
@@ -9844,6 +9847,9 @@ class MyProcessor extends AudioWorkletProcessor {
 				output[channel][i] = input[channel][i] * myParam[0];
 			}
 		}
+
+		// To keep this processor alive.
+		return true;
 	}
 }
 </pre>


rtoy:
This looks fine, but I think you're also missing the examples from https://webaudio.github.io/web-audio-api/#AudioWorklet-Examples

And in particular, I'm not sure what the VUMeter process() method is really trying to do by returning `this._volume >= meterMinimum`.

I think there's also a bug here in that if the VUMeter was connected (and computing the volume), and it gets disconnected so `inputs[0].length` is 0, `this._volume` isn't updated anymore, so it holds the previous value. I don't think that's what I want from a VUMeter that has no inputs. (That should probably be fixed in a different PR.  There's probably also a typo in that `this.volume` is assigned a value. It should probably be `this._volume`.)

ArnaudBienner:
I assumed the examples returning something were correct, and focus on the examples not returning anything.
I can have a look at the VUMeter example. I agree it should be a different pull request though.

rtoy:
This looks fine, but I think you're also missing the examples from https://webaudio.github.io/web-audio-api/#AudioWorklet-Examples

And in particular, I'm not sure what the VUMeter process() method is really trying to do by returning `this._volume >= meterMinimum`.

I think there's also a bug here in that if the VUMeter was connected (and computing the volume), and it gets disconnected so `inputs[0].length` is 0, `this._volume` isn't updated anymore, so it holds the previous value. I don't think that's what I want from a VUMeter that has no inputs. (That should probably be fixed in a different PR.  There's probably also a typo in that `this.volume` is assigned a value. It should probably be `this._volume`.)

rtoy:
This PR as it stands is fine.  Let's land this and move on to any other issues.

ArnaudBienner:
@rtoy Sorry I forgot to comment on this before you merged: regarding comment from Karl on #1593 I'm not sure anymore what I changed is correct.

PR #1598, opened 2018-05-03
Fix processorOptions in audioworkletnodeoptions IDL

ArnaudBienner:
I'm starting to implement this on Firefox and have build error because the following '?' was missing.
I'm not an expert on WebIDL, but someone pointed me to [1] which indeed seems to state that '?' is required.

[1]: https://heycam.github.io/webidl/#idl-nullable-type


rtoy:
Yeah, you're right.  We missed this.

PR #1602, opened 2018-05-06
ChannelSplitterNode should receive ChannelSplitterOptions

saschanaz:
https://webaudio.github.io/web-audio-api/#ChannelSplitterNode-constructors

Its constructor should receive a dictionary but instead is receiving an instance of itself.

***
<a href="https://pr-preview.s3.amazonaws.com/saschanaz/web-audio-api/pull/1602.html" title="Last updated on May 6, 2018, 7:41 AM GMT (ee04138)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1602/567584d...saschanaz:ee04138.html" title="Last updated on May 6, 2018, 7:41 AM GMT (ee04138)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -6249,7 +6249,7 @@ desired.
 
 <pre class="idl">
 [Exposed=Window,
- Constructor (BaseAudioContext context, optional ChannelSplitterNode options)]
+ Constructor (BaseAudioContext context, optional ChannelSplitterOptions options)]
 interface ChannelSplitterNode : AudioNode {
 };
 </pre>


saschanaz:
Any idea why the linking errors happen?

rtoy:
Hmm.  I'm not really sure. I'll have to try this out myself and get back to you.

(For the record, there are bugs in bikeshed where it doesn't handle some of the constructs, so we have an expected-results file to make sure we don't add new errors.  it could be that your change added a new error, which we do really want.)

rtoy:
Ran this locally.  I don't see any new additional errors.  Something about travis CI?

rtoy:
Ah, Travis CI is using the very latest version of bikeshed, and I wasn't.  I know see similar issues locally.

rtoy:
Possibly related to an issue where bikeshed no longer recognizes `DOMString` as something that is a known type to bikeshed. I think we'll have to wait and see what happens when that is fixed.

We might also consider using a fixed version of bikeshed for Travis CI and also for local builds.  Maybe.

rtoy:
Bikeshed fixed and no issues now.

PR #1636, opened 2018-05-24
Refine rendering loop algorithm

### LINKED ISSUE #1632: Ill-defined steps and algorithms in rendering loop
In https://webaudio.github.io/web-audio-api/#rendering-loop, step 3.2.2 says `nodes` is the all nodes in the context that are still alive.  Then 3.2.8 says reverse the order of `nodes`.  But there was no order in the original set, so reversing is meaningless.

Perhaps 3.2.8 is supposed to reverse the order of `ordered`, since these have a definite order as given in 3.2.7.  But the next step is reversal, so that's not clear either.  Perhaps 3.2.7 should be indented one level and really be 3.2.6.3 instead.  That makes ordering and reversing sensible.

hoch:
Fixes #1632.

- Fix typo.
- Clarify some variables.

***
<a href="https://pr-preview.s3.amazonaws.com/hoch/web-audio-api/pull/1636.html" title="Last updated on May 24, 2018, 8:10 PM GMT (8ec61f5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1636/8094348...hoch:8ec61f5.html" title="Last updated on May 24, 2018, 8:10 PM GMT (8ec61f5)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -10558,7 +10558,7 @@ filling up the requested buffer size. Along with the primary message queue, the
 operation such as resolution of {{Promise}}s in the {{AudioWorkletGlobalScope}}.
 
 <div id="rendering-initialization" algorithm="initialize rendering loop">
-	The following step MUST be performance once before the rendering loop starts.
+	The following step MUST be performed once before the rendering loop starts.
 
 	1. Set the internal slot <dfn attribute
 		for="BaseAudioContext">[[current frame]]</dfn> of the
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -10558,7 +10558,7 @@ filling up the requested buffer size. Along with the primary message queue, the
 operation such as resolution of {{Promise}}s in the {{AudioWorkletGlobalScope}}.
 
 <div id="rendering-initialization" algorithm="initialize rendering loop">
-	The following step MUST be performance once before the rendering loop starts.
+	The following step MUST be performed once before the rendering loop starts.
```
hoch: Typo.

index.bs
```diff
@@ -10615,21 +10615,20 @@ operation such as resolution of {{Promise}}s in the {{AudioWorkletGlobalScope}}.
 				2. [=Visit=] <var>node</var>.
 
 				<div algorithm="visiting a node">
-					<dfn lt="Visit">Visiting a node</dfn> <var>node</var> means performing
+					<dfn lt="Visit">Visiting a node</dfn> means performing
```
hoch: Since this step includes the recursion, I think it's better to keep it as a separate algorithm.

PR #1651, opened 2018-05-31
Fix #1650: Clean up azimuth/elevation algorithm

### LINKED ISSUE #1650: Azimuth/Elevation algorithm invalid (slightly)
In the [Azimuth/Elevation](https://webaudio.github.io/web-audio-api/#azimuth-elevation) algorithm, we reference things like `panner.positionX`.  This should probably be `panner.positionX.value`.

Probably also want to add some comments that `context` is some previously created context and `panner` is some panner node in the graph.  Otherwise, these are pretty much undefined (but relatively clear meaning).

rtoy:
Basically add `.values` to all of the `AudioParam`s referenced in the
code.  We forgot to update the code when these attributes were changed
to be `AudioParam`s.

Also use xmp markup instead of pre so we don't have to escape special
characters.

Finally used clang-format to indent the code so that it says within 80 characters so you probably don't need a scrollbar to view all of the code.

***
<a href="https://pr-preview.s3.amazonaws.com/rtoy/web-audio-api/pull/1651.html" title="Last updated on May 31, 2018, 6:02 PM GMT (00aed82)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1651/fdda95d...rtoy:00aed82.html" title="Last updated on May 31, 2018, 6:02 PM GMT (00aed82)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -11502,14 +11502,17 @@ attribute.
 During audio rendering, a <em>distance</em> value will be calculated
 based on the panner and listener positions according to:
 
-<pre highlight="js">
+<xmp highlight="js" line-numbers>
 function distance(panner) {
-	let pannerPosition = new Vec3(panner.positionX, panner.positionY, panner.positionZ);
-	let listener = context.listener;
-	let listenerPosition = new Vec3(listener.positionX, listener.positionY, listener.positionZ);
-	return pannerPosition.diff(listenerPosition).magnitude;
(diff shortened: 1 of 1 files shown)


rtoy:
The cone gain algorithm has the same problem, so I fixed this up here as well.  The `distance` function was fixed up as well.

Issue #1649, opened 2018-05-30
WaveShaper set curve algorithm is ill-defined


rtoy:
The [algorithm](https://webaudio.github.io/web-audio-api/#dom-waveshapernode-curve) is ill-defined, just like for #1648.  `new curve` can be null.

Copy the algorithm for ABSN for WaveShaper.

rtoy:
PR #1653 fixed this.

PR #1653, opened 2018-05-31
Fix #1648, #1649: refine algorithm for setting buffer and curve

### LINKED ISSUE #1648: Convolver set buffer algorithm is ill-defined
The [algorithm for setting the convolver buffer](https://webaudio.github.io/web-audio-api/#dom-convolvernode-buffer) says in part:

1. Let new buffer be the AudioBuffer to be assigned to buffer.
1. If new buffer is not null and [[buffer set]] is true, throw an InvalidStateError and abort these steps.

The first step says `new buffer` is an `AudioBuffer`.  Hence in step 2, it can't be null.

I think we should basically copy the algorithm for ABSN set buffer; I think that was the original intent when designing the buffer setting algorithms for ABSN and Convolver.

rtoy:
These algorithms for setting the buffer (ConvovlerNode) and the curve
(WaveShaperNode) should be consistent with how the buffer for an
AudioBufferSourceNode is handled.  The original intent was that you
can only set the buffer(curve) just once, but you can set them to null
as many times as you want, but you can't then set it to a non-null
buffer.

***
<a href="https://pr-preview.s3.amazonaws.com/rtoy/web-audio-api/pull/1653.html" title="Last updated on May 31, 2018, 11:06 PM GMT (301542b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1653/5e40eb0...rtoy:301542b.html" title="Last updated on May 31, 2018, 11:06 PM GMT (301542b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -6481,17 +6481,17 @@ Attributes</h4>
 		<div algorithm="ConvolverNode.buffer">
 			To set the {{ConvolverNode/buffer}} attribute, execute these steps:
 
-			1. Let <code>new buffer</code> be the {{AudioBuffer}} to be
+			1. Let <var>new buffer</var> be the {{AudioBuffer}} or <code>null</code> to be
 				assigned to {{ConvolverNode/buffer}}.
 
-			2. If <code>new buffer</code> is not <code>null</code> and
+			2. If <var>new buffer</var> is not <code>null</code> and
 				{{ConvolverNode/[[buffer set]]}} is true, <span class="synchronous">throw an {{InvalidStateError}} and abort
 				these steps</span>.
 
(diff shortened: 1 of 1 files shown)


rtoy:
I was just copying the existing text from the AudioBufferSourceNode. I can change this, but since my current fork is messed up with my auto-deploy I'd prefer to merge this as is and fix it up afterwards if needed.  I don't think the text is wrong but it could, perhaps, be phrased better.

mdjp:
@rtoy I'm happy with that.

rtoy:
Thanks!

rtoy:
#1683 filed to rephrase this if needed.

PR #1656, opened 2018-06-06
Fix #1590: Be more explicit about number of inputs/outputs

### LINKED ISSUE #1590: Valid number inputs/outputs for ChannelMerger and ChannelSplitter?
In both https://webaudio.github.io/web-audio-api/#dom-baseaudiocontext-createchannelmerger and https://webaudio.github.io/web-audio-api/#dom-baseaudiocontext-createchannelsplitter, the spec says the errors are thrown for invalid values for the number of inputs or outputs, respectively.

But I can't find anywhere that says the the valid range is.  

In particular, is 0 allowed? I would say no.  Creating such nodes are pretty much useless.

Also, let's say the browser supports just 32 channels, the minimum required.  Can I create a merger with 33 inputs?  I think that should throw an error because that imply 33 channels in the output, which is not supported.

What about a splitter with 33 outputs?  The splitter is pretty well-defined so that the 33rd output will always be silence because we can't have more than 32 channels in the input.  But maybe for consistency with the merger, it should also throw?

rtoy:
Specify more clearly the valid range for the number of inputs and outputs for createChannelMerger and createChannelSplitter.  The valid range is 1 to the number of supported channels.

***
<a href="https://pr-preview.s3.amazonaws.com/rtoy/web-audio-api/pull/1656.html" title="Last updated on Jun 6, 2018, 7:35 PM GMT (d44b6fd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1656/5e40eb0...rtoy:d44b6fd.html" title="Last updated on Jun 6, 2018, 7:35 PM GMT (d44b6fd)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -885,8 +885,8 @@ Methods</h4>
 		<a>Factory method</a> for a
 		{{ChannelMergerNode}} representing a channel
 		merger. <span class="synchronous">An
-		{{IndexSizeError}} exception MUST be thrown for
-		invalid parameter values.</span>
+		{{IndexSizeError}} exception MUST be thrown if
+		{{BaseAudioContext/createChannelMerger(numberOfInputs)/numberOfInputs}} is less than 1 or is greater than the number of supported channels.</span>
 
 		<pre class=argumentdef for="BaseAudioContext/createChannelMerger(numberOfInputs)">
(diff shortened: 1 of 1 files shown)


karlt:
LGTM, thanks.

rtoy:
Thanks for looking at this @karlt. (I tried adding you as reviewer, but you're not listed there.  So double thanks for noticing this PR.)

PR #1670, opened 2018-06-12
Fix #1599: Mention that some AudioBuffer methods may throw errors

### LINKED ISSUE #1599: Errors on creating AudioBuffer and other nodes?
The spec doesn't say anything about `createBuffer()` or `new AudioBuffer()` failing.  So it seems as if you could say `context.createBuffer(32, 1e9, 96000)`.  This needs to allocate 32 `Float32Array`s of length 1e9 each.  That's about 128 GB of memory.  That seems unlikely.  On chrome and firefox `new Float32Array(1e9)` throws an error because both can't allocate arrays of that size. (FWIW, firefox allows you to create such an audio buffer, but getChannelData fails.  That's weird.  Chrome throws an error.)

Do we need to say that createBuffer can fail in this way? I think other nodes can fail in this way too, but none of the memory that would be needed is exposed to Javascript, so that would probably a normal out-of-memory condition.

rtoy:
Say that `copyToChannel` and `getChannelData` may signal errors if the
internal arrays can't be allocated.

***
<a href="https://pr-preview.s3.amazonaws.com/rtoy/web-audio-api/pull/1670.html" title="Last updated on Jul 11, 2018, 9:12 PM GMT (fd90c6a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1670/f9a90c3...rtoy:fd90c6a.html" title="Last updated on Jul 11, 2018, 9:12 PM GMT (fd90c6a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2283,6 +2283,10 @@ Methods</h4>
 		the specified channel of the {{AudioBuffer}} from the
 		<code>source</code> array.
 
+		A {{RangeError}} may be thrown if
+		{{AudioBuffer/copyToChannel()/source}} cannot be
+		copied to the buffer.
+
 		Let <code>buffer</code> be the {{AudioBuffer}} with
 		\(N_b\) frames, let \(N_f\) be the number of elements in the
 		{{AudioBuffer/copyToChannel()/source}} array, and \(k\) be the value of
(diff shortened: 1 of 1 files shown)


rtoy:
PTAL.  Not really sure what the error type should be (just used Chrome as an example which throws RangeError if the array can't be created).  Also not sure if this should be a non-normative Note.

As mentioned in #1599, I've only added this for `copyToChannel` and `getChannelData`.

padenot:
iirc `UnknownError` is usually thrown in OOM scenarios.

PR #1673, opened 2018-06-15
Add issue templates

rtoy:
Add a basic bug report template and a feature request template.

In particular, the bug report says to use something else if it's not
an issue with the spec itself.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,14 @@
+name: Bug Report
+about: Create a report to fix the specification
+
+**Describe the issue**
+A clear and concise description of what the problem is in the spec.
+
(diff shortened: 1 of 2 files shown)


rtoy:
If this actually works, it should look something like https://github.com/rtoy/webaudio-hacks/issues/new/choose

The actual text is different, though.

rtoy:
Dang.  This didn't actually work (clicking "new issue" doesn't show the chooser box).  Either it got merged to the wrong branch or there's some additional magic that's needed.

rtoy:
I think those files need to be on the master branch.  At least they are on my webaudio-hacks site where it works.

rtoy:
Yes, I'm going to try that today.  But I think the last step isn't really needed (for now), because my test repo still has my Github Pages Source as gh-pages.

I think it might be ok to create master and leave it, but from some reading, we can probably delete gh-pages and master will be used to serve up the pages.  I hope.

PR #1675, opened 2018-06-18
Set up issue templates

rtoy:
I think the problem with the original is that I was missing the `---` markers to tell github about the descriptions of the templates.

Maybe.


rtoy:
Yep.  When I click on "New Issue", I see two templates now. Hurray!

PR #1681, opened 2018-06-21
Rework definition of ABSN output behavior for rate-adjusted duration param

### LINKED ISSUE #1660: duration parameter to AudioBufferSourceNode.start is handled incorrectly
Per the [spec](https://webaudio.github.io/web-audio-api/#dom-audiobuffersourcenode-start):

> The duration parameter describes the duration of the sound (in seconds) to be played. If this parameter is passed, this method has exactly the same effect as the invocation of `start(when, offset)` followed by `stop(when + duration)`.

None of the major browsers implement it this way. If the `playbackRate` is set to anything other than 1, actual duration of the sound emitted is the value of the argument divided by `playbackRate`. The duration applies to the length of the source sample before it is adjusted by the `playbackRate`. This is *not* the same as using `start` without a duration followed by `stop`, as described. The latter will play the sped up or slowed down clip for the correct duration.

Here is a link to a simplified test case:
https://codepen.io/brianchirls/pen/zaKgva?editors=1111

The above applies to recent versions of Firefox, Chrome, Safari and Edge. I don't really mind whether the spec gets changed or the browsers change behavior as long as it's consistent. I figured I'd file here first for discussion rather than try to chase down four browser vendors.

joeberkovitz:
Fixes #1660

***
<a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/pull/1681.html" title="Last updated on Jul 12, 2018, 7:50 PM GMT (a57fe08)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1681/93a16ff...a57fe08.html" title="Last updated on Jul 12, 2018, 7:50 PM GMT (a57fe08)">Diff</a>


karlt:
LGTM, thank you.

index.bs
```diff
@@ -4650,7 +4650,7 @@ Methods</h4>
...
+			duration: The {{AudioBufferSourceNode/start(when, offset, duration)/duration}} parameter describes the duration of sound to be played, expressed as seconds of total buffer content to be output, including any whole or partial loop iterations. The units of {{AudioBufferSourceNode/start(when, offset, duration)/duration}} are independent of the effects of {{AudioBufferSourceNode/playbackRate}}. This behavior contrasts with the absolute time units employed by {{AudioScheduledSourceNode/stop()}}. Thus, a {{AudioBufferSourceNode/start(when, offset, duration)/duration}} of 5 seconds with a playback rate of 0.5 will output 5 seconds of buffer content at half speed, producing 10 seconds of audible output. <span class="synchronous">A {{RangeError}} exception MUST be thrown if <code>duration</code> is negative.</span>
```
rtoy: I don't think we need to say anything about `stop()`.

I'm not really sure how much to say here actually about what happens.  It's probably best to point to the algorithm.  I think what you're saying here is correct and is understandable if we're not looping, but if we are, it's much harder to say clearly in words what happens.

index.bs
```diff
@@ -4816,21 +4816,22 @@ let loopEnd;
...
 // Handle invocation of start method call
-function handleStart(when, pos, duration) {
+function handleStart(when, pos, dur) {
 	if (arguments.length >= 1) {
 		start = when;
 	}
 	offset = pos;
 	if (arguments.length >= 3) {
-		stop = when + duration;
+		duration = dur;
```
rtoy: Need to define `duration`?

joeberkovitz: The `duration` variable is already defined, on line 4819 (same place as `offset`).

index.bs
```diff
@@ -4650,7 +4650,7 @@ Methods</h4>
...
-			duration: The {{AudioBufferSourceNode/start(when, offset, duration)/duration}} parameter describes the duration of the sound (in seconds) to be played. If this parameter is passed, this method has exactly the same effect as the invocation of <code>start(when, offset)</code> followed by <code>stop(when + duration)</code>. <span class="synchronous">A {{RangeError}} exception MUST be thrown if <code>duration</code> is negative.</span>
+			duration: The {{AudioBufferSourceNode/start(when, offset, duration)/duration}} parameter describes the duration of sound to be played, expressed as seconds of total buffer content to be output, including any whole or partial loop iterations. The units of {{AudioBufferSourceNode/start(when, offset, duration)/duration}} are independent of the effects of {{AudioBufferSourceNode/playbackRate}}. For example, a {{AudioBufferSourceNode/start(when, offset, duration)/duration}} of 5 seconds with a playback rate of 0.5 will output 5 seconds of buffer content at half speed, producing 10 seconds of audible output. <span class="synchronous">A {{RangeError}} exception MUST be thrown if <code>duration</code> is negative.</span>
```
rtoy: I think (as mentioned previously), we shouldn't really try to say in words what supposed to happen.  Can we just refer to the algorithm?

index.bs
```diff
@@ -4894,16 +4896,16 @@ function process(numberOfFrames) {
 
 	// Render each sample frame in the quantum
 	for (let index = 0; index &lt; numberOfFrames; index++) {
-		// Check that currentTime is within allowable range for playback
-		if (currentTime &lt; start || currentTime >= stop) {
+		// Check that currentTime and bufferDuration are within allowable range for playback
+		if (currentTime &lt; start || currentTime >= stop || bufferDuration >= duration) {
```
rtoy: I think this is wrong?  Let duration = 129/sampleRate and let the playbackRate = 1.  Since bufferDuration is initialized to 0 in process(), and framesToProcess is 128, bufferDuration is at most 128/sampleRate.  So bufferDuration >= duration is never true.  This means the following lines produce an output.  I think that's wrong.  Maybe this should be bufferTime instead?

karlt: Perhaps `bufferDuration` was intended to be declared outside the function?

index.bs
joeberkovitz: I feel strongly that here (as in many other places in the spec) it's valuable to have a short textual description of what a thing is, as a counterpoint to the algorithm.

index.bs
joeberkovitz: Thanks @karlt for catching my error here. Now fixed.

index.bs
rtoy: I'm only concerned if it's inconsistent with the fine details of the algorithm.  A non-normative note is acceptable to me.

joeberkovitz: I think it's consistent -- if you would like to rephrase it somehow, please propose (or commit) some alternate wording.

index.bs
rtoy: Ack.

index.bs
```diff
@@ -4894,16 +4896,16 @@ function process(numberOfFrames) {
...
 			output.push(0); // this sample frame is silent
 			currentTime += dt;
 			continue;
 		}
 
 		if (!started) {
 			// Take note that buffer has started playing and get initial playhead position.
-			bufferTime = offset + ((currentTime - start) * computedPlaybackRate);
+			bufferTime = offset + bufferDuration;
```
rtoy: Is this right?  The initial value for bufferTime is now different between this version and the old version. I think the original is the right value.

index.bs
```diff
@@ -4816,21 +4816,23 @@ let loopEnd;
...
 // Variables for the node's playback parameters
-let start = 0, offset = 0; // Set by start()
-let stop = Infinity; // Set by stop(), or by start() with a supplied duration
+let start = 0, offset = 0, duration = Infinity; // Set by start()
+let stop = Infinity; // Set by stop()
+
 
 // Variables for tracking node's playback state
 let bufferTime = 0, started = false, enteredLoop = false;
+let bufferDuration = 0;
```
rtoy: Can we rename this?  Maybe `playbackDuration` or something? I find it confusing because there's `duration`, `buffer.duration`, and now `bufferDuration`.

index.bs
rtoy: As stated, I think a non-normative note would be fine.  And I think the duration here also applies to looping.  The looping playback will play for duration secs, taking into account the playback rate.

index.bs
joeberkovitz: You're correct about the inconsistency. But another way to go would be to initialize `bufferDuration` to `(currentTime - start) * computedPlaybackRate` in this clause. Wouldn't that be an even better fix? It would cause `bufferDuration` to respect fractional-frame start times.

index.bs
joeberkovitz: Sure, `playbackDuration` seems OK.

joeberkovitz:
What was the "Ack" comment about above - can you clarify? We need a temporary variable name for the value that gets stored in `duration`. How would you like this to look?

index.bs
joeberkovitz: Actually, I am now remembering that I called it `bufferDuration` because it is duration in terms of the buffer frame rate, _not_ the audible playback duration.

index.bs
karlt: I noticed this change too, but thought it was fixing a bug.

For the `when` parameter, "If 0 is passed in for this value or if the value is
less than currentTime, then the sound will start playing immediately."  Also,
"The `offset` parameter supplies a playhead position where playback will begin."
I expect playback to begin with the head at `offset` regardless of `when`.

I couldn't make sense of ((currentTime - start) * computedPlaybackRate).
It seems to be aiming to make playback start with the head at some point
after `offset` if `when` is less than `currentTime`.

(The spec text is helpful to indicate the intention of the algorithm.)

rtoy: Yeah, I didn't look to see what the original is doing, but I agree with @karlt: playback should start at `offset` independent of `when`.  I think `bufferDuration` should actually start at 0.  If you choose any other value, then the actual playback duration will be less then the `duration` arg, even if playback rate is 1.

index.bs
rtoy: So, what should it be?  I think that it's now really the audible playback duration, right?

rtoy:
By "Ack", I meant that you're right and `duration` is already defined; I just missed it.

index.bs
karlt: This PR is changing `duration` from a measurement in the audible timeline to a measurement across the buffer. i.e. the same units as `offset`.

rtoy: Now I'm thoroughly confused.  I thought if `start(when, offset,1)` were called and the playback rate is 1/10, the audio would be heard for 10 sec and not 1 sec.  (Assuming there's at least 1 sec of audio in the buffer, or we're looping.)

karlt: Yes, for hearing the 10s.  playback would end when bufferDuration is 1.

Perhaps `bufferTimeElapsed`?

index.bs
karlt: `bufferDuration` starts at 0.  Simply initializing `bufferTime` to `offset` would be more clear.

index.bs
joeberkovitz: @rtoy we agree on the meaning.  I'll let you modify the branch as heading out of touch for 2 weeks.

index.bs
joeberkovitz: @rtoy could you please change this as per @karlt's last suggestion? I will not be able to work on this for a couple of weeks and this PR is close to done.

index.bs
rtoy: @joeberkovitz I'll take over this PR.  Thanks for your work on this and see in a few weeks.

I think there's just a couple of minor things that need to be fixed:

* Rename `bufferDuration`
* Set `bufferTime` probably to just `offset`.

rtoy:

rtoy:
@joeberkovitz is an invited expert and former chair of the working group.  The changes here are minor tweaks to the existing algorithm that @joeberkovitz actually wrote.

index.bs
rtoy: Renamed `bufferDuration` and initialized `bufferTime` to `offset`.

PR #1701, opened 2018-07-18
Adds "MyProcessor" code example back again

### LINKED ISSUE #1679: Example 12 misses required parameterDescriptors static property
**Describe the issue**

[1.33.4.3](https://webaudio.github.io/web-audio-api/#AudioWorkletProcessor-methods) requires `parameterDescriptors` static property for subclasses but [the example 12](https://webaudio.github.io/web-audio-api/#AudioWorklet-concepts) lacks it.

>The subclass MUST define a method named process() that implements the audio processing algorithm and have a valid static property named parameterDescriptors which is an iterable of AudioParamDescriptor that is looked up by the AudioWorkletProcessor constructor to create instances of AudioParam in the parameters maplike storage in the node.

>```webidl
>// bypass-processor.js script file, runs on AudioWorkletGlobalScope
>class BypassProcessor extends AudioWorkletProcessor {
>  process (inputs, outputs) {
>    // Single input, single channel.
>    let input = inputs[0];
>    let output = outputs[0];
>    output[0].set(input[0]);
>
>    // Process only while there are inputs.
>    return false;
>  }
>});
>
>registerProcessor('bypass-processor', BypassProcessor);
>```

(0af0f3dadbc0ed9dbee867328b69bff21dd01d4c somehow removed the MyProcessor example so the BypassProcessor is the only subclassing example)

hoch:
Fixes #1679.

- Adds `MyProcessor` example back again.
- Adds more comments on the "k-rate" specifier.
- A minor fix: adds "may" in the paragraph to clarify `parameterDescriptors` can be omitted.

***
<a href="https://pr-preview.s3.amazonaws.com/hoch/web-audio-api/pull/1701.html" title="Last updated on Jul 18, 2018, 5:41 PM GMT (117b511)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1701/e8171b6...hoch:117b511.html" title="Last updated on Jul 18, 2018, 5:41 PM GMT (117b511)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -9930,7 +9930,9 @@ class MyProcessor extends AudioWorkletProcessor {
 		let output = outputs[0];
 		let myParam = parameters.myParam;
 
-		// A simple amplifier for single input and output.
+		// A simple amplifier for single input and output. Note that the
+		// automationRate is "k-rate", so it will have a single value at index [0]
+		// for each render quantum.
 		for (let channel = 0; channel &lt; output.length; ++channel) {
 			for (let i = 0; i &lt; output[channel].length; ++i) {
 				output[channel][i] = input[channel][i] * myParam[0];
--- a/index.bs
+++ b/index.bs
@@ -9912,35 +9912,38 @@ validity of a given {{AudioWorkletProcessor}} subclass.
 		</div>
 </dl>
 
-<pre class="example" highlight="js" title="Subclassing AudioWorkletProcessor">
+The example below shows how {{AudioParam}} can be defined and used in an
+{{AudioWorkletProcessor}}.
+
+<xmp class="example" highlight="js" title="Subclassing AudioWorkletProcessor">
 class MyProcessor extends AudioWorkletProcessor {
-	static get parameterDescriptors() {
-		return [{
-			name: 'myParam',
-			defaultValue: 0.5,
-			minValue: 0,
-			maxValue: 1,
-			automationRate: "k-rate"
-		}];
-	}
+  static get parameterDescriptors() {
+    return [{
+      name: 'myParam',
+      defaultValue: 0.5,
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -9911,7 +9911,37 @@ validity of a given {{AudioWorkletProcessor}} subclass.
 			<em>Return type:</em> {{boolean}}
 		</div>
 </dl>
-	
+
+<pre class="example" highlight="js" title="Subclassing AudioWorkletProcessor">
```
rtoy: Bikeshed lets you use `<xmp>` instead of `<pre>` so you can use `<` directly instead of using the html entity.

index.bs
```diff
@@ -9911,7 +9911,37 @@ validity of a given {{AudioWorkletProcessor}} subclass.
 			<em>Return type:</em> {{boolean}}
 		</div>
 </dl>
-	
+
```
rtoy: You should have an introductory sentence here just to make clear the separation between the example and the parameters of the process() method.  Now it just reads funny.

index.bs
```diff
@@ -9911,7 +9911,37 @@ validity of a given {{AudioWorkletProcessor}} subclass.
...
+<pre class="example" highlight="js" title="Subclassing AudioWorkletProcessor">
+class MyProcessor extends AudioWorkletProcessor {
+	static get parameterDescriptors() {
```
rtoy: Not required to change, but it would be nice if we didn't indent by 8.  The rendered output indents by two anyway, so might as well make the source code look more like the final output.

index.bs

index.bs

index.bs

PR #1703, opened 2018-07-19
Fix #1698: Restrict nominal range of OscillatorNode.detune

### LINKED ISSUE #1698: Allowed range for OscillatorNode.detune
According to https://webaudio.github.io/web-audio-api/#OscillatorNode-attributes, the valid range for `detune` is all finite values.  But the `frequency` attribute is limited to Nyquist frequency.  This seems odd  because you can specify a `detune` value that results in a frequency well beyond Nyquist.  

Perhaps we should make the nominal range something like 1200 * log2(2 * Nyquist) so that the detuned frequency stays within Nyquist.  (The extra factor of 2 so that if `frequency` = -Nyquist, we can detune it to produce +Nyquist.)  But also, detune forms a compound parameter whose nominal range is +/- Nyquist anyway.  Still, it's nice to have a limited range for `detune` when it would cause clipping anyway.

rtoy:
The current range for the detune attribute is way too wide because the
computed oscillator frequency needs to be within the Nyquist
frequency.

The new limits allow the detune parameter to combine with any valid
frequency to result in a frequency that still covers the entire
frequency range.  Thus you can detune a frequency of -Nyquist to
Nyquist if so desired.

***
<a href="https://pr-preview.s3.amazonaws.com/rtoy/web-audio-api/pull/1703.html" title="Last updated on Jul 19, 2018, 5:24 PM GMT (2c0b0a0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1703/cf0ad00...rtoy:2c0b0a0.html" title="Last updated on Jul 19, 2018, 5:24 PM GMT (2c0b0a0)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -7979,16 +7979,19 @@ Attributes</h4>
 		{{OscillatorNode/frequency}} by the given amount. Its default
 		<code>value</code> is 0. This parameter is <a>a-rate</a>. It
 		forms a <a>compound parameter</a> with {{OscillatorNode/frequency}}
-		to form the <a>computedOscFrequency</a>.
+		to form the <a>computedOscFrequency</a>.  The nominal
+		range listed below allows this parameter to detune the
+		{{OscillatorNode/frequency}} over the entire possible
+		range of frequencies.
 
 		<pre class=include>
 		path: audioparam.include
 		macros:
 			default: 0
-			min: <a>most-negative-single-float</a>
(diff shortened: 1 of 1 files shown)


rtoy:
Before reviewing, first see https://github.com/WebAudio/web-audio-api/issues/1698#issuecomment-406426007; we need to decide what the appropriate limit should be.

PR #1710, opened 2018-07-20
Honor the given value in `outputChannelCount`.

### LINKED ISSUE #1686: Configuring AudioWorklet output channels under-specified
In https://webaudio.github.io/web-audio-api/#configuring-channels-with-audioworkletnodeoptions, an algorithm is specified for the number of output channels.

Consider the scenario where `numberOfInputs` = `numberOfOutputs` = 1 and `outputChannelCount` is given.  From the algorithm, the second item is true.  However, we only say what happens when `outputChannelCount` is not given; we don't say what happens if it is given.  (The last item isn't applicable in this case.)

What is the expected behavior here?

hoch:
PTAL - Fixes #1686.

Nothing controversial here, but I tried to minimize the text change. I also wanted to add @karlt as a reviewer, but apparently he is not on the list yet.

cc: @karlt

***
<a href="https://pr-preview.s3.amazonaws.com/hoch/web-audio-api/pull/1710.html" title="Last updated on Aug 22, 2018, 8:42 PM GMT (3860048)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1710/9e6efaf...hoch:3860048.html" title="Last updated on Aug 22, 2018, 8:42 PM GMT (3860048)">Diff</a>


index.bs
```diff
@@ -9733,7 +9733,9 @@ various channel configurations can be achieved.
 	{{AudioWorkletNodeOptions/numberOfOutputs}} = 1
 :: If {{AudioWorkletNodeOptions/outputChannelCount}} is
 	unspecified, the output channel count will match
-	<a>computedNumberOfChannels</a> from the input.
+	<a>computedNumberOfChannels</a> from the input. If
+	{{AudioWorkletNodeOptions/outputChannelCount}} is specified
+	with one value, use the value to configure the output channel.
```
rtoy: I think you also need to say what happens if outputChannelCount is specified with more than one value.

index.bs
```diff
@@ -9733,7 +9733,9 @@ various channel configurations can be achieved.
...
 
 : All other cases
 :: If {{AudioWorkletNodeOptions/outputChannelCount}} is
```
rtoy: Maybe we should add an additional sentence here that if outputChannelCount is specified, the values are used as given? (Not really part of the issue, but it's a useful related clarification.)

index.bs
rtoy: No need to do anything here then.

svgeesus: I like the suggestion of

> "If {{AudioWorkletNodeOptions/outputChannelCount}} is specified with one value, the output channel count will match the given value"

index.bs
```diff
@@ -9733,11 +9733,14 @@ various channel configurations can be achieved.
...
-	unspecified, it will be mono for all outputs.
+	unspecified, it will be mono for all outputs. If specified,
+	the value will be used as given.
```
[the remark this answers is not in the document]

index.bs

hoch:
Could you take another look, @karlt?

karlt:
Looks good.  Thank you for the fine tuning.

PR #1733, opened 2018-08-30
Fix #1728: PannerNode constructor throws errors

### LINKED ISSUE #1728: PannerNode constructor should enforce attribute constraints
https://webaudio.github.io/web-audio-api/#pannernode

Things like the [coneOuterGain](https://webaudio.github.io/web-audio-api/#dom-pannernode-coneoutergain) have setters that enforce that they're within a range, but the constructor doesn't. It probably should (or otherwise define this behavior as a clamp).

rtoy:
If a member of the options dictionary would assign an invalid value,
then the contructor must throw an error of the same type as assigning
the attribute would throw.

***
<a href="https://pr-preview.s3.amazonaws.com/rtoy/web-audio-api/pull/1733.html" title="Last updated on Aug 31, 2018, 6:07 PM GMT (46185f2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1733/2e726d7...rtoy:46185f2.html" title="Last updated on Aug 31, 2018, 6:07 PM GMT (46185f2)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -8334,7 +8334,16 @@ Constructors</h4>
 <dl dfn-type=constructor dfn-for="PannerNode">
 	: <dfn>PannerNode(context, options)</dfn>
 	::
-		Let <var>node</var> be a new {{PannerNode}} object.
+		If {{PannerNode/PannerNode()/options}} is given, check
+		for valid values.  If values are not valid, an error
+		is thrown.  Validity is determined as if the the
+		attribute of the same name as the
+		{{PannerNode/PannerNode()/options}} emmber were set to
+		the given value. If an error would be thrown for when
+		setting this value, the same error MUST be thrown when
+		constructing this node.
+
+		Otherwise, let <var>node</var> be a new {{PannerNode}} object.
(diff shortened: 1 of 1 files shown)


rtoy:
This is kind of terse because I didn't want to spell out every single case.

Also, perhaps it would be better to incorporate this condition in step 4.3 of initializing an object in section http://localhost:7100/#AudioNode-creation.  (We really need links to algorithms!)

index.bs
```diff
@@ -8334,7 +8334,16 @@ Constructors</h4>
 <dl dfn-type=constructor dfn-for="PannerNode">
 	: <dfn>PannerNode(context, options)</dfn>
 	::
-		Let <var>node</var> be a new {{PannerNode}} object.
+		If {{PannerNode/PannerNode()/options}} is given, check
+		for valid values.  If values are not valid, an error
+		is thrown.  Validity is determined as if the the
+		attribute of the same name as the
+		{{PannerNode/PannerNode()/options}} emmber were set to
```
padenot: `s/emmber/member/`

PR #1748, opened 2018-09-13
1747 adding changelog from last WD

mdjp:
***
<a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/pull/1748.html" title="Last updated on Sep 14, 2018, 9:57 AM GMT (5a1e1a3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1748/357cfcf...5a1e1a3.html" title="Last updated on Sep 14, 2018, 9:57 AM GMT (5a1e1a3)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -12044,7 +12044,7 @@ Change Log</h2>
 * Add images for convolution doc
 * Initialize bufferDuration outside of process() in ABSN algorithm.
 * Rework definition of ABSN output behavior to account for playbackRate’s interaction with the start(…duration) argument.
-* Add mention of <video> in ultrasonic attack surface.
+* Add mention of video element in ultrasonic attack surface.
 
 <h3 id="changestart2">
 Since Working Draft of 08 December 2015</h3>
--- a/index.bs
+++ b/index.bs
@@ -12031,7 +12031,7 @@ class Vec3 {
 
 <h2 id="changelog">
 Change Log</h2>
-<h3 id="changestart">
+<h3 id="changestart1">
 * Since Working Draft of 19 June 2018</h3>
 * Update implementation-report.html
 * Widen range for detune values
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -12031,7 +12031,22 @@ class Vec3 {
 
 <h2 id="changelog">
 Change Log</h2>
-<h3 id="changestart">
+<h3 id="changestart1">
+Since Working Draft of 19 June 2018</h3>
+* Update implementation-report.html
+* Widen range for detune values
```
rtoy: This is actually restricting the range. Previously it was +/- inf (exclusive), the we reduced it to a much smaller range, and but then made it wider to cover more.  So compared to before, it's a restriction of the range.

index.bs
```diff
@@ -12031,7 +12031,22 @@ class Vec3 {
...
+* PannerNode constructor throws errors
+* Rephrase algorithm for setting buffer and curve
+* Refine startRendering algorithm.
+* Spec more precisely events overlapping with SetValueCurveAtTime
```
rtoy: "Spec more precisely events" reads better as
"Spec, more precisely, events" or "Spec events ... SetValueCurveAtTime more precisely".

Maybe we shouldn't say "Spec" but "Specify"

index.bs
```diff
@@ -12031,7 +12031,22 @@ class Vec3 {
...
+* Add implementation report to gh-pages
+* Honor the given value in `outputChannelCount`
+* Add images for convolution doc
```
rtoy: The convolution doc isn't part of the spec, so I'd leave this out.(If by convolution doc you mean convolution.html)

svgeesus: Right, this is in our repo but doesn't get copied over when the spec is published

mdjp:
Ok, so more haste less speed, I rushed this to get it in before the meeting - can you tell?

index.bs
mdjp: yes, this one crept through my filter.

index.bs
mdjp: agreed, just copying from the git logs, guess I should have read some of them a bit more closely

PR #1750, opened 2018-09-14
Clarify that one must always keep around the last max-fft-size frames.

### LINKED ISSUE #1746: AnalyserNode: Clarify how much time domain data we're supposed to keep around
Currently, according to the spec, changing fftSize resets the smoothing data.

However, nothing is mentioned of the [time domain data](https://webaudio.github.io/web-audio-api/#current-time-domain-data).

We can do one of three things here:

 - Reset the time domain data whenever fftSize is changed. This does mean that we'll have a sharp discontinuity which will take some time to subside. This is pretty efficient, however.
 - Only store `fftSize` sample frames of time domain data. If `fftSize` is increased, fill with silence. This also has a discontinuity problem, but less of one -- if we jump from 1024 to 2048, we still get to keep most of our history
 - Store 32768 (the max value of `fftSize`) sample frames of time domain data. This means we never have a discontinuity problem, but it also means that each analyser node compulsorily to allocate a megabyte of memory. This is what Firefox currently does (but it would be nice to change it)

cc @rtoy @padenot

Manishearth:
r? @rtoy

fixes #1746

***
<a href="https://pr-preview.s3.amazonaws.com/Manishearth/web-audio-api/pull/1750.html" title="Last updated on Sep 14, 2018, 5:48 AM GMT (a0dcf9f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1750/9c67621...Manishearth:a0dcf9f.html" title="Last updated on Sep 14, 2018, 5:48 AM GMT (a0dcf9f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4117,6 +4117,13 @@ Attributes</h4>
 		used for <a href="#smoothing-over-time">smoothing over time</a>
 		is set to 0 for all \(k\).
 
+		Note that increasing {{AnalyserNode/fftSize}} does mean that the
+		<a>current time-domain data</a> must be expanded to include past
+		frames that it previously did not. This means that the
+		{{AnalyserNode}} effectively MUST keep around the last 32768 sample-
+		frames and the <a>current time-domain data</a> is the most recent
+		{{AnalyserNode/fftSize}} sample-frames out of that.
+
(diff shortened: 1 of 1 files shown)


padenot:
Thanks !

PR #1758, opened 2018-09-18
Explicitly include time-domain processing code for BiquadFilterNode

Manishearth:
The spec mentions the transfer function but not the time-domain
equation which is what is more useful for implementors.

It should mention this and also mention how it was obtained.

r? @rtoy

***
<a href="https://pr-preview.s3.amazonaws.com/Manishearth/web-audio-api/pull/1758.html" title="Last updated on Sep 28, 2018, 7:08 PM GMT (c181491)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1758/a225429...Manishearth:c181491.html" title="Last updated on Sep 28, 2018, 7:08 PM GMT (c181491)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5920,15 +5920,24 @@ implement, as they determine the characteristics of the different
 filter types. They are inspired by formulas found in the
 <a href="https://www.w3.org/2011/audio/audio-eq-cookbook.html">Audio EQ Cookbook</a>.
 
-The transfer function for the filters implemented by the
-{{BiquadFilterNode}} is:
+The {{BiquadFilterNode}} processes audio with a transfer function of
 
 <pre nohighlight>
 $$
-H(z) = \frac{\frac{b_0}{a_0} + \frac{b_1}{a_0}z^{-1} + \frac{b_2}{a_0}z^{-2}}
-						{1+\frac{a_1}{a_0}z^{-1}+\frac{a_2}{a_0}z^{-2}}
+ H(z) = \frac{\frac{b_0}{a_0} + \frac{b_1}{a_0}z^{-1} + \frac{b_2}{a_0}z^{-2}}
+                                          {1+\frac{a_1}{a_0}z^{-1}+\frac{a_2}{a_0}z^{-2}}
 $$
 </pre>
+
+which is equivalent to a time-domain equation of:
+
+<pre nohighlight>
+$$
+a_0 y(n) + a_1 y(n-1) + a_2 y(n-2) =
+	b_0 x(n) + b_1 x(n-1) + b_2 x(n-2)
+$$
+</pre>
+
 
 The initial filter state is 0.
 The coefficients in the transfer function above are different for


rtoy:
No problems with including the time-domain equations.  However, while others may disagree, I don't think an explanation of how they're related is needed.  The spec is not a tutorial; it should be written for those "skilled in the art" so you need to know something about DSP and Fourier Transforms and such.

These kinds of things are best left to tutorials and other resources.

rtoy:
Actually, there are lots of different ways to implement a filter and I don't think we want to say this is the recommended way.  I do agree that writing out the time-domain equation is useful, though.

Manishearth:
I left a link to the derivation in an unobtrusive way. Having such breadcrumbs is still useful.

rtoy:
I'm still opposed to these links.  But if someone else wants to approve of them, then I won't object.

index.bs
```diff
@@ -5930,6 +5930,21 @@ H(z) = \frac{\frac{b_0}{a_0} + \frac{b_1}{a_0}z^{-1} + \frac{b_2}{a_0}z^{-2}}
 $$
 </pre>
 
+Transfer functions work as \(Y(z) = H(z) X(z)\), where \(Y(z), X(z)\) are the
+<a href="https://en.wikipedia.org/wiki/Laplace_transform">Laplace transforms</a>
```
rtoy: Well, these aren't Laplace transforms but z-transforms. :-)

index.bs
```diff
@@ -5930,6 +5930,21 @@ H(z) = \frac{\frac{b_0}{a_0} + \frac{b_1}{a_0}z^{-1} + \frac{b_2}{a_0}z^{-2}}
...
+of the input and output respectively.
+
+This <a href="https://en.wikipedia.org/wiki/Infinite_impulse_response#Transfer_function_derivation">
+is equivalent</a> to the following time-domain equation:
+
+<pre nohighlight>
+$$
+y[n] = \frac{b_0}{a_0} x[n] + \frac{b_1}{a_0} x[n - 1] + \frac{b_2}{a_0} x[n - 2]
```
rtoy: This notation doesn't match the time-domain formula for the IIR filter.  Make it consistent.  Actually, I think I'd just copy the IIR result, but expand it out with N = 2 and M = 2.

rtoy:
WDYT @padenot and @hoch?

Manishearth:
Oh, sorry, I left in the mention of laplace transforms. How about now?

index.bs
Manishearth: I feel like the IIR filter should either use subscripts or square brackets, `()` for indexing discrete data is not as common IME

rtoy: This probably depends on the textbooks you're familiar with, and it's very common in Matlab. :-)

Manishearth: Fixed

index.bs
```diff
@@ -5930,6 +5930,17 @@ H(z) = \frac{\frac{b_0}{a_0} + \frac{b_1}{a_0}z^{-1} + \frac{b_2}{a_0}z^{-2}}
...
 </pre>
 
+This <a href="https://en.wikipedia.org/wiki/Infinite_impulse_response#Transfer_function_derivation">
+is equivalent</a> to the following time-domain equation:
+
+<pre nohighlight>
+$$
+y(n) = \frac{b_0}{a_0} x(n) + \frac{b_1}{a_0} x(n - 1) + \frac{b_2}{a_0} x(n - 2)
+         - \frac{a_1}{a_0} y(n - 1) - \frac{a_2}{a_0} y(n - 2)
+$$
```
rtoy: For consistency with the IIR filter, I'd write

a0*y(n)+a1*y(n-1)+a2*y(n-2) = b0*x(n)+b1*x(n-1)*b2*x(n-2)

index.bs
```diff
@@ -5930,6 +5930,17 @@ H(z) = \frac{\frac{b_0}{a_0} + \frac{b_1}{a_0}z^{-1} + \frac{b_2}{a_0}z^{-2}}
 $$
 </pre>
 
+This <a href="https://en.wikipedia.org/wiki/Infinite_impulse_response#Transfer_function_derivation">
+is equivalent</a> to the following time-domain equation:
```
rtoy: I don't think this link is helpful.  If you're going to implement the filters, you have to know a lot already.  And since we never talk about the transfer function again, we could just delete the stuff about the transfer function and just give the time-domain difference equation.

Manishearth: I mean, I was implementing the filters and didn't know this :smile:

Manishearth:
Done.

Instead of making the biquad equation one with the `y`s on the LHS, what if we change the IIR equation to have everything but `y(n)` on the RHS? That feels slightly more relevant. Or we can go the other way.

Manishearth:
Updated, made Biquad match IIR

index.bs
```diff
@@ -5920,16 +5920,16 @@ implement, as they determine the characteristics of the different
...
-H(z) = \frac{\frac{b_0}{a_0} + \frac{b_1}{a_0}z^{-1} + \frac{b_2}{a_0}z^{-2}}
-						{1+\frac{a_1}{a_0}z^{-1}+\frac{a_2}{a_0}z^{-2}}
+y(n) + \frac{a_1}{a_0} y(n-1) + \frac{a_2}{a_0} y(n-2) =
+	\frac{b_0}{a_0} x(n) + \frac{b_1}{a_0} x(n-1) + \frac{b_2}{a_0} x(n-2)
 $$
 </pre>
 
+
 The initial filter state is 0.
 The coefficients in the transfer function above are different for
```
rtoy: My mistake.  We do reference the transfer function here.  Maybe best to leave the transfer function in?  (Sorry about that!)

index.bs
```diff
@@ -5920,16 +5920,16 @@ implement, as they determine the characteristics of the different
...
 
-The transfer function for the filters implemented by the
-{{BiquadFilterNode}} is:
+The {{BiquadFilterNode}} processes audio with a time-domain equation of:
 
 <pre nohighlight>
 $$
-H(z) = \frac{\frac{b_0}{a_0} + \frac{b_1}{a_0}z^{-1} + \frac{b_2}{a_0}z^{-2}}
-						{1+\frac{a_1}{a_0}z^{-1}+\frac{a_2}{a_0}z^{-2}}
+y(n) + \frac{a_1}{a_0} y(n-1) + \frac{a_2}{a_0} y(n-2) =
```
rtoy: This is inconsistent with the IIRFilter. I think it's best to make them consistent.  (Change this or change the IIRFIlter.)

Manishearth: Wait, how? I moved the y(n) to the LHS, I'm not sure how much more consistent you want it. Do you want it to use summation notation?

rtoy: Sorry for being unclear.  The IIRFilter formula is, basically

```
a0*y(n) + a1*y(n-1) + ... = b0*x(n) + b1*x(n-1) + ...
```
You have
```
y(n) + a1/a0*y(n-1) + ... = b0/a0 * x(n-1) + b1/a0*x(n-1) + ...
```
I'm fine with either form. I just think the Biquad and IIR should use the same form for the time-domain equations.

Manishearth: fixed

PR #1768, opened 2018-09-24
fix typo

chrisguttandin:
This pull request simply fixes a typo. In fact it only removes a single character. It causes a weird indentation of the [attributes section](https://webaudio.github.io/web-audio-api/#AudioNode-attributes).

It looks like it was accidentally introduced with this pull request: https://github.com/WebAudio/web-audio-api/pull/1742/files


dontcallmedom:

PR #1770, opened 2018-09-24
Link to correct algorithm for StereoPannerNode, mention algorithm is equal-power

Manishearth:
StereoPannerNode currently links to PannerNode's algorithm, which asks for the computation of the azimuth. This isn't necessary, StereoPannerNode's `pan` argument is the equivalent of the azimuth, in different units.

r? @rtoy

***
<a href="https://pr-preview.s3.amazonaws.com/Manishearth/web-audio-api/pull/1770.html" title="Last updated on Sep 24, 2018, 10:53 PM GMT (113885f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1770/a225429...Manishearth:113885f.html" title="Last updated on Sep 24, 2018, 10:53 PM GMT (113885f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -9034,7 +9034,7 @@ Attributes</h4>
 The {{StereoPannerNode}} Interface</h3>
 
 This interface represents a processing node which positions an
-incoming audio stream in a stereo image using a low-cost <a href="#Spatialization-equal-power-panning">equal-power panning
+incoming audio stream in a stereo image using <a href="#stereopanner-algorithm">a low-cost panning
 algorithm</a>. This panning effect is common in positioning audio
 components in a stereo stream.
 


index.bs
```diff
@@ -11506,7 +11506,7 @@ sound.
 </figure>
 
 <h4 id="stereopanner-algorithm">
-StereoPannerNode Panning</h4>
+StereoPannerNode Panning (equal-power only)</h4>
```
rtoy: Don't think saying equal-power only adds anything.  And nowhere in the algorithm does it say equal-power.

index.bs
```diff
@@ -9034,7 +9034,7 @@ Attributes</h4>
 The {{StereoPannerNode}} Interface</h3>
 
 This interface represents a processing node which positions an
-incoming audio stream in a stereo image using a low-cost <a href="#Spatialization-equal-power-panning">equal-power panning
+incoming audio stream in a stereo image using <a href="#stereopanner-algorithm">a low-cost equal-power panning
```
rtoy: Maybe we should just leave out "equal-power"?  If so, we should also remove "equai-power" from the API Overview for the stereo panner.

index.bs
Manishearth: I mentioned it because it was mentioned in the link, but I'll remove both

Manishearth:
Updated

Issue #1767, opened 2018-09-24
Question about when to check if user's class implements AudioWorkletProcessor


ArnaudBienner:
**Describe the issue**
Currently the spec states that a UA should check if a given processor implements the  AudioWorkletProcessor interface at construction time [[1.33.5]](https://webaudio.github.io/web-audio-api/#instantiation-of-AudioWorkletNode-and-AudioWorkletProcessor)
I wonder why this is not done when calling registorProcessor in the AudioWorkletGlobalScope, since there we already check a bunch of things (existence of processor method, etc.) [1.33.2.2](https://webaudio.github.io/web-audio-api/#AudioWorkletGlobalScope-methods)

I believe it would be better for the user to know right away if a processor is not valid, instead of failing at construction time later on.

**Where Is It**
See links above.

**Additional Information**
I'm not an expert on all those things, so forgive me if I missed the reason why it's like this :)

hoch:
I think this is somewhat related to the [lack of error report mechanism of AudioWorkletGlobalScope](https://github.com/WebAudio/web-audio-api/issues/1282). `registerProcessor()` runs on the worklet thread thus the associated exceptions will be handled by AWGS. Unfortunately AWGS is not equipped with the error handling and user won't get notified by it. It silently fails.

However, we still have the step 6 in `registerProcessor`:
> If the result of IsCallable(argument=Get(O=prototype, P="process")) is false, throw a TypeError and abort these steps.

This will just work when WGS has a proper error reporting channel.

Also the reason why we have another check in the AudioWorkletProcessor instantiation because here we can properly deliver the error in the construction process with `onprocessorerror`. This is a practical approach and it is under our control. (no external dependency)

So I believe the current spec technically doesn't have an error - the problem here is that the context above is not really well explained anywhere.

ArnaudBienner:
OK, thanks for the clarification.
But sorry: I'm not quite convinced though :)
However, I don't think there is any error here: I just believe the current spec is confusing.

* Agreed WGS lacks a proper error mechanism to report, making it impossible to know exactly why addModule would fail.
Since everything logged would end up in the console, it's OK for debugging purpose I guess, but I can imagine usecases where you want to programmatically check the error type if something fails.
But this point should be discussed in #1282 you're right.

* What I don't like here is the incoherence: some errors would be raised when calling addModule. Others when creating the AudioWorkletNode.
1) I'm in favor of raising errors as soon as we can (so in AWGS::registerProcessor).
[...]

PR #1773, opened 2018-09-27
Revise sanity checks on AudioWorkletProcessor.process() method

hoch:
Fixes: #1767

- Removes the `process` function check in AudioWorkletProcessor instantiation.
- Introduce the new `callable process` internal slot to AudioWorkletProcessor.
- Revise "Rendering an Audio Graph" section to add a new check on `process` function.
- Minor editorial fixes. (numbering ordered list)

***
<a href="https://pr-preview.s3.amazonaws.com/hoch/web-audio-api/pull/1773.html" title="Last updated on Oct 10, 2018, 4:16 PM GMT (6e98edd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1773/f9b883c...hoch:6e98edd.html" title="Last updated on Oct 10, 2018, 4:16 PM GMT (6e98edd)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -9848,7 +9848,8 @@ custom audio node. {{AudioWorkletProcessor}} can
 only be instantiated by the construction of an
 {{AudioWorkletNode}} instance. Every
 {{AudioWorkletProcessor}} has an associated <dfn>node
-reference</dfn>, initially null.
+reference</dfn> and <dfn>callable process</dfn> that are
+initially null and false respectively.
 
 <pre class="idl">
 [Exposed=AudioWorklet,
@@ -9918,8 +9919,7 @@ algorithm and may have a valid static property named
 of {{AudioParamDescriptor}} that is looked up by the
 {{AudioWorkletProcessor}} constructor to create instances of
 {{AudioParam}} in the <code>parameters</code> maplike storage
-in the node. The <a href="#check-prototype-type">prototype type check step</a> and the <a href="#check-is-callable">process method is callable step</a> of <a data-link-for="AudioWorkletGlobalScope">registerProcessor()</a> ensure the
-validity of a given {{AudioWorkletProcessor}} subclass.
+in the node.
 
 <dl dfn-type=method dfn-for="AudioWorkletProcessor">
 	: <dfn>process(inputs, outputs, parameters)</dfn>
@@ -9929,13 +9929,17 @@ validity of a given {{AudioWorkletProcessor}} subclass.
 
 		The {{process()}} method is called
 		synchronously by the audio <a>rendering thread</a> at
-		every <a>render quantum</a>, if ANY of the following
-		<dfn dfn>active processing conditions</dfn> are true:
+		every <a>render quantum</a>, if both conditions are met:
 
-		1. The associated {{AudioWorkletProcessor}}'s
-			<a>active source</a> flag is equal to `true`.
+		1. {{callable process}} is `true`.
 
-		2. There are one or more connected inputs to the {{AudioWorkletNode}}.
+		1. <dfn>active processing conditions</dfn> are `true`, meaning
+			ANY of the following states are met:
+
+			* The associated {{AudioWorkletProcessor}}'s
+				<a>active source</a> flag is equal to `true`.
+
+			* There are one or more connected inputs to the {{AudioWorkletNode}}.
 
 		The return value of this method controls the lifetime of the
 		{{AudioWorkletProcessor}}'s associated
@@ -10210,12 +10214,10 @@ thread and the rendering thread.
 	1. Set <var>processor</var>'s <a>node reference</a> to
 		<var>node</var>.
 
+	1. Set {{callable process}} to `true`.
+
 	1. Set <var>node</var>'s <a>processor reference</a> to
 		<var>processor</var>.
-
-	1. <a>Queue a task</a> to the <a>control thread</a> to set the associated
-		{{AudioWorkletNode}}'s state to <code>running</code>, then fire
-		a <code>statechange</code> event.
 </div>
 
 <h4 id="AudioWorklet-Sequence">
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -10196,29 +10196,25 @@ thread and the rendering thread.
...
-	7. Set <var>processor</var>'s <a>node reference</a> to
+	1. Set <var>processor</var>'s <a>node reference</a> to
 		<var>node</var>.
 
-	8. Set <var>node</var>'s <a>processor reference</a> to
+	1. Set <var>node</var>'s <a>processor reference</a> to
 		<var>processor</var>.
 
-	9. <a>Queue a task</a> the <a>control thread</a> to set the associated
+	1. <a>Queue a task</a> the <a>control thread</a> to set the associated
```
rtoy: Since you're modifying this anyway: "Queue a task the" -> "Queue a task to the".

index.bs
```diff
@@ -10798,13 +10794,25 @@ operation such as resolution of {{Promise}}s in the {{AudioWorkletGlobalScope}}.
 				[=Computing a block of audio|compute a block of audio=], and
 				[=Making a buffer available for reading|make it available for reading=].
 
-			4. If this {{AudioNode}} is an {{AudioWorkletNode}}, invoke associated
-				{{AudioWorkletProcessor}}'s {{AudioWorkletProcessor/process()}} method
-				to [=Computing a block of audio|compute a block of audio=] with
-				the argument of [=input buffer=], output buffer and
-				[=input AudioParam buffer=].
+			4. If this {{AudioNode}} is an {{AudioWorkletNode}}, execute these step:
```
rtoy: "step:" -> "steps:"

index.bs
```diff
@@ -10798,13 +10794,25 @@ operation such as resolution of {{Promise}}s in the {{AudioWorkletGlobalScope}}.
...
+				1. Let <var>processor</var> be the associated {{AudioWorkletProcessor}}
+					instance.
+
+				2. Let <var>processFunction</var> be the result of
+					<code><a href="http://www.ecma-international.org/ecma-262/6.0/#sec-get-o-p">
+					Get</a>(O=<i><var>processor</var></i>, P="process")</code>.
+
+				3. If <code><a href="http://www.ecma-international.org/ecma-262/6.0/#sec-iscallable">
+					IsCallable</a>(argument=<i><var>processFunction</var></i>)</code> is
+					false, throw a {{TypeError}}.
```
rtoy: Not ReferenceError if the function doesn't exist?

index.bs

index.bs

index.bs
hoch: I am no expert on ES6 abstract operation, but technically `IsCallable` checks if the target is `function` type. So I think `TypeError` is correct here. 

Example: https://www.ecma-international.org/ecma-262/7.0/#sec-getmethod

index.bs
```diff
@@ -10796,11 +10795,30 @@ operation such as resolution of {{Promise}}s in the {{AudioWorkletGlobalScope}}.
...
 				[=Making a buffer available for reading|make it available for reading=].
 
-			4. If this {{AudioNode}} is an {{AudioWorkletNode}}, invoke associated
-				{{AudioWorkletProcessor}}'s {{AudioWorkletProcessor/process()}} method
-				to [=Computing a block of audio|compute a block of audio=] with
-				the argument of [=input buffer=], output buffer and
-				[=input AudioParam buffer=].
+			4. If this {{AudioNode}} is an {{AudioWorkletNode}}, execute these steps:
+
+			  1. If <a>callable process</a> is `false`, abort the following sub steps.
```
rtoy: I think you need to indent this some more.  I see that this step is labeled 5.

Also, when the following steps are aborted, where do we go from here?  What's the next step in the algorithm?

index.bs
```diff
@@ -10796,11 +10795,30 @@ operation such as resolution of {{Promise}}s in the {{AudioWorkletGlobalScope}}.
...
+
+				1. Let <var>processFunction</var> be the result of
+					<code><a href="http://www.ecma-international.org/ecma-262/6.0/#sec-get-o-p">
+					Get</a>(O=<i><var>processor</var></i>, P="process")</code>.
+
+				1. If <code><a href="http://www.ecma-international.org/ecma-262/6.0/#sec-iscallable">
+					IsCallable</a>(argument=<i><var>processFunction</var></i>)</code> is
+					false, execute the following:
+
+					1. Queue a control message to fire `onprocessorerror` event on the
```
rtoy: It would be nice (but not necessary) if `onprocessorerror` were a link to the attribute.

index.bs
```diff
@@ -10796,11 +10795,30 @@ operation such as resolution of {{Promise}}s in the {{AudioWorkletGlobalScope}}.
...
+						associated AudioWorkletNode.
+
+					1. Set <a>callable process</a> to `false`.
```
rtoy: The algorithm flow is to continue after setting this to false.  Is that what you want?  I think not since `IsCallable` returned false, you shouldn't call the `process` method in the next step.

index.bs
```diff
@@ -9848,7 +9848,8 @@ custom audio node. {{AudioWorkletProcessor}} can
 only be instantiated by the construction of an
 {{AudioWorkletNode}} instance. Every
 {{AudioWorkletProcessor}} has an associated <dfn>node
-reference</dfn>, initially null.
+reference</dfn> and <dfn>callable process</dfn> that are
```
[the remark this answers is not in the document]
rtoy: I don't know.  But we use it all over some for slots like `[[buffer set]]` or `[[internal reduction]]` and these aren't all ECMAScript operations.

Nevertheless, the 'dfn` don't define anything really, so a little bit of text about what it is would be useful.  Or maybe you really just want them to be `<var>`s for the algorithm.  This probably requires a bit more work.

index.bs
```diff
@@ -9858,6 +9856,19 @@ interface AudioWorkletProcessor {
...
+{{AudioWorkletProcessor}} has two internal slots:
+
+<dl dfn-type=attribute dfn-for="AudioWorkletProcessor">
+	: <dfn>[[node reference]]</dfn>
+	::
+		A reference to the associated {{AudioWorkletNode}}.
+
+	: <dfn>[[callable process]]</dfn>
+	::
+		A boolean flag represents whether {{AudioWorkletProcessor/process()}} is
```
rtoy: Probably s/represents/representing/

index.bs
```diff
@@ -9858,6 +9856,19 @@ interface AudioWorkletProcessor {
...
+		a valid function that can ba invoked.
```
rtoy: s/ba invoked/be invoked/

index.bs
```diff
@@ -9929,13 +9941,17 @@ validity of a given {{AudioWorkletProcessor}} subclass.
 
 		The {{process()}} method is called
 		synchronously by the audio <a>rendering thread</a> at
-		every <a>render quantum</a>, if ANY of the following
-		<dfn dfn>active processing conditions</dfn> are true:
+		every <a>render quantum</a>, if both conditions are met:
+
+		* {{{[[callable process]]}} is `true`.
```
rtoy: s/{{{/{{/

index.bs
```diff
@@ -10194,31 +10210,26 @@ thread and the rendering thread.
...
-
-	7. Set <var>processor</var>'s <a>node reference</a> to
+	1. Set <var>processor</var>'s {{[[node reference]]}}to
 		<var>node</var>.
 
-	8. Set <var>node</var>'s <a>processor reference</a> to
-		<var>processor</var>.
+	1. Set {{[[callable process]]}} to `true`.
 
-	9. <a>Queue a task</a> the <a>control thread</a> to set the associated
```
rtoy: Don't need to queue anything anymore?

hoch: No. `statechange` event for AudioWorkletNode is removed. So we don't have to do anything here.

index.bs
```diff
@@ -10796,16 +10807,42 @@ operation such as resolution of {{Promise}}s in the {{AudioWorkletGlobalScope}}.
...
+						IsCallable</a>(argument=<i><var>processFunction</var></i>)</code>.
+
+					1. If {{[[callable process]]}} is `true`,
+						invoke <var>processFunction</var> to
+						[=Computing a block of audio|compute a block of audio=] with the
+						argument of [=input buffer=], output buffer and
+						[=input AudioParam buffer=].
+
+					1. Else if {{[[callable process]]}} is `false`,
+						<a>Queue a control message</a> to fire `onprocessorerror` event on
```
rtoy: Can we specify the type of error that occurred?

hoch: It will be `Event` on the main thread, so it is not an actual error event. You can't throw an error on the main thread, because technically an error on the main thread side. What we can do instead is to throw a `TypeError` in AWGS before queueing a message.

Of course, this won't show up (from user land) unless the console in the browser supports the exception on WorkletGlobalScope.

hoch: I agree, but the `Event` object does not have a valid field to fill with the detailed error message. For the `Event.type`, you'll get `processorerror`.

Also defining a better error reporting for AWP should be a separate piece of work. I've been meaning to do that, but this becomes also useless when you have a proper error handling mechanism in WGS.

PR #1776, opened 2018-10-02
Fix #1481: ABSN outputs mono silence when stopped

### LINKED ISSUE #1481: Tighten spec on ABSN output channels?
This is follow up from https://github.com/WebAudio/web-audio-api/issues/1471#issuecomment-361827095.

I think the proposal there is that ABSNs output exactly one channel of silence until the buffer starts playing and when the buffer stops playing (if ever), the node produces again a single channel of silence.

The difficulty with this is if the ABSN is scheduled to start in the middle of a render quantum.  It seems impossible to change the channel counts at that point, so the count has to happen at the render quantum before or after the the start time.  Stopping the source has exactly the same problem.

I think Chrome sets the output channel count when the AudioBuffer is assigned.

rtoy:
Specify that when an ABSN is done playing, the number of channels in
the output is one, instead of the number of channels in the
`AudioBuffer`.  Update the playback algorithm as well.

***
<a href="https://pr-preview.s3.amazonaws.com/rtoy/web-audio-api/pull/1776.html" title="Last updated on Oct 19, 2018, 6:06 AM GMT (952bcac)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1776/f840c41...rtoy:952bcac.html" title="Last updated on Oct 19, 2018, 6:06 AM GMT (952bcac)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4558,11 +4558,10 @@ The number of channels of the output always equals the number of
 channels of the AudioBuffer assigned to the {{AudioBufferSourceNode/buffer}} attribute, or is one
 channel of silence if {{AudioBufferSourceNode/buffer}} is <code>null</code>.
 In addition, the {{AudioBufferSourceNode}} must output a single
-channel of silence when the end of the
-{{AudioBufferSourceNode/buffer}} has been reached; the
-{{AudioBufferSourceNode/start(when, offset, duration)/duration}} has
-been reached; or the {{AudioScheduledSourceNode/stop(when)/when|stop}}
-time has been reached.
+channel of silence at a rendering quantum boundary after the time at which any one of the following conditions holds:
+	* the end of the {{AudioBufferSourceNode/buffer}} has been reached;
+	* the {{AudioBufferSourceNode/start(when, offset, duration)/duration}} has been reached;
+	* the {{AudioScheduledSourceNode/stop(when)/when|stop}} time has been reached.
 
 A <dfn>playhead position</dfn> for an {{AudioBufferSourceNode}} is
 defined as any quantity representing a time offset in seconds,
@@ -5048,7 +5047,7 @@ function process(numberOfFrames) {
 
 	if (currentTime >= stop) {
 		// End playback state of this node.  No further invocations of process()
-		// will occur.  Set the number of output channels to 1.
+		// will occur.  Schedule a change to set the number of output channels to 1.
 	}
 
 	return output;
--- a/index.bs
+++ b/index.bs
@@ -4557,6 +4557,12 @@ macros:
 The number of channels of the output always equals the number of
 channels of the AudioBuffer assigned to the {{AudioBufferSourceNode/buffer}} attribute, or is one
 channel of silence if {{AudioBufferSourceNode/buffer}} is <code>null</code>.
+In addition, the {{AudioBufferSourceNode}} must output a single
+channel of silence when the end of the
+{{AudioBufferSourceNode/buffer}} has been reached; the
+{{AudioBufferSourceNode/start(when, offset, duration)/duration}} has
+been reached; or the {{AudioScheduledSourceNode/stop(when)/when|stop}}
+time has been reached.
 
 A <dfn>playhead position</dfn> for an {{AudioBufferSourceNode}} is
 defined as any quantity representing a time offset in seconds,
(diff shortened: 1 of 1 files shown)


rtoy:
@karlt I'd like your opinion on this approach.

index.bs
```diff
@@ -5041,8 +5047,8 @@ function process(numberOfFrames) {
 	} // End of render quantum loop
 
 	if (currentTime >= stop) {
-		// end playback state of this node.
-		// no further invocations of process() will occur.
+		// End playback state of this node.  No further invocations of process()
+		// will occur.  Set the number of output channels to 1.
```
karlt: To make it clear that output channel count is not set to 1 immediately, perhaps
drop the last sentence and use something like "will occur, as output for
future rendering quanta will contain only a single channel of silence."

index.bs
```diff
@@ -4557,6 +4557,12 @@ macros:
 The number of channels of the output always equals the number of
```
karlt: This "always" is no longer right, and so would be best removed.

Perhaps something like:
"For rendering quanta in which some playback occurs, the number of channels of
output equals [...].  For rendering quanta in which no playback occurs, the
output is a single channel of silence."

If you prefer something more similar to the text proposed here, then maybe
... "channel of silence for rendering quanta after and not including the
quantum during which the earliest of the following occurs: the end of the" ...

index.bs
rtoy: Chrome can't guarantee that this happens exactly on the rendering quantum after stop.  It may take several renders depending on how all he locks line up.

rtoy:
In general Chrome cannot make the channel count changes at the times you want due to how all the mutexes work to protect access of various state between the audio thread and the main thread.  

Without a major rewrite, I think the best Chrome can do is to change the channel count when `start()` is called (assuming a buffer is already assigned) or when the buffer is assigned (assuming `start()` was already called).

The change to a single channel of silence will happen at some render quantum after the end.  Perhaps the best Chrome can do is the change happens when `onended` would be delivered.

index.bs
karlt: Perhaps something like "Schedule a change in the number of output channels to 1."

index.bs
```diff
@@ -4557,6 +4557,12 @@ macros:
...
 channels of the AudioBuffer assigned to the {{AudioBufferSourceNode/buffer}} attribute, or is one
 channel of silence if {{AudioBufferSourceNode/buffer}} is <code>null</code>.
+In addition, the {{AudioBufferSourceNode}} must output a single
+channel of silence when the end of the
```
karlt: Perhaps s/when/after/ to add a hint that this is not in the middle of a rendering quantum.

karlt:
Thank you for explaining the situation with Chrome.  I think I have a better
understanding now of a model with control-thread decisions on channel count.

For simple GC of an ABSN that has not started, a single channel of silence
before `start()` would be sufficient.  A change in channel count on receipt
of `start()` doesn't seem ideal to me, but I can see that would be a nice
time for an implementation to allocate output buffers.  FWIW multiple output
channels before playback begins would require changes in Gecko.

It would be entirely reasonable to postpone specification of an
ABSN that has not started, so as not to block the solution for the more
comment situation of an ABSN that has ended.

rtoy:
I don't think we need to specify (or require) that multiple channels are output before playback starts.  Thus, both Firefox and Chrome are compliant, even if their behaviors are different.  I think the worst thing that happens is that downstream nodes process more channels than is strictly required.

padenot:
We talked about this last week, I had another look and I'm ok with this, and @karlt told me this morning that he agrees as well. We can merge it probably?

index.bs
karlt: Thank you, @hoch.

```suggestion
The number of channels of the output equals the number of
```

The remaining suggestions here were too precise for what rtoy wanted.

index.bs
```diff
@@ -4557,6 +4557,11 @@ macros:
 The number of channels of the output always equals the number of
 channels of the AudioBuffer assigned to the {{AudioBufferSourceNode/buffer}} attribute, or is one
 channel of silence if {{AudioBufferSourceNode/buffer}} is <code>null</code>.
+In addition, the {{AudioBufferSourceNode}} must output a single
+channel of silence at a rendering quantum boundary after the time at which any one of the following conditions holds:
```
karlt: The single channel of silence is not really *at* the rendering quantum boundary, which has zero length, but *begins* at the boundary.
Perhaps
"In addition, if the buffer has more than one channel, then the {{AudioBufferSourceNode}} output must change to a single channel of silence at a rendering".

hoch:
Merging after approvals from @padenot and @karlt.

PR #1779, opened 2018-10-25
Specify side effect when calling start() when AudioContext was blocked (fixes #1771)

### LINKED ISSUE #1771: Allow UA to resume AudioContext as a side effect of start()
**Describe the feature**
As part of the changes coming from https://github.com/WebAudio/web-audio-api/issues/836 Web Audio received a notion of autoplay. An AudioContext may not switch to `running` state if the UA decides so. During the F2F discussions at TPAC 2 years ago the WG decided to not allow the UA to resume the AudioContext as a side effect of calling `start()` (which Safari iOS does). Chrome has now decided to launch this and it may make sense to integrate this option in the specification to avoid future implementer to run into the same problems.

**Is there a prototype?**
Safari iOS implements this. Chrome has this implemented behind a flag.

**Describe the feature in more detail**
When `start()` is called, if the AudioContext is allowed to run and was prevented from running by the UA previously, the UA could decide to resume it.

mounirlamouri:
…d by autoplay.

***
<a href="https://pr-preview.s3.amazonaws.com/mounirlamouri/web-audio-api/pull/1779.html" title="Last updated on Jan 11, 2019, 4:42 PM UTC (de6eb5f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1779/58b5ddd...mounirlamouri:de6eb5f.html" title="Last updated on Jan 11, 2019, 4:42 PM UTC (de6eb5f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1645,7 +1645,7 @@ Methods</h4>
 			3. If the {{AudioContext}} is not <a>allowed to
 				start</a>, append <var>promise</var> to
 				<a>pendingResumePromises</a>, set the <a>was allowed to start</a> flag
-				associated to the {{AudioContext}} to <code>fase</code>, and abort these
+				associated to the {{AudioContext}} to <code>false</code>, and abort these
 				steps, returning <var>promise</var>.
 
 			4. Set the <em>control thread state</em> flag on the
--- a/index.bs
+++ b/index.bs
@@ -1272,6 +1272,11 @@ the current context. In other words, if the
 allowed to transition from <code>suspended</code> to
 <code>running</code>.
 
+An {{AudioContext}} has an associated <dfn>was allowed to start</dfn> flag,
+initially set to <code>true</code>. The flag, when set to <code>false</code>
+reflects that the {{AudioContext}} was prevented from transitioning to the
+<code>running/<code> state.
+
 Note: For example, a user agent could require that an
 {{AudioContext}} control thread state change to
 running is <a>triggered by user activation</a> (as described in [[HTML]]).
@@ -1362,7 +1367,9 @@ Constructors</h4>
 					amount.
 
 			5. If the {{AudioContext}} is <a>allowed to start</a>, send a
-				<a>control message</a> to start processing.
+				<a>control message</a> to start processing. Otherwise, set the
+				<a>was allowed to start</a> flag associated to the {{AudioContext}} to
+				<code>false</code>.
 
 			6. Return this {{AudioContext}} object.
 		</div>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1272,6 +1272,11 @@ the current context. In other words, if the
 allowed to transition from <code>suspended</code> to
 <code>running</code>.
 
+An {{AudioContext}} has an associated <dfn>was allowed to start</dfn> flag,
```
hoch: Can you explain why we need another flag `was allowed to start`? Can we just look up `is allowed to start` flag on the context when `start()` method is called?

Also it would be better to focus on `start()` and separate out the `resume()` operation into another PR. Or change the title of PR so it can be altogether here.

hoch:
Note: This fixes #1771.

index.bs
mounirlamouri: So sorry, this fell off my radar. The reason why I've added `was allowed to start` is that we need to know whether `start()` needs to resume the AudioContext. If there was no previous failed attempt to resume the AudioContext, we should not attempt to resume it.

hoch: One more question: can we use a document level flag instead of "allowed to start" state in the audio context? (Can't remember the spec proposal)

hoch:
I am afraid we have to resolve #1802 before pushing this PR. We need some checks for user-suspended context before the context can be unblocked. @mounirlamouri, please take a look at #1802 and let me know what you think.

index.bs
```diff
@@ -1637,8 +1644,9 @@ Methods</h4>
 
 			3. If the {{AudioContext}} is not <a>allowed to
 				start</a>, append <var>promise</var> to
-				<a>pendingResumePromises</a> and abort these steps, returning
-				<var>promise</var>.
+				<a>pendingResumePromises</a>, set the <a>was allowed to start</a> flag
+				associated to the {{AudioContext}} to <code>fase</code>, and abort these
```
hoch: s/fase/false

mounirlamouri:
> I am afraid we have to resolve #1802 before pushing this PR. We need some checks for user-suspended context before the context can be unblocked. @mounirlamouri, please take a look at #1802 and let me know what you think.

I replied there but I'm not sure I understand the concern. The "user suspended context" issue only exist if the context was suspended while not running. My solution wouldn't auto-resume the context if it had been running. At least, that was my intent.

hoch:
According to the comment [here](https://github.com/WebAudio/web-audio-api/issues/1802#issuecomment-472927439), WG decided not to pursue this PR. Closing.

PR #1785, opened 2018-11-06
Adjust the rendering algorithm for AudioListener

### LINKED ISSUE #1700: Specify what happens when you connect() to an AudioParam on the AudioListener
The AudioListener has many AudioParams, which you can set timeline events on. This is great.

However, you can also connect the outputs of other nodes to it, via `AudioNode.connect(param)`.

This seems pretty tricky to deal with correctly. The processing model doesn't handle it at all, it just mentions that you should "compute the values" for the params, which isn't possible for when nodes are connected to the params since we haven't started computing the outputs of the nodes yet.

If we were to include this in the graph algorithm there would be significant complexity added -- the cycle breaker algorithm needs to understand the "special" dependency on the listener from panner nodes -- this contributes to cycles but cannot be broken.

Ideally we should make connecting to such params a hard error on the DOM side (or a "does nothing" thing on the audio renderer's side). No browser implements these as params yet, so we should be good for now. I don't really see the value in allowing audio nodes to connect to listener params; this sounds like a recipe for massive vertigo and not much else.

cc @padenot @ferjm

padenot:
This is based off of #1782.

- There now an implicit connection from the AudioListener to all
PannerNode.
- The ordering algorithm now takes into account the AudioNodes connected
to AudioParams.
- There is an implicit connection between an AudioNode/AudioListener and
their AudioParam.

This fixes #1700.

***
<a href="https://pr-preview.s3.amazonaws.com/padenot/web-audio-api/pull/1785.html" title="Last updated on Nov 15, 2018, 10:48 AM GMT (98058c5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1785/816eff2...padenot:98058c5.html" title="Last updated on Nov 15, 2018, 10:48 AM GMT (98058c5)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5544,6 +5544,17 @@ Methods</h4>
 		</pre>
 </dl>
 
+<h4 id="listenerprocessing">
+Processing</h4>
+
+Since the {{AudioListener}} has {{AudioParam}}, and {{AudioNode}} can be
+connected to {{AudioParam}}, it is necessary to have the {{AudioListener}}
+participate in the {{AudioNode}} ordering algorithm.
+
+Additionaly {{PannerNode}} output depend on the value of the {{AudioListener}}'s
+parameter.  For the purpose of computing the ordering in which to process the
+{{AudioNode}}s, all the {{PannerNode}}s in the graph have the {{AudioListener}}
+as input.
 
 <!--
 ████████  ████████   ███████   ██████  ████████  ██████   ██████  ████ ██    ██  ██████
@@ -8838,7 +8849,6 @@ The set of <a href="#panner-channel-limitations">channel
 limitations</a> for {{StereoPannerNode}} also apply
 to {{PannerNode}}.
 
-
 <!-- Big Text: Periodic -->
 
 <h3 interface lt="PeriodicWave">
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -5521,6 +5521,17 @@ Methods</h4>
...
 </dl>
 
+<h4 id="listenerprocessing">
+Processing</h4>
+
+Since the {{AudioListener}} has {{AudioParam}}, and {{AudioNode}} can be
+connected to {{AudioParam}}, it is necessary to have the {{AudioListener}}
+participate in the {{AudioNode}} ordering algorithm.
+
+Additionaly {{PannerNode}} output depend on the value of the {{AudioListener}}'s
```
hoch: Is `PannerNode` here plural or singular?

index.bs
```diff
@@ -5521,6 +5521,17 @@ Methods</h4>
 		</pre>
 </dl>
 
+<h4 id="listenerprocessing">
+Processing</h4>
+
+Since the {{AudioListener}} has {{AudioParam}}, and {{AudioNode}} can be
+connected to {{AudioParam}}, it is necessary to have the {{AudioListener}}
+participate in the {{AudioNode}} ordering algorithm.
```
hoch: How about something like this? [5527-5534]

Because AudioListener's parameters can be connected with AudioNodes and also they affect the output of PannerNodes in the same graph, the node ordering algorithm should take the AudioListener into consideration when computing the order of processing. For this reason, all the PannerNodes in the graph have the AudioListener as input.

index.bs
```diff
@@ -6879,6 +6890,37 @@ Dictionary {{DelayOptions}} Members</h5>
...
 </dl>
 
+<h4 id="DelayNode-processing">Processing</h4>
+
+A {{DelayNode}} has an internal buffer that holds {{DelayNode|delayTime}}
+seconds of audio.
+
+The processing of a {{DelayNode}} is broken down in two parts: writing to the
+delay line, and reading from the delay line. This is done via two internal
+{{AudioNode}}s (that are not accessible to content and exist only to ease
```
hoch: "that are not accessible to content and " seems redundant.

index.bs
```diff
@@ -6879,6 +6890,37 @@ Dictionary {{DelayOptions}} Members</h5>
...
+<a>source node</a>.
+
+When processing a input buffer, a <a>DelayWriter</a> MUST write the audio to the
+internal buffer of the {{DelayNode}}.
+
+When producing an output buffer, a <a>DelayReader</a> MUST yield exactly the audio
+that was written to the corresponding <a>DelayWriter</a> {{DelayNode|delayTime}}
+seconds ago.
+
+Note: This means that channel count changes are reflected after the delay time
```
hoch: Nit: "The channel count change is reflected after..."

index.bs
```diff
@@ -10772,21 +10829,22 @@ operation such as resolution of {{Promise}}s in the {{AudioWorkletGlobalScope}}.
...
+					3. If <var>node</var> is an {{AudioNode}}, [=Visit=] each
+						{{AudioNode}} connected to the input of <var>node</var>.
 
-			7. Add <var>node</var> to the beginning of <var>ordered node list</var>.
+					4. For each {{AudioParam}} <var>param</var> of <var>node</var>:
+						1. For each {{AudioNode}} <var>param input node</var> connected to <var>param</var>:
+							1. [=Visit=] <var>param input node</var>
 
-			8. Reverse the order of <var>ordered node list</var>.
+					5. Add <var>node</var> to the beginning of <var ignore=>ordered node list</var>.
```
hoch: Let's remove `=` after `ignore`.

padenot:
Rebased + review comments addressed.

PR #1787, opened 2018-11-08
Specify rules to determine the output channel count for AudioNodes that have a tail time, when the input channel count changes

### LINKED ISSUE #1719: channel count changes in filter nodes with tail time
Consider a graph where a filter node with a tail-time initially has a mono
input, then a stereo input is added, then the stereo input is removed.

In order to avoid glitches in the audio signal, implementations should take
care to up-mix the tail when the stereo input is added, and continue to
produce stereo output for a the tail period after the stereo input is removed.

It may be helpful to mention this in the spec.

padenot:
This fixes #1719.

***
<a href="https://pr-preview.s3.amazonaws.com/padenot/web-audio-api/pull/1787.html" title="Last updated on Nov 14, 2018, 2:27 PM GMT (ee34370)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1787/1d0eec3...padenot:ee34370.html" title="Last updated on Nov 14, 2018, 2:27 PM GMT (ee34370)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1132,12 +1132,21 @@ Methods</h4>
 			Note: Multiple <em>decoding threads</em> can run in parallel to
 			service multiple calls to <code>decodeAudioData</code>.
 
-			1. Attempt to decode the encoded {{BaseAudioContext/decodeAudioData(audioData, successCallback, errorCallback)/audioData!!argument}} into linear
-				PCM.
+			1. Let <var>can decode</var> be a boolean flag, initially set to true.
 
-			2. If a decoding error is encountered due to the audio format
-				not being recognized or supported, or because of
-				corrupted/unexpected/inconsistent data, then <a>queue a task</a> to
+			2. Attempt to determine the MIME type of
+				{{BaseAudioContext/decodeAudioData(audioData, successCallback,
+				errorCallback)/audioData!!argument}}, using
+				[[mimesniff#matching-an-audio-or-video-type-pattern]]. If the audio or
+				video type pattern matching algorithm returns <code>undefined</code>,
+				set <var>can decode</var> to <em>false</em>.
+
+			3. If <var>can decode</var> is <em>true</em>, attempt to decode the encoded
+				{{BaseAudioContext/decodeAudioData(audioData, successCallback,
+				errorCallback)/audioData!!argument}} into linear PCM. In case of
+				failure, set <var>can decode</var> to <em>false</em>.
+
+			4. If <var>can decode</var> is <em>false</em>, <a>queue a task</a> to
 				execute the following step, on the <a>control thread</a>'s
 				event loop:
 
@@ -1148,7 +1157,7 @@ Methods</h4>
 
 				3. If {{BaseAudioContext/decodeAudioData()/errorCallback!!argument}} is not missing, invoke {{BaseAudioContext/decodeAudioData()/errorCallback!!argument}} with <var>error</var>.
 
-			3. Otherwise:
+			5. Otherwise:
 				1. Take the result, representing the decoded linear PCM
 					audio data, and resample it to the sample-rate of the
 					{{AudioContext}} if it is different from
--- a/index.bs
+++ b/index.bs
@@ -11290,6 +11290,35 @@ defined below, skipping over those channels not present.
 		<tr><td>17 <td>SPEAKER_TOP_BACK_RIGHT <td> <td> <td> <td>
 </table>
 
+<h3 id="channels-tail-time">
+Implication of tail-time on input and output channel count</h3>
+
+When an {{AudioNode}} has a non-null <a>tail-time</a>, and an output channel
+count that depends on the input channels count, any change in input channel
+count MUST take into account the {{AudioNode}}'s <a>tail-time</a>.
+
+When there is a decrease in input channel count, the change in output channel
+count MUST happen after the <a>tail-time</a> has passed.
+
+When there is an increase in input channel count, the behavious depends on the
+{{AudioNode}} type:
+
+- For a {{ConvolverNode}}, since the number of input channel determines the
+	<a href="#Convolution-channel-configurations"> audio processing algorithm </a>
+	to use, it MUST immediately start outputing two channels.
+
+	Note: this only applies to the case where the {{ConvolverNode}}'s impulse
+	response is mono. Otherwise, the {{ConvolverNode}} always outputs a stereo
+	signal regardless of its input channel count.
+
+- For other {{AudioNode}}s that have a <a>tail-time</a>, the number of output
+	channels MUST increase after the <a>tail-time</a> has passed.
+
+Note: Intuitively, this allows not losing stereo information as part of
+processing: when multiple audio blocks of different channel count contribute to
+an output block, then the output block's channel count is a superset of the
+input channel count of the input blocks.
+
 <h3 id="UpMix-sub">
 Up Mixing Speaker Layouts</h3>
 
--- a/index.bs
+++ b/index.bs
@@ -11305,13 +11305,12 @@ longer affects the output.
 When there is an increase in input channel count, the behavior depends on the
 {{AudioNode}} type:
 
-- For a {{DelayNode}} or a {{DynamicsCompressorNode}},
-	<a>tail-time</a>, the number of output channels MUST increase after the
-	<a>tail-time</a> has passed.
-
-- For other {{AudioNode}}s that have a <a>tail-time</a>, the number of output
+- For a {{DelayNode}} or a {{DynamicsCompressorNode}}, the number of output
 	channels MUST increase when the input that was received with greater channel
 	count begins to affect the output.
+
+- For other {{AudioNode}}s that have a <a>tail-time</a>, the number of output
+	channels MUST increase immediately.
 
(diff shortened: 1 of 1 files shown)


padenot:
@karlt, can you have a look at this?

index.bs
```diff
@@ -11290,6 +11290,35 @@ defined below, skipping over those channels not present.
 		<tr><td>17 <td>SPEAKER_TOP_BACK_RIGHT <td> <td> <td> <td>
 </table>
 
+<h3 id="channels-tail-time">
+Implication of tail-time on input and output channel count</h3>
+
+When an {{AudioNode}} has a non-null <a>tail-time</a>, and an output channel
```
hoch: Is `non-null` a common spec language?

index.bs
```diff
@@ -11290,6 +11290,35 @@ defined below, skipping over those channels not present.
...
+count that depends on the input channels count, any change in input channel
+count MUST take into account the {{AudioNode}}'s <a>tail-time</a>.
```
hoch: Perhaps:

`...MUST take the {{AudioNode}}'s <a>tail-time</a> into account.`

index.bs
```diff
@@ -11290,6 +11290,35 @@ defined below, skipping over those channels not present.
...
+
+When there is a decrease in input channel count, the change in output channel
+count MUST happen after the <a>tail-time</a> has passed.
+
+When there is an increase in input channel count, the behavious depends on the
+{{AudioNode}} type:
+
+- For a {{ConvolverNode}}, since the number of input channel determines the
+	<a href="#Convolution-channel-configurations"> audio processing algorithm </a>
```
hoch: nit: redundant spaces within tags.

index.bs
```diff
@@ -11290,6 +11290,35 @@ defined below, skipping over those channels not present.
...
+	to use, it MUST immediately start outputing two channels.
+
+	Note: this only applies to the case where the {{ConvolverNode}}'s impulse
+	response is mono. Otherwise, the {{ConvolverNode}} always outputs a stereo
+	signal regardless of its input channel count.
+
+- For other {{AudioNode}}s that have a <a>tail-time</a>, the number of output
+	channels MUST increase after the <a>tail-time</a> has passed.
+
+Note: Intuitively, this allows not losing stereo information as part of
```
hoch: I am not sure if this section is normative or not. If so, we can block this paragraph as an informative note.

index.bs
```diff
@@ -11290,6 +11290,35 @@ defined below, skipping over those channels not present.
...
+processing: when multiple audio blocks of different channel count contribute to
+an output block, then the output block's channel count is a superset of the
```
hoch: What's `block` here? I believe it's a block of frames produced by the renderer, but It might not be clear for readers.

index.bs
padenot: It's regular English I think?

index.bs
padenot: Notes are informative in specs. I don't think this is necessary to implement the correct behaviour, but I like the way @karlt explained it, I felt like it could make understanding the issue easier.

index.bs
padenot: Better indeed, thanks.

index.bs
padenot: Fixed.

index.bs
padenot: Sure, I'll make it clearer.

index.bs
hoch: So it could be zero, but must be defined. Right?

index.bs
hoch: I was wondering if `Note:` is a magic keyword that puts this paragraph into a green box in the rendered document.

index.bs
```diff
@@ -11290,6 +11290,35 @@ defined below, skipping over those channels not present.
...
+- For a {{ConvolverNode}}, since the number of input channel determines the
+	<a href="#Convolution-channel-configurations">audio processing algorithm</a>
+	to use, it MUST immediately start outputing two channels.
+
+	Note: this only applies to the case where the {{ConvolverNode}}'s impulse
+	response is mono. Otherwise, the {{ConvolverNode}} always outputs a stereo
+	signal regardless of its input channel count.
+
+- For other {{AudioNode}}s that have a <a>tail-time</a>, the number of output
+	channels MUST increase after the <a>tail-time</a> has passed.
```
karlt: This is for DelayNode.

I guess it could also be for DynamicsCompressor, because pre-delay of 0.006 is more than one quantum, but the delay is small enough that I guess it doesn't matter which category is used for DynamicsCompressor.

This change should occur in time for the effects of the greater channel count and so not strictly "after".
```suggestion
	channels MUST increase when the input that was received with greater channel count begins to affect the output.
```

index.bs
```diff
@@ -11290,6 +11290,35 @@ defined below, skipping over those channels not present.
...
+Implication of tail-time on input and output channel count</h3>
+
+When an {{AudioNode}} has a non-null <a>tail-time</a>, and an output channel
+count that depends on the input channels count, any change in input channel
+MUST take the {{AudioNode}}'s <a>tail-time</a> into account
+
+When there is a decrease in input channel count, the change in output channel
+count MUST happen after the <a>tail-time</a> has passed.
+
+When there is an increase in input channel count, the behavious depends on the
```
karlt:
```suggestion
When there is an increase in input channel count, the behavior depends on the
```
for American spelling, as is elsewhere.

index.bs
```diff
@@ -11290,6 +11290,35 @@ defined below, skipping over those channels not present.
...
 
+<h3 id="channels-tail-time">
+Implication of tail-time on input and output channel count</h3>
+
+When an {{AudioNode}} has a non-null <a>tail-time</a>, and an output channel
+count that depends on the input channels count, any change in input channel
+MUST take the {{AudioNode}}'s <a>tail-time</a> into account
+
+When there is a decrease in input channel count, the change in output channel
+count MUST happen after the <a>tail-time</a> has passed.
```
karlt: I'm trying to think whether this is appropriate for DelayNode.  That would be fine if tail-time corresponds to the current `delayTime` in the node.  But perhaps "has any internal processing state which has not yet been emitted" corresponds more closely with `maxDelayTime`.  I assume we don't want to always wait `maxDelayTime`.

```suggestion
count MUST happen when the input that was received with greater channel count no longer affects the output.
```
which would still be fine for other nodes.

padenot: I think it's impossible to talk in terms of tail time for this indeed, and it's best to be more generic.

index.bs
```diff
@@ -11290,6 +11290,35 @@ defined below, skipping over those channels not present.
 		<tr><td>17 <td>SPEAKER_TOP_BACK_RIGHT <td> <td> <td> <td>
 </table>
 
+<h3 id="channels-tail-time">
+Implication of tail-time on input and output channel count</h3>
+
+When an {{AudioNode}} has a non-null <a>tail-time</a>, and an output channel
+count that depends on the input channels count, any change in input channel
+MUST take the {{AudioNode}}'s <a>tail-time</a> into account
```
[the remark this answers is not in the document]
padenot: Yes, better, thanks.

index.bs
padenot: Yes, it's a shorthand to make a div that is an informative note.

index.bs
```diff
@@ -11290,6 +11290,35 @@ defined below, skipping over those channels not present.
...
+
+When there is a decrease in input channel count, the change in output channel
+count MUST happen after the <a>tail-time</a> has passed.
+
+When there is an increase in input channel count, the behavious depends on the
+{{AudioNode}} type:
+
+- For a {{ConvolverNode}}, since the number of input channel determines the
+	<a href="#Convolution-channel-configurations">audio processing algorithm</a>
+	to use, it MUST immediately start outputing two channels.
```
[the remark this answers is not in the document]
padenot: I've move or less reversed the prose, only mentioning ConvolverNode in the informative note.

padenot:
I'm trying this github suggestions things out, but had to made adjustments locally, so I'll have to push a merge commit, hopefully everything is still readable.

index.bs
```diff
@@ -11299,25 +11299,26 @@ count that depends on the input channels count, the {{AudioNode}}'s
...
 	<a>tail-time</a>, the number of output channels MUST increase after the
 	<a>tail-time</a> has passed.
 
-- For the other {{AudioNode}}, since the number of input channel determines the
-	audio processing algorithm to use, the increase of the number of channels to
-	match the input channel count MUST happen immediately.
-	
+- For other {{AudioNode}}s that have a <a>tail-time</a>, the number of output
+	channels MUST increase when the input that was received with greater channel
+	count begins to affect the output.
```
karlt: I suspect something has got confused in this merge, but I can't see how to choose the other parent for the diff in the github UI.

The "input that was received with greater channel count begins to affect the output" was intended to replace "after the tail-time has passed" above, which is what was applied at
https://github.com/WebAudio/web-audio-api/pull/1787/commits/f45873a0a8d7140508573caa9812c8c1f81aef8f
but now it is replacing the immediate change.

padenot:
@hoch, do you want to have a final look at this before merging?

PR #1792, opened 2018-11-14
Fix the argument order in the automation diagram

### LINKED ISSUE #1790: Automation example chart has an error (reversed function arguments)
https://webaudio.github.io/web-audio-api/#example1-AudioParam

The first linearRamp label on the graph should read `linearRamp(1, 0.3)` not `linearRamp(0.3, 1)`.

padenot:
This fixes #1790.


padenot:
We don't have the source it seems, I've just used photoshop.

PR #1793, opened 2018-11-14
Add missing second 'new AudioContext()'

0joshuaolson1:
***
<a href="https://pr-preview.s3.amazonaws.com/0joshuaolson1/web-audio-api/pull/1793.html" title="Last updated on Nov 14, 2018, 11:12 PM GMT (22d40c6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1793/8b66e51...0joshuaolson1:22d40c6.html" title="Last updated on Nov 14, 2018, 11:12 PM GMT (22d40c6)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -9556,6 +9556,7 @@ registerProcessor('bypass-processor', BypassProcessor);
 // The main global scope
 let context = new AudioContext();
 context.audioWorklet.addModule('bypass-processor.js').then(() =&gt; {
+	let context = new AudioContext();
 	let bypassNode = new AudioWorkletNode(context, 'bypass-processor');
 });
 </pre>


padenot:
I don't understand why we would need a second context here?

0joshuaolson1:
I agree, but see all other code in the spec. As noted, this doesn't work in Chrome anyway.

padenot:
There is exactly one occurrence of having two `AudioContext`, and it's a bug indeed.

The fact that it does not work on Chrome is irrelevant here, you can file a ticket on Chromium [here](https://bugs.chromium.org/p/chromium/issues/entry?template=Audio%2FVideo+Issue).

@hoch, does the above make sense? Are you aware of the Chromium bug described above?

0joshuaolson1:
There are two; see my [new PR](https://github.com/WebAudio/web-audio-api/pull/1794) (removing them).

hoch:
My mistake. The line 3 in the example 15 must be removed.

PR #1808, opened 2019-01-11
Fix Media{Element,Stream}* ctors to spec the order in which the exceptions are to be thrown.

### LINKED ISSUE #1805: Exception ordering on error
This was found by @pehrsons when he was reviewing the patches for `MediaStreamTrackAudioSourceNode` in Gecko.

A instantiation of this `AudioNode` can fail for two reasons: either it's being created on an `OfflineAudioContext`, or the `MediaStreamTrack` passed in is not an audio track.

We should define the order in which we check both requirements.

I propose checking the context type first, and then the track type, to follow the order of the formal parameter.

Surely there are other occurrences of this pattern. Sometimes, we use an algorithm, which makes things very clear,  though, maybe we should do that more often.

padenot:
This fixes #1805.

***
<a href="https://pr-preview.s3.amazonaws.com/padenot/web-audio-api/pull/1808.html" title="Last updated on Jan 11, 2019, 1:32 PM UTC (0db12cb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1808/93d1edc...padenot:0db12cb.html" title="Last updated on Jan 11, 2019, 1:32 PM UTC (0db12cb)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -7707,13 +7707,15 @@ Constructors</h4>
 <dl dfn-type=constructor dfn-for="MediaElementAudioSourceNode">
 	: <dfn>MediaElementAudioSourceNode(context, options)</dfn>
 	::
-		Let <var>node</var> be a new {{MediaElementAudioSourceNode}}
-		object. <a href="#audionode-constructor-init">Initialize</a>
-		<var>node</var>, and return <var>node</var>.
+		1. If <var>context</var> is not an {{AudioContext}}, throw an
+			{{NotSupportedError}} exception and abort these steps.
+		2. Let <var>node</var> be a new {{MediaElementAudioSourceNode}} object.
+			<a href="#audionode-constructor-init">Initialize</a> <var>node</var>, and
+			return <var>node</var>.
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -7903,13 +7907,21 @@ Constructors</h4>
 <dl dfn-type=constructor dfn-for="MediaStreamAudioSourceNode">
 	: <dfn>MediaStreamAudioSourceNode(context, options)</dfn>
 	::
-		Let <var>node</var> be a new {{MediaStreamAudioSourceNode}}
-		object. <a href="#audionode-constructor-init">Initialize</a>
-		<var>node</var>, and return <var>node</var>.
-
+			1. If <var>context</var> is not an {{AudioContext}}, throw an
```
hoch: An editorial nit: why this line is indented deeper?

padenot: I don't know, I get an error from bikeshed if I don't do this, so I did it.

> +FATAL ERROR: Line 7710 isn't indented enough (needs 2 indents) to be valid Markdown:
> ...

PR #1810, opened 2019-01-15
Update the text to reflect the resolution on GC observability

### LINKED ISSUE #1471: AudioNode Lifetime section seems to attempt to make garbage collection observable
Attempting to describe observable behavior on garbage collection wouldn't make
sense because garbage collection is the removal of objects that are no longer
needed.  If their removal would cause changes in behaviour, then that implies
that the objects are still needed.

https://w3ctag.github.io/design-principles/#js-gc states the requirement in
reverse: "There must not be a way for author code to deduce when/if garbage
collection of JavaScript objects has run."

https://webaudio.github.io/web-audio-api/#lifetime-AudioNode
describes conditions that would keep an AudioNode alive and
the deletion of AudioNodes when none of these conditions are present.

The section appears to be describing garbage collection.  The first condition
is "A normal reference obeying normal garbage collection rules".  I'm not
clear exactly what this means.  Garbage collection attempts to detect objects
that are no longer needed.  Typically, implementations ensure that objects
that are still needed have "normal references" reachable from roots.

All other "types of references" listed seem to be explicit conditions where the
AudioNode is still needed.

One situation that is not explicitly described is when an AudioNode has
connections on its outputs.  In this situation non-unit output channel counts
can affect downstream input mixing, and the behavior of some kinds of
downstream nodes (e.g. AudioWorkletNode) depends on whether or not these
connections exist.  An implementation however is able to consider an otherwise
unneeded upstream AudioNode as unneeded if it maintains the (now unchangeable)
observable effects on downstream nodes when garbage collection the upstream node.

The section causing the most confusion is in

```
When an AudioNode has no references it will be deleted. Before it is deleted,
it will disconnect itself from any other AudioNodes which it is connected to.
In this way it releases all connection references (3) it has to other nodes.
```

The releasing of "connection references" to other nodes is kind of consistent
with the model that permits otherwise unneeded upstream nodes to be deleted,
but "disconnect itself from any other AudioNodes which it is connected to"
could be interpreted to mean a disconnection similar to an explicit
disconnect().  Such a disconnection can produce observable behavior changes on
downstream nodes.
https://github.com/WebAudio/web-audio-api/issues/1453#issuecomment-349340594 at least is interpreting this to mean that the effects of the connections are
removed together with the connection references when the AudioNode is deleted.

If this section is trying to define an object lifetime model that is different
from that of JS objects then that is problematic.  AudioNodes are JS objects,
and efficient and effective garbage collection algorithms cannot immediately
detect when an object becomes otherwise unneeded.  There is a indeterminate
delay before the object is deleted.

Requiring behavior changes when the object is otherwise unneeded leads to
unpredictable behavior, as the behavior may either continue as if the object
is otherwise needed for a long period of time or not at all.  Such variations
may occur either between implementations or within implementations.

padenot:
This is largely based off of @karlt's prose, with minor adjustments, and fixes a bunch of things. The commit that is the most important is the second one.

This fixes #1471.

***
<a href="https://pr-preview.s3.amazonaws.com/padenot/web-audio-api/pull/1810.html" title="Last updated on Apr 29, 2019, 12:23 PM UTC (ef592a5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1810/79d5743...padenot:ef592a5.html" title="Last updated on Apr 29, 2019, 12:23 PM UTC (ef592a5)">Diff</a>


index.bs
```diff
@@ -2581,6 +2581,37 @@ value explicitly. <em>block-size</em> is defined to be 128
...
 44.1 kHz.
 
+<h4 id="AudioNode-tail">AudioNode tail-time</h4>
+
+An {{AudioNode}} can have a <dfn>tail-time</dfn>. This means that even when the
+{{AudioNode}} is fed silence, non-silent output can be output.
+
+<h4 id="AudioNode-actively-processing">AudioNode lifetime</h4>
+
+{{AudioNode}} can be <dfn id="actively-processing">actively processing</dfn>
```
hoch: Perhaps this can be:

An AudioNode accessed during a render quantum is referred to as actively processing. When an AudioNode is not actively processing, it MUST generate a single channel of silence as an output. This state of "actively processing" varies on the type of AudioNodes:

index.bs
```diff
@@ -10057,7 +10007,8 @@ in the node.
 				while connected inputs and/or script references exist. Such
 				nodes SHOULD return <code>false</code> from
 				{{process()}} which allows the presence or absence of
-				connected inputs to determine whether active processing occurs.
+				connected inputs to determine whether the {{AudioWorkletNode}}s
+				[=actively processing=] occurs.
```
hoch: `actively processing occurs.` => `that is actively processing.`

index.bs
```diff
@@ -10057,7 +10007,8 @@ in the node.
 				while connected inputs and/or script references exist. Such
 				nodes SHOULD return <code>false</code> from
 				{{process()}} which allows the presence or absence of
-				connected inputs to determine whether active processing occurs.
+				connected inputs to determine whether the {{AudioWorkletNode}}s
```
hoch: `{{AudioWorkletNode}}s` => `{{AudioWorkletNode}}`

audionode.include
```diff
@@ -27,8 +27,8 @@
 			<td>"{{ChannelInterpretation/[CC-INTERP]}}"
 			<td>[CC-INTERP-NOTES?]
 		<tr>
-			<td><a>tail-time</a> reference
+			<td>tail-time
```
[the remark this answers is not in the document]
padenot: Ah yes sorry, this is a mistake.

index.bs
```diff
@@ -2581,6 +2581,37 @@ value explicitly. <em>block-size</em> is defined to be 128
 sample-frames which corresponds to roughly 3ms at a sample-rate of
 44.1 kHz.
 
+<h4 id="AudioNode-tail">AudioNode tail-time</h4>
+
+An {{AudioNode}} can have a <dfn>tail-time</dfn>. This means that even when the
```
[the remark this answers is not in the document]
padenot: I'm rephrasing this. I'm not a fan of saying that all nodes have X, but X has no consequences for most of the nodes.

index.bs
padenot: I find this to be too vague. When we're checking whether an `AudioScheduledSource` is supposed to write something in its output, we're somehow accessing it, aren't we?

index.bs
padenot: I rephrased this, what was above was a bit bogus.

index.bs
hoch: Then how about "being rendered" instead of "accessed"?

index.bs
```diff
@@ -2581,6 +2581,38 @@ value explicitly. <em>block-size</em> is defined to be 128
 sample-frames which corresponds to roughly 3ms at a sample-rate of
 44.1 kHz.
 
+<h4 id="AudioNode-tail">AudioNode tail-time</h4>
```
rtoy: If I read this correctly, I think this is the wrong place.  This new section should come just before 1.5.1 AudioNode Creation. Or maybe after.

Why?  Because in all other nodes the Interface section ends with the IDL.  If you do that here, I think the IDL ends up being in the lifetime section.

index.bs
```diff
@@ -2581,6 +2581,38 @@ value explicitly. <em>block-size</em> is defined to be 128
...
+
+An {{AudioNode}} can have a <dfn>tail-time</dfn> that is greater than zero. This
```
rtoy: You're defining tail-time here, but that's doesn't actually say what it is.  Need t move the actual definition of tail-time to here.  I guess we don't really have a definition of just tail-time.  It's tail-time reference.

In any case, we need a better definition.

index.bs
```diff
@@ -2581,6 +2581,38 @@ value explicitly. <em>block-size</em> is defined to be 128
...
+means that even when the {{AudioNode}} is fed silence, non-silent audio can be
+output.
+
+<h4 id="AudioNode-actively-processing">AudioNode lifetime</h4>
+
+{{AudioNode}} can be <dfn id="actively-processing">actively processing</dfn>
+during a <a>render quantum</a>.
+
+- An {{AudioScheduledSourceNode}} is [=actively processing=] when and only when it
```
rtoy: I think in English it's usually phrased "if and only if".

index.bs
```diff
@@ -2581,6 +2581,38 @@ value explicitly. <em>block-size</em> is defined to be 128
...
+	is playing for at least part of the current rendering quantum.
+- A {{DelayNode}} in a cycle is actively processing only when the absolute value
+	of any output sample for the current [=render quantum=] is greater than or equal
+	to \( 2^{-126} \).
```
rtoy: Why this value?  That's like -700 dB.

index.bs
```diff
@@ -2581,6 +2581,38 @@ value explicitly. <em>block-size</em> is defined to be 128
...
+- A {{ScriptProcessorNode}} is [=actively processing=] when its input or output is
+	connected.
+- An {{AudioWorkletNode}} is [=actively processing=] when its
+	{{AudioWorkletProcessor}}'s {{[[callable process]]}} is <code>true</code> and
+	either its [=active source=] flag is <code>true</code> or any {{AudioNode}}
+	connected to one of its inputs is [=actively processing=].
+- All other {{AudioNode}}s start [=actively processing=] when any
+	{{AudioNode}} connected to one of its inputs is [=actively processing=], and
```
rtoy: Need to say something about MediaElement/MediaStream sources which aren't AudioScheduledSourceNodes and don't have inputs.

index.bs
```diff
@@ -2581,6 +2581,38 @@ value explicitly. <em>block-size</em> is defined to be 128
...
+	stops [=actively processing=] when the input that was received from other
+	[=actively processing=] {{AudioNode}} no longer affects the output.
```
rtoy: "no longer affects the output" seems too vague.  Not sure if we can say anything better, but we should try.  I'm pretty sure Chrome and Firefox differ on when biquad tails are finished.

index.bs
```diff
@@ -2581,6 +2581,38 @@ value explicitly. <em>block-size</em> is defined to be 128
...
+
+Note: This takes into account {{AudioNode}}s that have a [=tail-time=].
```
rtoy: What does "This" refer to here?  Maybe you meant to indent this to match the last bullet item?

index.bs
```diff
@@ -2581,6 +2581,38 @@ value explicitly. <em>block-size</em> is defined to be 128
...
+
+{{AudioNode}} that are not [=actively processing=] output a single channel of
```
rtoy: Typo: "AudioNode" -> "AudioNodes"

index.bs
```diff
@@ -8346,7 +8297,7 @@ macros:
 	cc-mode-notes: Has <a>channelCountMode constraints</a>
 	cc-interp: speakers
 	tail-time: Maybe
-	tail-time-notes:  If the {{PannerNode/panningModel}} is set to "{{PanningModelType/HRTF}}", the node will produce non-silent output for silent input due to the inherent processing for head responses. Otherwise there are no tail-time references.
+	tail-time-notes:  If the {{PannerNode/panningModel}} is set to "{{PanningModelType/HRTF}}", the node will produce non-silent output for silent input due to the inherent processing for head responses. Otherwise there are no tail-time.
```
rtoy: Maybe "there are no tail-time" -> ", the tail-time is zero."  But I guess this really depends on how we define tail-time, which we haven't yet.

index.bs
```diff
@@ -11159,11 +11107,14 @@ Here's how the above diagram could be constructed in JavaScript:
...
 
-<h3 id="dynamic-lifetime-background">
+<h3 id="dynamic-lifetime-background" class="informative">
 Background</h3>
 
-<em>This section is non-normative. Please see <a href="#lifetime-AudioContext">AudioContext lifetime</a> and <a href="#lifetime-AudioNode">AudioNode lifetime</a> for normative
-requirements.</em>
+<em>This section is non-normative.</em>
+
+Note: The normative description of {{AudioContext}} and {{AudioNode}} lifetime characteristics is
```
rtoy: I think this Note should go before the paragraph that says the section is non-normative.

index.bs
```diff
@@ -11159,11 +11107,14 @@ Here's how the above diagram could be constructed in JavaScript:
...
+	available at <a href="#lifetime-AudioContext">AudioContext lifetime</a> and <a
+	href="#AudioNode-actively-processing">AudioNode lifetime</a>
```
rtoy: Typo:  Need a closing period to end the sentence.

index.bs
rtoy: As a definition, this sentence basically ends without defining what actively processing.  Maybe say that it's actively processing if the any of the following conditions hold?  This probably also means that the links to "actively processing" shouldn't be links.

index.bs
rtoy: Sorry, you actually define it in the next sentence. Perhaps combine the two sentences into one longer sentence.  Or maybe something like:

Each AudioNode has a tail-time.  The `<dfn>tail-time</dfn> is the time for silent input to be produced at the output of the node.  Some nodes like the BiquadFilterNode have mathematically infinite tail times; it is at the discretion of the UA to determine an appropriate finite value for the tail time.

index.bs
rtoy: Maybe we should not say "no longer effects output" but say that the tail-time of the node has passed.

Also, I think we want to rephrase "the input that was received from other actively processing AudioNode".  If the input is actively processing, the current node is also processing, per the previous bullet item.  I think we need to say that if all inputs are not actively processing, the current node continues to process until the tail time has passed at which point the node is no longer actively processing.

Is that too complicated?

index.bs
karlt: Yes, thanks.  How about these?

"An AudioScheduledSourceNode or MediaElementAudioSourceNode is actively
processing if and only if it or its `mediaElement` is playing for at least part of the current rendering quantum."

"A MediaStreamAudioSourceNode or MediaStreamTrackAudioSourceNode is actively
processing when the MediaStreamTrack is `live`, `unmuted`, and `enabled`."

index.bs
karlt: Referring to the tail-time is fine with me if that's clearer.  See my comment on tail-time for an attempt to more clearly describe this there.  We need to be a little careful to avoid the two definitions becoming too cyclic.

"the input that was received from other" is no longer required if using tail-time.  Perhaps:

"When there are no longer any actively processing AudioNodes connected to its inputs, an AudioNode continues actively processing for the length of its tail-time.  The AudioNode stops actively processing when its tail-time expires before it again has any actively processing AudioNodes connected to its inputs."

index.bs
karlt: BiquadFilterNode and others already describe their tail-times reasonably well.
One thing that is perhaps not perfectly clear from their descriptions is that the tail-time applies only after input changes from non-silent to silent.

The mention of "internal processing state" is dropped with the lifetime section.  It could be retained if that helps with the concept of "affects":

"Nodes have a non-zero tail-time if they have internal processing state such that input received at one time affects the output at later times.  Such nodes that have received non-silent input continue to produce non-silent output for the length of the tail-time after the input becomes silent."

index.bs
karlt: 2^126 is FLT_MIN, the smallest magnitude representable in IEEE 754 binary32 format without loss of precision.

Web audio uses 32-bit float sample format for "High dynamic range", presumably to provide support for values beyond what would be audible (which would be representable with merely 16-bit formats).  We don't know what gain might be applied to the signal, nor to what parameter the signal may be connected.

Implementations already often flush values smaller than FLT_MIN to zero in order to avoid potential performance problems with some hardware.  This is not currently specified, but FLT_MIN is chosen to define a cutoff that is likely already used for other reasons, rather than adding any new arbitrary cutoff.

I suspect FLT_TRUE_MIN would be harder for implementations to support with acceptable performance on all common hardware.  There is also significant loss of precision near this value, which can mean that the value is actually never reached in decay loops.

index.bs
padenot: I've picked @karlt sentence, that is more generic.

index.bs
padenot: Fixed.

index.bs
padenot: Fixed, doing something like @karlt's suggestions.

padenot:
I've addressed all comments, often with something based of off @karlt's suggestions.

padenot:
This is now green on travis, but I pushed this before pushing the workaround in https://github.com/WebAudio/web-audio-api/commit/354a1931ec7e9ba597a8afabebb8f78e68775cd2.

index.bs
```diff
@@ -2778,6 +2778,52 @@ enum ChannelInterpretation {
...
+	{{MediaElementAudioSourceNode/mediaElement}} is playing for at least part of
+	the current rendering quantum.
+- A {{MediaStreamAudioSourceNode}} or {{MediaStreamTrackAudioSourceNode}} when
+	the [[mediacapture-streams#mediastreamtrack|MediaStreamTrack]] used as input
+	has a <code>readyState</code> of <code>live</code>, a <code>muted</code>
+	attribute that is <code>false</code> and an <code>enabled</code> attribute
+	that is <code>true</code>.
+- A {{DelayNode}} in a cycle is [=actively processing=] only when the absolute value
+	of any output sample for the current [=render quantum=] is greater than or equal
+	to \( 2^{-126} \).
```
rtoy: This is a really, really tight constraint.  Let's say you have a biquad filter in the loop.  It's possible, due to round-off that the biquad never produces 0 output for 0 input.  

Not sure if this is achievable in practice.

index.bs
```diff
@@ -2778,6 +2778,52 @@ enum ChannelInterpretation {
...
+- A {{ScriptProcessorNode}} is [=actively processing=] when its input or output is
+	connected.
+- An {{AudioWorkletNode}} is [=actively processing=] when its
+	{{AudioWorkletProcessor}}'s {{[[callable process]]}} is <code>true</code> and
```
rtoy: You mean "returns true" instead of "is true"?

index.bs
```diff
@@ -2778,6 +2778,52 @@ enum ChannelInterpretation {
...
+	either its [=active source=] flag is <code>true</code> or any {{AudioNode}}
+	connected to one of its inputs is [=actively processing=].
+- All other {{AudioNode}}s start [=actively processing=] when any
+	{{AudioNode}} connected to one of its inputs is [=actively processing=], and
+	stops [=actively processing=] when the input that was received from other
+	[=actively processing=] {{AudioNode}} no longer affects the output.
```
rtoy: "no longer affects the output" is rather vague.  Say you have a constant source with a constant value of 1 that is actively processing. Say a 4x WaveShaper follows the source and the wave curve is the identity, more or less..  After a while, the constant source no longer affects the output because it too is a constant.  Does this mean the WaveShaper is no longer actively processing?  If so, the sentence below says it must output a single channel of silence.  That would be wrong since it should continue to output a constant value.

Perhaps we should say it stops actively processing if the inputs are not actively processing and the tail-time has passed?

index.bs
```diff
@@ -10071,7 +10036,8 @@ in the node.
 				while connected inputs and/or script references exist. Such
 				nodes SHOULD return <code>false</code> from
 				{{process()}} which allows the presence or absence of
-				connected inputs to determine whether active processing occurs.
+				connected inputs to determine whether the {{AudioWorkletNode}}s
```
rtoy: Maybe replace "AudioWorkletNodes actively processing occurs" -> "AudioWorkletNode is actively processing"?

index.bs
```diff
@@ -11173,11 +11135,14 @@ Here's how the above diagram could be constructed in JavaScript:
...
 Dynamic Lifetime</h2>
 
-<h3 id="dynamic-lifetime-background">
+<h3 id="dynamic-lifetime-background" class="informative">
 Background</h3>
 
-<em>This section is non-normative. Please see <a href="#lifetime-AudioContext">AudioContext lifetime</a> and <a href="#lifetime-AudioNode">AudioNode lifetime</a> for normative
-requirements.</em>
+Note: The normative description of {{AudioContext}} and {{AudioNode}} lifetime characteristics is
+	available at <a href="#lifetime-AudioContext">AudioContext lifetime</a> and <a
```
rtoy: Maybe replace "is available at" with "is described by the AudioContext lifetime and AudioNode lifetime"? (With links you already have.)

index.bs
karlt: I wonder whether you may have missed my reply to your previous comment about this.
https://github.com/WebAudio/web-audio-api/pull/1810#discussion_r253250386

The reason for choosing FLT_MIN here rather than FLT_TRUE_MIN is that the output does not need to reach zero.

Because full precision is available at FLT_MIN, the round-off occurring with values around FLT_MIN is similar to round-off at larger values.  If the decay (biquad or other) is such that input of any magnitude will attenuate, then values around FLT_MIN will attenuate.  After values around FLT_MIN attenuate, the cutoff is passed and the node outputs a single channel of silence.

Regardless of the value of the cutoff, perhaps there may be a chance of instability due to the sharp gradient at the cutoff if a biquad does not attenuate that sufficiently.  I don't know whether or not the output from such a biquad would decay to zero without the cutoff.  Even if the cutoff is introducing some new instability, it would be preferable that any remaining non-zero output from this corner-case would be determined by the precision of float32 rather than an arbitrary cutoff.

index.bs
padenot: Yes.

index.bs
padenot: The input affects the output, because if the input were to stop or change in value, then the output would change.

index.bs
padenot: Fixed.

index.bs
rtoy: This is acceptable to me. I still find the constraint very tight.  Does Firefox stop a BiquadFilter tail when it reaches FLT_MIN?  Chrome just estimates the tail time by computing when the impulse response is about 60 dB down (I think).  Even then, there are filters whose tail time is in the thousands of seconds.

index.bs
rtoy: What if I have a gain node with a gain of 0?  The input doesn't affect the output.  

I'm not sure, but maybe what I'm looking for is that a nodes stops actively processing when all inputs are no longer actively processing?

index.bs
```diff
@@ -2778,6 +2778,52 @@ enum ChannelInterpretation {
...
+non-silent output for the length of the tail-time after the input becomes
+silent.
+
+<h4 id="AudioNode-actively-processing">AudioNode lifetime</h4>
+
+{{AudioNode}} can be <dfn id="actively-processing">actively processing</dfn>
+during a <a>render quantum</a>, if any of the following conditions hold.
+
+- An {{AudioScheduledSourceNode}} is [=actively processing=] if and only if it
+	is playing for at least part of the current rendering quantum.
```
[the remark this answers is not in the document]
hoch: 2797:2798

"An AudioScheduledSourceNode is actively processing if and only if it is playing during a partial or entire render quantum."

For the definition of "playing", I guess we need to do it in the AudioScheduledSourceNode section.

index.bs
```diff
@@ -2778,6 +2778,52 @@ enum ChannelInterpretation {
...
+- A {{MediaElementAudioSourceNode}} is [=actively processing=] if and only if its
+	{{MediaElementAudioSourceNode/mediaElement}} is playing for at least part of
+	the current rendering quantum.
+- A {{MediaStreamAudioSourceNode}} or {{MediaStreamTrackAudioSourceNode}} when
```
hoch: "A MediaStreamAudioSourceNode or MediaStreamTrackAudioSourceNode when ..."

This reads a bit weird. Can you take a look once more?

index.bs
```diff
@@ -2778,6 +2778,52 @@ enum ChannelInterpretation {
...
+	the [[mediacapture-streams#mediastreamtrack|MediaStreamTrack]] used as input
+	has a <code>readyState</code> of <code>live</code>, a <code>muted</code>
+	attribute that is <code>false</code> and an <code>enabled</code> attribute
+	that is <code>true</code>.
```
hoch: 2802:2806

A MediaStreamAudioSourceNode and MediaStreamTrackAudioSourceNode are actively processing when an associated MediaStreamTrack object has the following attribute values:
  - `false` for `readyState.live`
  - `false` for `muted`
  - `true` for `enabled`

index.bs
```diff
@@ -2785,6 +2785,52 @@ enum ChannelInterpretation {
 </table>
 </div>
 
+<h4 id="AudioNode-tail">AudioNode tail-time</h4>
```
rtoy: Nit:  Maybe capitalize as "AudioNode Tail-Time".

index.bs
```diff
@@ -2785,6 +2785,52 @@ enum ChannelInterpretation {
...
+
+An {{AudioNode}} can have a <dfn>tail-time</dfn>. This means that even when the
+{{AudioNode}} is fed silence, the output can be non-silent.
+
+{{AudioNode}}s have a non-zero tail-time if they have internal processing state
+such that input in the past affects the future output. {{AudioNode}}s
+may continue to produce non-silent output for the calculated tail-time even
+after the input transitions from non-silent to silent.
+
+<h4 id="AudioNode-actively-processing">AudioNode lifetime</h4>
```
rtoy: Nit: Capitalize as "AudioNode Lifetime".

index.bs
```diff
@@ -4034,6 +3999,11 @@ the source MUST then output silence (0).
 {{AudioScheduledSourceNode}} cannot be instantiated directly, but
 is instead extended by the concrete interfaces for the source nodes.
 
+An {{AudioScheduledSourceNode}} is said to be <dfn>playing</dfn> when
+its associated {{BaseAudioContext}}'s {{BaseAudioContext|currentTime}} is
+greater or equal to the time it's set to start, and lesser than the time it's
```
rtoy: "it's set to start" -> "AudioScheduledSourceNode is set to start"
"lesser than the time" -> "less than the time"

index.bs
hoch: I wouldn't say RT60 is an arbitrary cutoff. It is considered the industry standard for reverberation decay. I think it has been accepted as a reasonable threshold for "perceptually inaudible" sound.

Can we make it up to UA? I think the numeric difference between RT60 and FLT_MIN is quite significant, but the audible difference would not really big. In that case, UA might be able to try different optimization techniques to improve the rendering performance.

padenot:
The issue here is that there can be arbitrary processing after the delay, including very high gain, unlike a normal reverb. Any difference can be amplified to a point where it's really different.

PR #1811, opened 2019-01-15
Define which track to use in MediaStreamAudioSourceNode more clearly.

### LINKED ISSUE #1772: Sorting tracks alphabetically is underspecified
See https://github.com/w3c/i18n-activity/issues/598

padenot:
This fixes #1772.

***
<a href="https://pr-preview.s3.amazonaws.com/padenot/web-audio-api/pull/1811.html" title="Last updated on Feb 7, 2019, 3:06 PM UTC (6d9c6fb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1811/615bbec...padenot:6d9c6fb.html" title="Last updated on Feb 7, 2019, 3:06 PM UTC (6d9c6fb)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -7906,12 +7906,12 @@ Constructors</h4>
 				<code>kind</code> attribute has the value <code>"audio"</code>,
 				throw an {{InvalidStateError}} and abort these steps. Else, let
 				this stream be <var>inputStream</var>.
-			3. Let <var>tracks</var> the list of all
-				[[mediacapture-streams#mediastreamtrack|MediaStreamTrack]] of
+			3. Let <var>tracks</var> be the list of all
+				[[mediacapture-streams#mediastreamtrack|MediaStreamTrack]]s of
 				<var>inputStream</var> that have a <code>kind</code> of
 				<code>"audio"</code>.
 			4. Sort the elements in <var>tracks</var> based on their <code>id</code>
-				attribute, ordering lexicographic ordering on sequences of code unit
+				attribute using lexicographic ordering on sequences of code unit
 				values.
 			5. Let <var>node</var> be a new {{MediaStreamAudioSourceNode}}
 				object. <a href="#audionode-constructor-init">Initialize</a>
@@ -7923,7 +7923,7 @@ Constructors</h4>
 			7. Return <var>node</var>.
 
 			Note: The behaviour for picking the track to output is arbitrary for
-			legacy reasons. {{MediaStreamTrackAudioSourceNode}} should be used
+			legacy reasons. {{MediaStreamTrackAudioSourceNode}} can be used
 			instead to be explicit about which track to use as input.
 
 		<pre class=argumentdef for="MediaStreamAudioSourceNode/MediaStreamAudioSourceNode()">
--- a/index.bs
+++ b/index.bs
@@ -7869,17 +7869,7 @@ Attributes</h4>
 The {{MediaStreamAudioSourceNode}} Interface</h3>
 
 This interface represents an audio source from a
-[[mediacapture-streams#mediastream|MediaStream]]. The track that will be used as the source
-of audio and will be output from this node is the first
-[[mediacapture-streams#mediastreamtrack|MediaStreamTrack]] whose <code>kind</code> attribute has
-the value <code>"audio"</code>, when alphabetically sorting the
-tracks of this [[mediacapture-streams#mediastream|MediaStream]] by their <code>id</code>
-attribute. Those interfaces are described in
-[[!mediacapture-streams]].
-
-Note: The behaviour for picking the track to output is weird for legacy
-reasons. {{MediaStreamTrackAudioSourceNode}} should be used
-instead.
+[[mediacapture-streams#mediastream|MediaStream]].
 
 <pre class=include>
 path: audionode-noinput.include
@@ -7888,9 +7878,9 @@ macros:
 	tail-time: No
 </pre>
 
-The number of channels of the output corresponds to the number of
-channels of the [[mediacapture-streams#mediastreamtrack|MediaStreamTrack]]. If there is no valid
-audio track, then the number of channels output will be one silent
+The number of channels of the output corresponds to the number of channels of
+the [[mediacapture-streams#mediastreamtrack|MediaStreamTrack]]. If there is no
+valid audio track, then the number of channels output will be one silent
 channel.
 
 <pre class="idl">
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -7914,10 +7904,25 @@ Constructors</h4>
...
-				throw an {{InvalidStateError}} and abort these steps.
-			3. Let <var>node</var> be a new {{MediaStreamAudioSourceNode}}
+				throw an {{InvalidStateError}} and abort these steps. Else, let
+				this stream be <var>inputStream</var>.
+			3. Let <var>tracks</var> the list of all
+				[[mediacapture-streams#mediastreamtrack|MediaStreamTrack]] of
+				<var>inputStream</var> that have a <code>kind</code> of
+				<code>"audio"</code>.
+			4. Sort the elements in <var>tracks</var> based on their <code>id</code>
+				attribute, ordering by Unicode code point
```
hoch: nit: add a period after "point".
question: Do we need a reference link on "Unicode code point"?

index.bs
```diff
@@ -7914,10 +7904,25 @@ Constructors</h4>
...
+			5. Let the first element of <var>tracks</var> be the
```
hoch: Do we need to describe the rest of tracks as well?

index.bs
```diff
@@ -7914,10 +7904,25 @@ Constructors</h4>
...
+				[[mediacapture-streams#mediastreamtrack|MediaStreamTrack]] used as the
+				input audio for this {{MediaStreamAudioSourceNode}}.
+			6. Let <var>node</var> be a new {{MediaStreamAudioSourceNode}}
```
hoch: Also how do we attach these "sorted tracks" to the node? Should it be explicit? or is it fine as it is now?

index.bs
```diff
@@ -7914,10 +7904,25 @@ Constructors</h4>
...
 				object. <a href="#audionode-constructor-init">Initialize</a>
 				<var>node</var>, and return <var>node</var>.
+
+			Note: The behaviour for picking the track to output is weird for legacy
```
hoch: I am not sure what this means. Can you elaborate?

Also perhaps: s/weird/inappropriate/

index.bs
padenot: No we leave them alone, they are of no importance for us.

index.bs
padenot: I was thinking maybe that we should stick the chosen track into a private slot, what do you think?

index.bs
padenot: I just copy/pasted this from above. I'm going to use the word "arbitrary", I think it reflects the thinking here and is more neutral.

index.bs
padenot: Unclear. Strictly speaking what we want here is a "lexicographic ordering on sequences of code unit values". Maybe it's best to just say this. It's what `sort` uses in JavaScript.

I found this here: https://tc39.github.io/ecma262/#sec-abstract-relational-comparison

index.bs
padenot: Specifically, the other tracks aren't mentioned, because regardless of what happens to them, it does not affect this node. I think it's still good to stick the track into a slot though.

In general, those `Media*AudioSourceNode` are a bit hand-wavy when it comes to the actual processing.

It's not that hard, but implementations vary widely (in terms of latency, notably), and, for example what happens when the input CORS-status changes during processing is a bit unclear.

hoch: +1 on putting it into a private slot. Clarifying behaviors of `Media*AudioSourceNode` can be a good V2 issue.

index.bs
hoch: Would "unspecified" be too strong for this? So what we want to point out is there might be an interop issue.

index.bs

index.bs
hoch: +1 on your suggestion. We don't need a link to the ECMA doc.

index.bs
padenot: We can't say it's unspecified if we're speccing it. We know what is going to happen, without ambiguity, it's just a bit strange.

index.bs
```diff
@@ -7888,9 +7878,9 @@ macros:
 	tail-time: No
 </pre>
 
-The number of channels of the output corresponds to the number of
-channels of the [[mediacapture-streams#mediastreamtrack|MediaStreamTrack]]. If there is no valid
-audio track, then the number of channels output will be one silent
+The number of channels of the output corresponds to the number of channels of
```
rtoy: As best as I can tell, nothing really changed here except for line wrapping.  Please don't do that.  It makes review harder.

You can leave this, but please verify there were no changes.

index.bs
```diff
@@ -7914,10 +7904,28 @@ Constructors</h4>
 				[[mediacapture-streams#mediastream|MediaStream]] that has at least one
 				[[mediacapture-streams#mediastreamtrack|MediaStramTrack]] whose
 				<code>kind</code> attribute has the value <code>"audio"</code>,
-				throw an {{InvalidStateError}} and abort these steps.
-			3. Let <var>node</var> be a new {{MediaStreamAudioSourceNode}}
+				throw an {{InvalidStateError}} and abort these steps. Else, let
+				this stream be <var>inputStream</var>.
+			3. Let <var>tracks</var> the list of all
```
rtoy: Typo: "Let tracks the" -> "Let tracks be the"

index.bs
```diff
@@ -7914,10 +7904,28 @@ Constructors</h4>
...
+				[[mediacapture-streams#mediastreamtrack|MediaStreamTrack]] of
```
rtoy: Typo: "MediaStreamTrack" -> "MediaStreamTracks"  (but probably "[[...|MediaStreamTrack]]s"

index.bs
```diff
@@ -7914,10 +7904,28 @@ Constructors</h4>
...
+				<var>inputStream</var> that have a <code>kind</code> of
+				<code>"audio"</code>.
+			4. Sort the elements in <var>tracks</var> based on their <code>id</code>
+				attribute, ordering lexicographic ordering on sequences of code unit
```
rtoy: Probable typo:  "ordering" is used twice here.  Need to rephrase.

index.bs
```diff
@@ -7914,10 +7904,28 @@ Constructors</h4>
...
-				<var>node</var>, and return <var>node</var>.
+				<var>node</var>.
+			6. Set an internal slot [[input track]] on this
+				{{MediaStreamAudioSourceNode}} to be the first element of
+				<var>tracks</var>.  This is the track used as the input audio for this
+				{{MediaStreamAudioSourceNode}}.
+			7. Return <var>node</var>.
+
+			Note: The behaviour for picking the track to output is arbitrary for
+			legacy reasons. {{MediaStreamTrackAudioSourceNode}} should be used
```
rtoy: I think "should" has specific spec meaning here. Is this what we want?

index.bs
```diff
@@ -7914,10 +7904,28 @@ Constructors</h4>
...
-				throw an {{InvalidStateError}} and abort these steps.
-			3. Let <var>node</var> be a new {{MediaStreamAudioSourceNode}}
+				throw an {{InvalidStateError}} and abort these steps. Else, let
+				this stream be <var>inputStream</var>.
+			3. Let <var>tracks</var> be the list of all
+				[[mediacapture-streams#mediastreamtrack|MediaStreamTrack]]s of
+				<var>inputStream</var> that have a <code>kind</code> of
+				<code>"audio"</code>.
+			4. Sort the elements in <var>tracks</var> based on their <code>id</code>
+				attribute using lexicographic ordering on sequences of code unit
```
rtoy: code unit or code point?  They are different.  But  I see that Javascript sorts by code unit (which seems really odd), so we're good here.

padenot: Yes I'm purposefully using this term here so that it's aligned with what JavaScript's `sort` is doing.

PR #1816, opened 2019-01-17
Fix #1815: restore text audioparam insert

### LINKED ISSUE #1815: Multiple events at the same time
In https://webaudio.github.io/web-audio-api/#AudioParam, paragraph 9 (after the list of events), we describe how events are inserted.

Compare that with https://www.w3.org/TR/2013/WD-webaudio-20131010/#methodsandparams-AudioParam.  We've lost this text:

- If one of these events is added at a time where there is already an event of the exact same type, then the new event will replace the old one.
- If one of these events is added at a time where there is already one or more events of a different type, then it will be placed in the list after them, but before events whose times are after the event.

I think this is how browsers currently handle this.  We should put this text back.

rtoy:
Put back the bullet item that was somehow removed sometime after the 2015 published draft.

***
<a href="https://pr-preview.s3.amazonaws.com/rtoy/web-audio-api/pull/1816.html" title="Last updated on Jan 17, 2019, 5:33 PM UTC (750cf6c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1816/f9b883c...rtoy:750cf6c.html" title="Last updated on Jan 17, 2019, 5:33 PM UTC (750cf6c)">Diff</a>


rtoy:
Abandon and try again in a different PR.

PR #1823, opened 2019-01-29
Specify autoplay behavior

### LINKED ISSUE #1802: Clarify interaction between user-invoked suspend and autoplay policy
We still have [the PR under the review](https://github.com/WebAudio/web-audio-api/pull/1779), but I think this deserve another look by WG.

In Chrome, the autoplay policy automatically suspends the context when UA is _not allowed to start_ it. This "automatic" suspension by autoplay policy will be lifted off when 1) `context.resume()` call or 2) `source.start()` call upon user activation.

However, the problem here is that there is no way to distinguish between _automatic suspension_ and _user-invoked suspension_. Therefore, even if user explicitly suspended the context, the current design will unblock the context and resume the rendering.

Consider the following example:

```js
const context = new AudioContext();
const source = new OscillatorNode(context);
source.connect(context.destination);

context.suspend(); // Explicit suspension

someButton.addEventListener('click', () => {
  console.log(context.state); // A
  source.start();
  console.log(context.state); // B
});
```

Apparently violating user's explicit request for suspension is not acceptable. So A and B should both be `suspended`. But this still brings a couple of questions:

1. Then what happens to `source.start()`? Do we still have to stash it somewhere and play it when the suspension is lifted off? Or should it be disregarded completely?
2. How do we differentiate a suspension from autoplay policy and user code? If we have to introduce another flag for this, now we have to deal with 3 internal flags; `was allowed to start`, `is allowed to start`, and `user suspension`. I think this needs a careful review to figure out all the corner cases.

cc: @mounirlamouri

hoch:
Fixes #1802 based on [the comment](https://github.com/WebAudio/web-audio-api/issues/1802#issuecomment-454154319) in the discussion.

cc @mounirlamouri

***
<a href="https://pr-preview.s3.amazonaws.com/hoch/web-audio-api/pull/1823.html" title="Last updated on Feb 26, 2019, 9:32 PM UTC (d614956)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1823/03265e4...hoch:d614956.html" title="Last updated on Feb 26, 2019, 9:32 PM UTC (d614956)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1345,7 +1345,7 @@ activation</a> (as described in [[HTML]]).
 {{AudioContext}} has an internal slot:
 
 <dl dfn-type=attribute dfn-for="AudioContext">
-	: <dfn>[[is suspended by user]]</dfn>
+	: <dfn>[[suspended by user]]</dfn>
 	::
 		A boolean flag representing whether the context is suspended by user code.
 		The initial value is <code>false</code>.
@@ -1659,11 +1659,11 @@ Methods</h4>
 				promise with {{InvalidStateError}}, abort these steps,
 				returning <var>promise</var>.
 
-			3. Set {{[[is suspended by user]]}} to <code>false</code>.
+			3. Set {{[[suspended by user]]}} to <code>false</code>.
 
-			4. If the context is not <a>allowed to start</a>, append<var>promise</var>
-				to <a>pendingResumePromises</a> and abort these steps, returning
-				<var>promise</var>.
+			4. If the context is not <a>allowed to start</a>, append
+				<var>promise</var> to <a>pendingResumePromises</a> and abort these
+				steps, returning <var>promise</var>.
 
 			5. Set the <em>control thread state</em> flag on the
 				{{AudioContext}} to <code>running</code>.
@@ -1739,7 +1739,7 @@ Methods</h4>
 				with {{InvalidStateError}}, abort these steps,
 				returning <em>promise</em>.
 
-			3. Set {{[[is suspended by user]]}} to <code>true</code>.
+			3. Set {{[[suspended by user]]}} to <code>true</code>.
 
 			4. Set the <em>control thread state</em> flag on the {{AudioContext}} to <code>suspended</code>.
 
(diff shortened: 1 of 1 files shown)


rtoy:
Just a general comment.  As this currently stands, "is allowed to play" never changes so `start()` never does anything.  We need to say something about this so "is allowed to play" can change somehow. Like when start is called, it becomes true if the UA allows playing.  Or something.

mounirlamouri:
As discussed in the bug/other PR, my concern is that this does not match neither Safari nor Chrome's implementation. Even if it's possible that @padenot will implement it this way in Firefox and that Chrome will update its implementation, we do not know the compat impact it will have. We also should make sure Safari is willing to change its implementation.

mounirlamouri:
I agree that this is something that we should aim for but I think we should make sure we keep these things in mind:
 - We need to guarantee web compat by specifying something that matches implementation (and have implementation follow what's spec'd). Safari had this launched years ago so it would be good to make sure they would be willing to change.
 - We need to guarantee backward compat. I think in this case it shouldn't be a real problem but maybe some metrics that tell us if there would be a significant amount of websites that would end up no longer playing on a call to `start()`.

hoch:
I generally agree on those points, but a response from Safari team would be really helpful. We definitely want to address this as soon as possible.

hoch:
@rtoy I addressed your feedback by removing `is allowed to start` flag. Now it's just an internal state that is changed by UA so the algorithm does not touch it.

padenot:
Gecko has implemented what is being described here: respecting the authors's intent when the context has been explicitly suspended, and not resuming in this case.

hoch:
@padenot PTAL at this PR and see if this matches Gecko's implementation.

padenot:
@alastor0325, would you mind having a look at this one?

Gecko bug is [1519430](https://bugzilla.mozilla.org/show_bug.cgi?id=1519430), for reference.

alastor0325:
LGTM, but there is one thing looks not quite correct on gecko.

`An AudioContext is said to be allowed to start if the user agent allows the current context to produce audible output.` 

Even if the AudioContext doesn't output any audible sound, we would still not allow it to start in the beginning, so it seems to me the previous description, `allowed to transition from "suspended" to "running"` , is more suitable.

hoch:
Thanks for the feedback! To be precise, what we want is:
`An AudioContext is said to be allowed to start if the user agent allows the context state to transition from "suspended" to "running".`

PR #1829, opened 2019-02-19
Make MediaStreamAudioSourceNode a bit more precise

### LINKED ISSUE #1824: MediaStreamAudioSourceNode can't have no valid audio track
**Describe the issue**
I noticed that there is a little conflict in the description of the MediaStreamAudioSourceNode.

The general description says: "If there is no valid audio track, then the number of channels output will be one silent channel."

But the algorithm of the constructor says "If the mediaStream member of options does not reference a MediaStream that has at least one MediaStramTrack whose kind attribute has the value "audio", throw an InvalidStateError and abort these steps."

I think the fact that the constructor throws an error when there is no valid track in the given MediaStream eliminates the necessity to describe what happens in there is no valid track.

Is it possible to modify the MediaStream (add/remove tracks) while it is the input of an MediaStreamAudioSourceNode? I think in that case the first sentence makes sense again as a valid track could be present at construction time but might be removed later.

The [MediaStream Recording spec](https://w3c.github.io/mediacapture-record/MediaRecorder.html) does for example explicitly forbid to modify the MediaStream while it is used for recording. Maybe that needs to be added here as well if that's the case. (5.3 of that algorithm https://w3c.github.io/mediacapture-record/MediaRecorder.html#dom-mediarecorder-start)

There is also a little typo: "MediaStramTrack" should probably be "MediaStreamTrack".

**Where Is It**
https://webaudio.github.io/web-audio-api/#mediastreamaudiosourcenode
https://webaudio.github.io/web-audio-api/#dom-mediastreamaudiosourcenode-mediastreamaudiosourcenode

padenot:
This fixes #1824.

***
<a href="https://pr-preview.s3.amazonaws.com/padenot/web-audio-api/pull/1829.html" title="Last updated on Feb 25, 2019, 1:35 PM UTC (b0c5f0f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1829/1a6fa79...padenot:b0c5f0f.html" title="Last updated on Feb 25, 2019, 1:35 PM UTC (b0c5f0f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -7886,8 +7886,8 @@ macros:
 
 The number of channels of the output corresponds to the number of channels of
 the [[mediacapture-streams#mediastreamtrack|MediaStreamTrack]]. When the
-[[mediacapture-streams#mediastreamtrack|MediaStreamTrack]] ends, one channel of
-silence is output.
+[[mediacapture-streams#mediastreamtrack|MediaStreamTrack]] ends, this
+{{AudioNode}} outputs one channel of silence.
 
 <pre class="idl">
 [Exposed=Window,
--- a/index.bs
+++ b/index.bs
@@ -7908,7 +7908,7 @@ Constructors</h4>
 			2. If the {{MediaStreamAudioSourceOptions/mediaStream}} member of
 				<code>options</code> does not reference a
 				[[mediacapture-streams#mediastream|MediaStream]] that has at least one
-				[[mediacapture-streams#mediastreamtrack|MediaStramTrack]] whose
+				[[mediacapture-streams#mediastreamtrack|MediaStreamTrack]] whose
 				<code>kind</code> attribute has the value <code>"audio"</code>,
 				throw an {{InvalidStateError}} and abort these steps. Else, let
 				this stream be <var>inputStream</var>.
--- a/index.bs
+++ b/index.bs
@@ -7928,13 +7928,13 @@ Constructors</h4>
 				{{MediaStreamAudioSourceNode}}.
 			7. Return <var>node</var>.
 
-			Changes to the [[mediacapture-streams#mediastream|MediaStream]] that has
-			been passed to the constructor after construction do not affect the
-			underlying output of this {{AudioNode}}.
+			After construction, any change to the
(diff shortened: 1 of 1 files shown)


padenot:
@chrisguttandin, if you want to have a look ?

chrisguttandin:
Hi @padenot, thanks a lot. That all looks good to me.

I only spotted one little thing. I'm no native speaker, so this might be wrong but "When the MediaStreamTrack ends, one channel of silence is output." sounds a bit strange to me.

What about "When the MediaStreamTrack ends, the output is one channel of silence."?

padenot:
Thanks, fixed.

index.bs
```diff
@@ -7928,6 +7928,16 @@ Constructors</h4>
 				{{MediaStreamAudioSourceNode}}.
 			7. Return <var>node</var>.
 
+			Changes to the [[mediacapture-streams#mediastream|MediaStream]] that has
```
hoch: has > have

index.bs
```diff
@@ -7928,6 +7928,16 @@ Constructors</h4>
...
+			been passed to the constructor after construction do not affect the
```
hoch: How do you pass things to the constructor after construction? Perhaps this needs a bit of rephrasing?

index.bs
```diff
@@ -7928,6 +7928,16 @@ Constructors</h4>
...
+			underlying output of this {{AudioNode}}.
+
+			Note: This means that when removing the track chosen by the constructor
+			of the {{MediaStreamAudioSourceNode}} from the
+			[[mediacapture-streams#mediastream|MediaStream]] passed in this
```
hoch: in > into, perhaps?

index.bs
padenot: I'm missing a comma or something like this.

padenot: Ha no my sentence is just bad: I meant Changes to [the MediaStream that has been passed to the constructor] after construction [...]. I'm shuffling things around so it not ambiguous.

Issue #1812, opened 2019-01-15
Verifying the AudioWorkletProcessor superclass on prototype chain


hoch:
Branched from: https://github.com/w3c/css-houdini-drafts/issues/850

Currently AudioWorkletProcessor spec requires extending the processor class but the actual registration algorithm does not actually check the prototype chain to verify this.

To specify this correctly we need to check the prototype chain and see if any ancestor is actually the class we expect a.k.a. `IsSubclassOf(class, expected_ancestor_class)`. AFAICT, there is no such operation in ECMA262 to achieve this. 

Our options are (ordered by my personal preference):
- Implement this in our own spec e.g., by walking up the chain and using the [SameValue].
- Do not use super-class in spec algorithm.
- Ask ECMA262 to add such an operation.

The resolution also needs to be coordinate with the other worklet variants. I believe this doesn't block our progress, and is more of an editorial issue.

hoch:
Resolution: We will closely monitor the progress from Houdini and Animation Worklet group.

karlt:
I don't know of any reason why we should need to check the prototype chain.

AFAIK the only purpose for the AudioWorkletProcessor class is to "automatically" provide a MessagePort that is entangled with the `port` on the AudioWorkletNode.

I assume nodes that don't need to use the port don't need to extend AudioWorkletProcessor.

In most places where `AudioWorkletProcessor` is used in the spec, it is merely to identify the processing-thread object associated with the node.  Perhaps this could be expressed differently so as not to imply any need for extend the AudioWorkletProcessor class.

I guess entangled MessagePorts could even instead be passed via AudioWorkletNodeOptions.processorOptions, and so the AudioWorkletProcessor class is not actually _needed_ at all.  It is syntactic sugar.

karlt:
https://heycam.github.io/webidl/#es-interface when used with https://heycam.github.io/webidl/#construct-a-callback-function provides a way to detect whether the constructed processor object implements AudioWorkletProcessor.

https://github.com/WebAudio/web-audio-api/pull/1843#issuecomment-478590304 points out that step 6 of the second algorithm at https://webaudio.github.io/web-audio-api/#instantiation-of-AudioWorkletNode-and-AudioWorkletProcessor expects the constructed object to be an AudioWorkletProcessor.  I assume it would be possible to relax that requirement, but that does not seem the intention of the text.

majido:
[...]

PR #1843, opened 2019-03-26
Add `AudioWorkletProcessorConstructor` in AudioWorkletGlobalScope

### LINKED ISSUE #1839: parameters and return value of processorCtor parameter to AudioWorkletGlobalScope.registerProcessor()
AudioWorkletGlobalScope.registerProcessor() is declared in webidl with the processorCtor parameter of type VoidFunction.
https://webaudio.github.io/web-audio-api/#audioworkletglobalscope
"The VoidFunction callback function type is used for representing function values that take no arguments and do not return any value."
https://heycam.github.io/webidl/#VoidFunction

However, "the construction of an AudioWorkletProcessor" requires calling Construct() with a list of options, and so the processorCtor would be called with an argument.

Further, the return value is from the constructor is used in the Construct algorithm, and so I'm not clear how a missing return value would be enforced or ignored.
https://tc39.github.io/ecma262/#sec-construct
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/new

hoch:
Fixes #1839.

cc @karlt

***
<a href="https://pr-preview.s3.amazonaws.com/hoch/web-audio-api/pull/1843.html" title="Last updated on Mar 26, 2019, 4:31 PM UTC (4c3452c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1843/83cd1ad...hoch:4c3452c.html" title="Last updated on Mar 26, 2019, 4:31 PM UTC (4c3452c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -9670,9 +9670,11 @@ for that context. This prevents data races from occurring in global
 scope code running within concurrent threads.
 
 <pre class="idl">
+callback AudioWorkletProcessorConstructor = any (any options);
+
 [Global=(Worklet, AudioWorklet), Exposed=AudioWorklet]
 interface AudioWorkletGlobalScope : WorkletGlobalScope {
-	void registerProcessor (DOMString name, VoidFunction processorCtor);
+	void registerProcessor (DOMString name, AudioWorkletProcessorConstructor processorCtor);
 	readonly attribute unsigned long long currentFrame;
 	readonly attribute double currentTime;
 	readonly attribute float sampleRate;
--- a/index.bs
+++ b/index.bs
@@ -9670,7 +9670,7 @@ for that context. This prevents data races from occurring in global
 scope code running within concurrent threads.
 
 <pre class="idl">
-callback AudioWorkletProcessorConstructor = any (any options);
+callback AudioWorkletProcessorConstructor = AudioWorkletProcessor (object options);
 
 [Global=(Worklet, AudioWorklet), Exposed=AudioWorklet]
 interface AudioWorkletGlobalScope : WorkletGlobalScope {


karlt:
Makes sense to me, thank you.

annevk:
This doesn't work if AudioWorkletProcessorConstructor needs to be invoked as a constructor would be invoked. We should create a new IDL type for this.

bzbarsky:
Why is the return type of this callback any? What is the actual processing model?

In https://webaudio.github.io/web-audio-api/#instantiation-of-AudioWorkletNode-and-AudioWorkletProcessor second algorithm block (which is not directly linkable, sadly) step 6 certainly seems to make assumptions about the return value not just being any but being an AudioWorkletProcessor instance, right? As the spec is written, even with this PR, that's not at all guaranteed.

Also, the manual Construct call returns a completion, not just a value. Importantly, it does NOT throw an exception in the sense that IDL understands exceptions. So the way the spec is written right now, exception handling is undefined. It should probably use https://heycam.github.io/webidl/#construct-a-callback-function instead of ES Construct to turn abrupt completions into exceptions as needed. See https://html.spec.whatwg.org/multipage/custom-elements.html#concept-upgrade-an-element step 8 for an example.

karlt:
https://heycam.github.io/webidl/#construct-a-callback-function defines a way to invoke AudioWorkletProcessorConstructor as a constructor.
https://github.com/heycam/webidl/issues/701 proposes a new, more explicit way to declare constructors.

https://github.com/WebAudio/web-audio-api/issues/1812 discusses what the constructor should return.  Specifying the return value of AudioWorkletProcessorConstructor as AudioWorkletProcessor is a possible solution to that issue.

Step 7 of the first algorithm at https://webaudio.github.io/web-audio-api/#instantiation-of-AudioWorkletNode-and-AudioWorkletProcessor invokes StructuredSerialize() on `options`.  `options` is an IDL dictionary (from a trailing optional parameter to the AudioWorkletNode constructor), but StructuredSerialize() expects a JS value.  We should have an explicit conversion of `options` to a JS/ECMAScript object there.

In the second algorithm, we therefore expect `options` from StructuredDeserialize() to be a JS object.  This would be converted to the IDL type declared for the AudioWorkletProcessorConstructor parameter in order to use https://heycam.github.io/webidl/#construct-a-callback-function .
To keep that conversion as simple as possible, the parameter would be of type `object`.

i.e., pending resolution of https://github.com/heycam/webidl/issues/701:
callback AudioWorkletProcessorConstructor = AudioWorkletProcessor (object options);

bzbarsky:
>  callback AudioWorkletProcessorConstructor = AudioWorkletProcessor (object options);

This seems like the best option for now, yes.

hoch:
@padenot PTAL. My last revision is based on https://github.com/WebAudio/web-audio-api/pull/1843#issuecomment-479049620.

Issue #1453, opened 2017-11-20
[audioworklet] Permit process() calls to be skipped on inactive AudioWorkletProcessors with silent input


karlt:
An optimization that was found to be important in use cases where one-shot
AudioNodes were frequently created was to skip processing when the input
was known to be silent.  A similar optimization would be valuable for
AudioWorkletProcessors.

In active documents with many objects, it is not practical to perform full
garbage collection frequently, and so there is often a long delay before
unreferenced inactive nodes are collected.  Silent input, however, can be
determined promptly.

Prompt detection of unnecessary processing is very important in graphs with
many one-shot nodes such as at
http://webaudiodemos.appspot.com/MIDIDrums/index.html

Skipping process() is also important in the long term for releasing the
resources associated with the AudioWorkletProcessors.  While process()
calls are still required, the node and processor are observable and so
cannot be garbage collected.

If the process() call can be skipped when input is silent, then clients need
not explicitly disconnect all AudioWorklet inputs before removing their last
references to the nodes.

karlt:
I don't have a strong preference re whether implementations MUST or MAY skip
process() calls when input is silent (and the active source flag is false).
It is easy to test whether inputs are all silent.  OTOH AudioWorkletProcessors
need to be able to process at least one call with silent input because the
active source flag is initially true (with good reason), and so there is
probably little harm in a few unnecessary calls.  In general though
a determistic solution is preferable where practical, and it seems to be
practical here.

hoch:
I agree with the idea, but this might fall into the implementation detail side. The interesting part is users can observe the implementation difference if some browsers skip calling `process()` for the optimization.

joeberkovitz:
I can't see how this optimization is feasible in the general case, given that any processor could add its own generated signal to the input as part of processing, even when the input is silent.

So this feels like an optional optimization that a developer would have to request. As such I would put this into V.next since we would need to design a mechanism for such optional behaviors.

rtoy:
I agree with @joeberkovitz.  Silent inputs does not imply process can be skipped.  In fact, I think I might go farther and say we should never skip calls to process.  However, I am ok with adding hints to let process know if the inputs are silent.  This would allow process to do something intelligent, without having to check for itself if the inputs are silent.

rtoy:
I agree with @joeberkovitz.  Silent inputs does not imply process can be skipped.  In fact, I think I might go farther and say we should never skip calls to process.

[...]

PR #1875, opened 2019-05-14
Apply new GC changes to AudioWorkletNode's processing condition

hoch:
Fixes: #1453.

Revise the condition of Audio Worklet node that is "actively processing".

cc: @karlt

***
<a href="https://pr-preview.s3.amazonaws.com/hoch/web-audio-api/pull/1875.html" title="Last updated on Jun 3, 2019, 8:34 PM UTC (83bef39)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1875/7e24f10...hoch:83bef39.html" title="Last updated on Jun 3, 2019, 8:34 PM UTC (83bef39)">Diff</a>


karlt:
This change captures the intent, thanks @hoch, but note that callable process is already required to be true for the AudioWorkletNode to be actively processing.  This can therefore be simplified further as the first of the two conditions becomes redundant.

hoch:
So you mean "The AudioWorkletNode is actively processing" alone is good enough here?

karlt:
Yes, and so it could be simplified to "The process() method is called synchronously by the audio rendering thread at every render quantum, if the AudioWorkletNode is actively processing."

hoch:
That's better. Done!

Issue #1857, opened 2019-05-09
Fix typos and bad phrasing


rtoy:

rtoy:
I think with the last commit from #1890, we are finished with all of typos/rephrasing now.

File new issues if we find more.

PR #1913, opened 2019-05-17
Address #1857: Fix typos/phrasing through Requirements

rtoy:
* Fix typos and phrasing
 * Define a few terms and add some link to them.

This concludes the full proofreading of the spec.

***
<a href="https://pr-preview.s3.amazonaws.com/rtoy/web-audio-api/pull/1913.html" title="Last updated on May 30, 2019, 9:40 PM UTC (80c7fe0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1913/a014e46...rtoy:80c7fe0.html" title="Last updated on May 30, 2019, 9:40 PM UTC (80c7fe0)">Diff</a>


index.bs
```diff
@@ -11576,13 +11576,13 @@ Background</h3>
 
 A common feature requirement for modern 3D games is the ability to
 dynamically spatialize and move multiple audio sources in 3D space.
-Game audio engines such as OpenAL, FMOD, Creative's EAX, Microsoft's
-XACT Audio, etc. have this ability.
+Game audio engines such as <a href="https://openal.org">OpenAL</a>, <a href="http://www.fmod.org">FMOD</a>, <a href="https://www.creative.com/oem/technology/eax.asp">Creative's EAX</a>, <a href="https://www.creative.com/oem/technology/eax.asp">Microsoft's
```
rtoy: Not really sure if we should provide links here.  And I'm not sure I got them all right.  And XACT Audio has reached end-of-life, so maybe it should be removed completely.

index.bs
```diff
@@ -11705,7 +11705,10 @@ provides basic, but reasonable results. It is used for the for the
 {{PannerNode}} when the {{PannerNode/panningModel}} attribute is set to
 "{{PanningModelType/equalpower}}", in which case the <em>elevation</em>
 value is ignored. This algorithm MUST be implemented using
-<a>a-rate</a> parameters.
+at the appropriate rate as specified by the
```
rtoy: This changes the spec, but I think we need to say this because the params can be either a-rate or k-rate.

index.bs
hoch: Well, I am not sure why we have to provide those specific names in here. I am fine with removing all of them.

index.bs
```diff
@@ -11596,7 +11596,7 @@ example, the sound could be omnidirectional, in which case it would
 be heard anywhere regardless of its orientation, or it can be more
 directional and heard only if it is facing the listener.
 {{AudioListener}} objects (representing a person's
-ears) have <em>forward</em> and <em>up</em> vectors
+ears) have [=front|forward=] and [=up=] vectors
```
hoch: Perhaps `front` should be `forward` as well?

index.bs
hoch: I agree, but let's confirm with @padenot.

index.bs
```diff
@@ -12171,7 +12173,7 @@ to Consider</a>
 	been demonstrated
 	<a href="https://pdfs.semanticscholar.org/cfd2/6a17234696593919df3f880a235d6ac5871d.pdf">
 	by Nakibly et. al. for HTML</a>. The <a href="https://w3c.github.io/hr-time/#sec-privacy-security">
-	Security & Privacy appending of High Resolution Time</a> should be consulted for further
+	Privacy & Security appendix of High Resolution Time</a> should be consulted for further
```
hoch: The section title is "Security and Privacy Consideration", so we might want to flip the order.

index.bs
rtoy: WFM.  I'll just remove this.

index.bs
rtoy: Yeah, I'll do that. I was just preserving the existing text.  But I'll have to check other places that refer to "front" and change them to forward.  Maybe I'll do that in a different PR?

index.bs
rtoy: Aargh! I added a typo here that I need to fix:  "using at the appropriate..."

index.bs
rtoy: But the HR-time spec calls it "Privacy and Security" if you follow the link.  That's why I changed the order.

index.bs
rtoy: Added @padenot as reviewer to look at this.

index.bs
rtoy: Per offline discussion, we'll remove everything but leave OpenAL since that was used as the basis for much of the `PannerNode`.

index.bs
rtoy: I think it's better to leave this as is in the PR.  If we really want to, a follow-up PR can change all references from "front" to "forward".

index.bs

index.bs
hoch: SGTM

rtoy:
I'm going to land this.  If there are errors, @padenot, can you file a new issue, please?

PR #1930, opened 2019-05-24
Fix #1905: Remove paragraph on observable rendering callbacks

### LINKED ISSUE #1905: Rendering callbacks are observable
The note about rendering in [Section 2.4](https://webaudio.github.io/web-audio-api/#rendering-loop) says

> After some time, the underlying audio system will call the callback again, and the algorithm will be executed again. This is an implementation detail that should not be observable, apart from the latency implications.

Actually, I think this implementation detail was always observable.  If the main thread kept asking for `currentTime`, it can see that `currentTime` gets updated several times very close together, and then there's a pause of a few milliseconds before this burst of updates happen again

Also, [getOutputTimestamp](https://webaudio.github.io/web-audio-api/#dom-audiocontext-getoutputtimestamp) can show this too.

Delete the sentence about not being observable?

rtoy:
Remove the part of the note about how buffering happens and howmultiple renders for each callback is not observable because it is.    

Add a little bit more for the offline context describing how it worksso that text flows together.

***
<a href="https://pr-preview.s3.amazonaws.com/rtoy/web-audio-api/pull/1930.html" title="Last updated on Jun 4, 2019, 5:49 PM UTC (8164d8a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1930/f81fa7f...rtoy:8164d8a.html" title="Last updated on Jun 4, 2019, 5:49 PM UTC (8164d8a)">Diff</a>


index.bs
```diff
@@ -10775,21 +10775,15 @@ in the algorithm of <a href="#rendering-a-graph">rendering a graph</a>.
...
-	some time, the underlying audio system will call the callback
-	again, and the algorithm will be executed again. This is an
-	implementation detail that should not be observable, apart from the
-	latency implications.
-
-	{{OfflineAudioContext}} will execute the algorithm continuously,
-	until {{OfflineAudioContext/length}} frames have been rendered.
+	an isochronous fashion. 
+
+	An {{OfflineAudioContext}} normally doesn't have this
```
hoch: What does it mean by "normally" here?

rtoy: Maybe we can say an oac is not required to a system-level audio callback?

rtoy: PTAL at the updated sentence.

PR #1940, opened 2019-06-03
Fix #1858: Multiline source text loses space

### LINKED ISSUE #1858: Multiline source text loses space between words
I filed tabatkins/bikeshed#1470 for this.  I think it's a bikeshed problem with wrapping lines, but until that gets fixed, I think we need to make the source text be one giant line.  If we don't, the spec looks embarrassingly bad with missing spaces.

rtoy:
Use the workaround and make the paragraph all on one line.  This fixes
the issue.

I only looked through the Description column of parameter descriptions
for methods.  These are all I could find; there could be others or in
other places, but I didn't see any when proofreading the whole spec a
while ago.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -854,13 +854,10 @@ Methods</h4>
 		nominal range.</span>
 
 		<pre class=argumentdef for="BaseAudioContext/createBuffer()">
-		numberOfChannels: Determines how many channels the buffer will have. An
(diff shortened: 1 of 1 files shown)


rtoy:
TBR @hoch 

Since this is purely editorial with no change in phrasing or meaning, I'm landing this.

PR #1941, opened 2019-06-03
Fix #1893: active processing for panner/convolver/merger

### LINKED ISSUE #1893: Active processing for Panner/Convolver/ChannelMerger
Although it is specified that nodes that are not actively processing produce a single channel of silence, we need to make sure the various nodes alloow for this.

For example, [PannerNode](https://webaudio.github.io/web-audio-api/#pannernode), paragraph 3 says:

> The output of this node is hard-coded to stereo (2 channels) and cannot be configured.

That's inconsistent with nodes that aren't actively processing.  We should update the PannerNode, the ConvolverNode, and the ChannelMergerNode to allow for this.

rtoy:
State that the number of chanels in the output is 1 if the node is not
actively processing any inputs.

***
<a href="https://pr-preview.s3.amazonaws.com/rtoy/web-audio-api/pull/1941.html" title="Last updated on Jun 4, 2019, 5:15 PM UTC (d073d57)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1941/5fb8063...rtoy:d073d57.html" title="Last updated on Jun 4, 2019, 5:15 PM UTC (d073d57)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -6241,7 +6241,9 @@ This interface represents an {{AudioNode}} for
 combining channels from multiple audio streams into a single audio
 stream. It has a variable number of inputs (defaulting to 6), but not
 all of them need be connected. There is a single output whose audio
-stream has a number of channels equal to the number of inputs.
+stream has a number of channels equal to the number of inputs when any
+of the inputs is [=actively processing=].  If none of the inputs are 
+[=actively processing=], then output is a single channel of silence.
 
 To merge multiple inputs into one stream, each input gets downmixed
 into one channel (mono) based on the specified mixing rule. An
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -6762,6 +6764,10 @@ Developers desiring more complex and arbitrary matrixing can use a
 {{ConvolverNode}}s and a
 {{ChannelMergerNode}}.
 
+If this node is not [=actively processing=], the output is a single
+channel of silence, even if the diagrams below show two channels of
+output.
```
hoch: Nit: The diagram is supposed to be "supportive" of the text content. Can we not mention the diagram at all? We can add one more diagram for the case of "single channel of silence", but that's a really low priority.

rtoy: Just added a note that the diagrams show the active processing result.

PR #1999, opened 2019-07-15
Replace `var` to `const` or `let` in sample code

Korilakkuma:
The modern browsers that can use Web Audio API can also use `const` or `let`.
Therefore, legacy code like `var` avoid as much as possible.

***
<a href="https://pr-preview.s3.amazonaws.com/Korilakkuma/web-audio-api-1/pull/1999.html" title="Last updated on Jul 17, 2019, 11:16 PM UTC (b295e13)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/1999/b45b2a7...Korilakkuma:b295e13.html" title="Last updated on Jul 17, 2019, 11:16 PM UTC (b295e13)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -333,10 +333,10 @@ containing a single {{AudioDestinationNode}}:
 Illustrating this simple routing, here's a simple example playing a single sound:
 
 <pre class="example" highlight="js">
-var context = new AudioContext();
+const context = new AudioContext();
 
 function playSound() {
-	var source = context.createBufferSource();
+	const source = context.createBufferSource();
 	source.buffer = dogBarkingBuffer;
 	source.connect(context.destination);
 	source.start(0);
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -10377,16 +10377,16 @@ a lower bit-depth), and by quantizing in time resolution
 {{AudioWorkletProcessor}}.
 
 <pre line-numbers class="example" highlight="js" title="BitCrusher - Global Scope">
-let context = new AudioContext();
+const context = new AudioContext();
 context.audioWorklet.addModule('bitcrusher.js').then(() =&gt; {
-	let osc = new OscillatorNode(context);
-	let amp = new GainNode(context);
+	const osc = new OscillatorNode(context);
```
Korilakkuma: It is a little problem that constructor and factory method of `AudioNode` mix in sample code. I think that it is better to unify constructor or factory method than this.

hoch:

index.bs
hoch: We should use the constructor where it is possible.

hoch:
Thanks for your contribution!

PR #2000, opened 2019-07-17
Revise the initialization of AudioWorkletProcessor.port

### LINKED ISSUE #1973: VUMeter example uses AudioWorkletProcessor's port in a way that can't work per spec
https://webaudio.github.io/web-audio-api/#vu-meter-mode has this bit:
```
registerProcessor('VUMeter', class extends AudioWorkletProcessor {
...
  constructor (options) {    
    super(options);
    this.port.onmessage = event => {
...
```
but this can't possibly work.  Indeed, let's consider what happens when `new VUMeterNode` is called on the control thread.  That lands in https://webaudio.github.io/web-audio-api/#instantiation-of-AudioWorkletNode-and-AudioWorkletProcessor the "In order to process a control message for the construction of an AudioWorkletProcessor" steps, right?

Now step 4 does `Construct(processorCtor, options)`, where `processorCtor` is in this case the class declaration above.  That will invoke the `constructor` function defined above.  The `super(options)` call lands in https://webaudio.github.io/web-audio-api/#dom-audioworkletprocessor-audioworkletprocessor which does not initialize the "port" member in any way as far as I can tell.  That initialization happens in step 5 of the "process a control message for the construction of an AudioWorkletProcessor" steps, which is _after_ the scripted constructor has already returned.

So what happens when the script then does `this.port`?  https://webaudio.github.io/web-audio-api/#AudioWorkletProcessor-attributes seems to assume that the port is always initialized when the getter can be called, but that's clearly not the case.

The only way I see to make this (that is, being able to use the port in constructors of subclasses of `AudioWorkletProcessor`) work is to make the port a required argument to the `AudioWorkletProcessor` constructor, either directly or via the existing options object.  Then the constructor could set up the port, and there would be no issue here.

@padenot @karlt 

P.S. I looked at what Chrome implements, and it simply looks nothing like the spec.  What they do is effectively store the port in a slot on the global right before step 4 of "process a control message for the construction of an AudioWorkletProcessor" and putt it out of there in the `AudioWorkletProcessor` constructor, so it's initialized by the time the `super()` call returns.  If the expectation is that UAs need to do that, then the spec should say so instead of saying something totally different.

hoch:
Fixes #1973.

Break down the processor constructor into two parts:
  1. Describe some implicit initialization in super() constructor. (port, node reference and etc)
  2. Then invoke the user-supplied constructor if it exists.

***
<a href="https://pr-preview.s3.amazonaws.com/hoch/web-audio-api/pull/2000.html" title="Last updated on Jul 17, 2019, 6:22 PM UTC (73c367b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/2000/b45b2a7...hoch:73c367b.html" title="Last updated on Jul 17, 2019, 6:22 PM UTC (73c367b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -10320,7 +10320,7 @@ thread and the rendering thread.
 			1. Set {{AudioWorkletProcessor/port}} field to
 				<var>processorPort</var>.
 
-			1. Set {{[[node reference]}} to <var>node</var>.
+			1. Set {{[[node reference]]}} to <var>node</var>.
 
 			1. Set {{[[callable process]]}} to `true`.
 
--- a/index.bs
+++ b/index.bs
@@ -10320,7 +10320,7 @@ thread and the rendering thread.
 			1. Set {{AudioWorkletProcessor/port}} field to
 				<var>processorPort</var>.
 
-			1. Set {{[[node reference}} to <var>node</var>.
+			1. Set {{[[node reference]}} to <var>node</var>.
 
 			1. Set {{[[callable process]]}} to `true`.
 
--- a/index.bs
+++ b/index.bs
@@ -10315,7 +10315,7 @@ thread and the rendering thread.
 	1. Perform Construct(<var>processorCtor</var>, « <var>options</var> »).
 
 		1. During the execution of the AudioWorkletProcessor super class
-		constructor, perform the followings:
+			constructor, perform the followings:
 
 			1. Set {{AudioWorkletProcessor/port}} field to
 				<var>processorPort</var>.
--- a/index.bs
+++ b/index.bs
@@ -10290,11 +10290,12 @@ thread and the rendering thread.
 	{{AudioWorkletProcessor}}, given a string <var>nodeName</var>, a
 	serialization record <var>processorPortSerialization</var>, and an
 	{{AudioWorkletNode}} <var>node</var>, perform the following
-	steps on the <a>rendering thread</a>. If any of these steps throws
-	an exception (either explicitly or implicitly), abort the rest of
-	steps and <a>queue a task</a> on the <a>control thread</a> to fire
-	{{AudioWorkletNode/onprocessorerror}} event to
-	<var>node</var>.
+	steps on the <a>rendering thread</a>.
+
+	If any of these steps throws an exception (either explicitly or
+	implicitly), abort the rest of steps and <a>queue a task</a> on the
+	<a>control thread</a> to fire {{AudioWorkletNode/onprocessorerror}}
+	event to <var>node</var>.
 
 	1. Let <var>processorPort</var> be
 		[$StructuredDeserializeWithTransfer$](<var>processorPortSerialization</var>,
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -10311,13 +10312,19 @@ thread and the rendering thread.
 	1. Let <var>processor</var> be the result of
 		Construct(<var>processorCtor</var>, « <var>options</var> »).
 
-	1. Set <var>processor</var>'s {{AudioWorkletProcessor/port}} to
-		<var>processorPort</var>.
+	1. Perform Construct(<var>processorCtor</var>, « <var>options</var> »).
 
-	1. Set <var>processor</var>'s {{[[node reference]]}}to
-		<var>node</var>.
+		1. During the execution of the AudioWorkletProcessor super class
```
bzbarsky: Shouldn't these steps actually go inside that constructor?  That would be a lot more clear in terms of what needs to happen.  The way this is written right now, someone implementing that constructor needs to know about these steps here that monkeypatch it...

If the steps were there, that would also make it a lot more clear how things should work if multiple `AudioWorkletProcessor` constructors run under the `Construct` call here, etc.

This does require making it explicit how the information propagates from here to the relevant constructor, but that's a feature in this case, I would think, because it enables implementations to actually do this interoperably.

hoch: > Shouldn't these steps actually go inside that constructor?

That's why I wrote "During the execution of...". The substeps are supposed to be the implicit part of `super()` call. After the `super()` constructor, these fields are ready to use.

> The way this is written right now, someone implementing that constructor needs to know about these steps here that monkeypatch it...

I think of it as a contract of using AudioWorkletProcessor class. These are features provided by UA. Not sure what you meant by monkeypatching. Some of the references are hidden and only used internally.

> how things should work if multiple AudioWorkletProcessor constructors run under the Construct call here

What do you mean by "multiple AudioWorkletProcessor constructors"?

> because it enables implementations to actually do this interoperably.

Do you see that these substeps can be the source of interoperability issues?

bzbarsky: > That's why I wrote "During the execution of...". The substeps are supposed to be the implicit part of `super()` call.

I don't understand why they are implicit.  Why not just make them explicit somewhere in <https://webaudio.github.io/web-audio-api/#dom-audioworkletprocessor-audioworkletprocessor>? 

> I think of it as a contract of using AudioWorkletProcessor class. These are features provided by UA. Not sure what you meant by monkeypatching. Some of the references are hidden and only used internally.

I mean how is the UA engineer implementing https://webaudio.github.io/web-audio-api/#dom-audioworkletprocessor-audioworkletprocessor supposed to know that in addition to the 3 steps it lists these other steps also need to happen at that same point?

> What do you mean by "multiple AudioWorkletProcessor constructors"?
```
  class A extends AudioWorkletProcessor {};
  class B extends AudioWorkletProcessor {
    constructor() {
      let x = new A();
      super();
      let y = new A();
    }
  };
```
If `B` is now registered as a processor, when we do the `Construct` call for it, we will get three executions of the `AudioWorkletProcessor` constructor: one under the `new A()` call that sets `x`, one for the `super()` call, and another for the `new A()` call that sets `y`.  Which, if any, of those `AudioWorkletProcessor` instances should have `.port` set to the given port and why?  The proposed spec text does not make that clear at all.

> > because it enables implementations to actually do this interoperably.
> 
> Do you see that these substeps can be the source of interoperability issues?

Yes, of course.  See my example above.

hoch: > I don't understand why they are implicit. Why not just make them explicit somewhere in https://webaudio.github.io/web-audio-api/#dom-audioworkletprocessor-audioworkletprocessor?

By adding `port` as an explicit argument of AudioWorkletProcessor constructor? I am really reluctant to change the constructor interface at this point. If you want me to move those substeps to the AWP constructor algorithm, I can try that.

> If B is now registered as a processor, when we do the Construct call for it, we will get three executions of the AudioWorkletProcessor constructor: one under the new A() call that sets x...

The AWP constructor step 1 does not allow the direction invocation of AWP constructor. I understand that you do not agree with that design, but the current spec does not allow such construction.

bzbarsky: > By adding `port` as an explicit argument of AudioWorkletProcessor constructor? 

No, by being explicit in the `AudioWorkletProcessor` constructor about how it determines the port.

> If you want me to move those substeps to the AWP constructor algorithm, I can try that.

Yes, that is what I want you to do.  That will force you to think about how the constructor will actually get this data, and what happens if multiple constructors run.

> The AWP constructor step 1 does not allow the direction invocation of AWP constructor.

My test code doesn't do direct invocation; it only constructs subclasses.  Everything it does is perfectly allowed by the spec as it is written right this second.

hoch: Thanks for your help! I'll follow up on your suggestion with a different PR.

hoch:
This approach doesn't seem to work. Closing.

PR #2002, opened 2019-07-18
Fix #1943: Describe how WaveShaper curve works

### LINKED ISSUE #1943: Describe how WaveShaperNode shapes the input with the curve
The description of the [WaveShaperNode](https://webaudio.github.io/web-audio-api/#waveshapernode) says linear interpolation is used (see also #87).  But the fine details are a bit lacking.

Perhaps a formula and/or picture would make it clearer on how it's done.  At least for the case of `oversample` = `none`.  The oversampling cases are purposely rather loose so you can do the up-sampling and down-sampling in whatever way.  But the application of the curve is basically the same. in all cases.

rtoy:
Add some formulas to show how to compute an output values from an
input value using the given curve.

Also updated MathJax to 2.7.5 from 2.6.1.  Version 2.6.1 was producing
some stack overflow processing our doc.  Version 2.7.5 doesn't have
that problem.

***
<a href="https://pr-preview.s3.amazonaws.com/rtoy/web-audio-api/pull/2002.html" title="Last updated on Jul 18, 2019, 4:59 PM UTC (cef27a5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/2002/b45b2a7...rtoy:cef27a5.html" title="Last updated on Jul 18, 2019, 4:59 PM UTC (cef27a5)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -9415,15 +9415,15 @@ Attributes</h4>
 		1] range. This means that a {{WaveShaperNode/curve}} with a
 		even number of value will not have a value for a signal at
 		zero, and a {{WaveShaperNode/curve}} with an odd number of
-		value will have a value for a signal at zero.  the
+		value will have a value for a signal at zero.  The
 		output is determined by the following algorithm.
 
 		<div algorithm="wave-shaper-curve">
 			1. Let \(x\) be the input sample, \(y\) be the
 				corresponding output of the node,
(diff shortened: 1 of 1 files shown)


rtoy:
It's probably best to use the preview to see what's changed and to see the formulas nicely.

PR #2003, opened 2019-07-18
Use ErrorEvent against AudioWorkletNode.onprocessorerror

### LINKED ISSUE #1984: Handling of exceptions in audio worklet is not very clear
https://webaudio.github.io/web-audio-api/#dom-audioworkletnode-onprocessorerror says:

> When an exception is thrown from the processor’s constructor, process method, or any user-defined class method, the processor will queue a task on the control thread to fire onprocessorerror event to the node. Note that once an exception is thrown, the processor will output silence throughout its lifetime.

First, I assume it means that an event of type "processorerror" is fired, even though that's not what it says, right?

Second, is this event just an `Event` or does it implement some other event interface?  Is there a reason to not make this an `ErrorEvent`, and name the attribute "onerror"?

hoch:
Fixes #1984 based on https://github.com/WebAudio/web-audio-api/issues/1984#issuecomment-512516835.

***
<a href="https://pr-preview.s3.amazonaws.com/hoch/web-audio-api/pull/2003.html" title="Last updated on Jul 19, 2019, 3:55 PM UTC (19cb50c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/2003/06882c0...hoch:19cb50c.html" title="Last updated on Jul 19, 2019, 3:55 PM UTC (19cb50c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -9837,7 +9837,7 @@ Attributes</h5>
 		<code>constructor</code>, <code>process</code> method,
 		or any user-defined class method, the processor will
 		<a>queue a task</a> to fire an
-		<a href="https://www.w3.org/TR/html50/webappapis.html#the-errorevent-interface">
+		<a href="https://html.spec.whatwg.org/multipage/webappapis.html#the-errorevent-interface">
 		ErrorEvent</a> to the associated {{AudioWorkletNode}}.
 
 		The <code>ErrorEvent</code> is created and initialized
--- a/index.bs
+++ b/index.bs
@@ -9833,7 +9833,7 @@ Attributes</h5>
 <dl dfn-type=attribute dfn-for="AudioWorkletNode">
 	: <dfn>onprocessorerror</dfn>
 	::
-		When an exception is thrown from the processor's
+		When an unhandled exception is thrown from the processor's
 		<code>constructor</code>, <code>process</code> method,
 		or any user-defined class method, the processor will
 		<a>queue a task</a> to fire an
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -9836,10 +9836,17 @@ Attributes</h5>
 		When an exception is thrown from the processor's
 		<code>constructor</code>, <code>process</code> method,
 		or any user-defined class method, the processor will
-		<a>queue a task</a> on the control thread to fire
-		{{onprocessorerror}} event to the node. Note that
-		once an exception is thrown, the processor will output
-		silence throughout its lifetime.
+		<a>queue a task</a> to fire an
+		<a href="https://www.w3.org/TR/html50/webappapis.html#the-errorevent-interface">
```
padenot: <https://html.spec.whatwg.org/multipage/webappapis.html#the-errorevent-interface> is better, right?

index.bs
```diff
@@ -9836,10 +9836,17 @@ Attributes</h5>
 		When an exception is thrown from the processor's
 		<code>constructor</code>, <code>process</code> method,
 		or any user-defined class method, the processor will
-		<a>queue a task</a> on the control thread to fire
-		{{onprocessorerror}} event to the node. Note that
-		once an exception is thrown, the processor will output
-		silence throughout its lifetime.
+		<a>queue a task</a> to fire an
```
padenot: What if the exception is handled? Is there a specific formulation to refer to this case?

hoch: A good point. I think we should only deliver the exception to the node when it is "unhandled". We could throw two exceptions on both sides, but it seems redundant and not particularly useful.

Issue #1972, opened 2019-07-05
parameterDescriptors handling during AudioWorkletNode initialization is probably wrong


bzbarsky:
https://webaudio.github.io/web-audio-api/#instantiation-of-AudioWorkletNode-and-AudioWorkletProcessor step 9.3 says:

> For each key-value pair (paramNameInOption to value) of options:

At this point _options_ is a `AudioWorkletNodeOptions` instance, so the only possible keys here are "numberOfInputs", "numberOfOutputs", "outputChannelCount", "parameterData", and "processorOptions", right?  But the various spec examples use all sorts of custom param names like "myParam".  Do those just not affect AudioParam values here?   Am I totally misreading what's going on?

Also, what does the "value is not a type of Number" bit even mean if the values are IDL types?

An informative note about what this is really trying to do would maybe help.... I can't shake the feeling that this is not working with the object it means to be working with.

karlt:
I assume the intention was that the key-value pairs should come from the value of the `parameterData` member on _options_.

"not a type of Number" can be dropped, I assume, because _value_ is of IDL type `double`.

hoch:
I'll update the example code based on the updated spec text.

hoch:
> But the various spec examples use all sorts of custom param names like "myParam". 

The incorrect usage in code example will be fixed by #1994.

> Do those just not affect AudioParam values here?

Yes, by passing |options| to `super()` in the constructor of AudioWorkletProcessor.

> Also, what does the "value is not a type of Number" bit even mean if the values are IDL types?

If the user-code does something like this, it will be ignored even if `myParam` is a valid entry in the retrieved `AudioParamMap` in the step 9.1. ([algorithm](https://webaudio.github.io/web-audio-api/#instantiation-of-AudioWorkletNode-and-AudioWorkletProcessor))
```js
parameterData: {
  myParam: 'hey'
}
```

I think we can drop the second bullet point about "not a type of Number" in the step.

bzbarsky:
> Yes, by passing |options| to super() in the constructor of AudioWorkletProcessor.

I don't understand this part, especially as it relates to the changes in <https://github.com/WebAudio/web-audio-api/pull/1994>.

I suspect that the real key here is https://github.com/WebAudio/web-audio-api/issues/1972#issuecomment-509137611 paragraph 1...

>   myParam: 'hey'

[...]

Issue #1963, opened 2019-07-05
NewTarget check for AudioWorkletProcessor isn't actually possible with a Web IDL constructor


bzbarsky:
https://webaudio.github.io/web-audio-api/#AudioWorketProcessor-constructors step 1 says:

> Compare the value of NewTarget with the active function object; if the two are equal, throw a TypeError.

but there is no NewTarget or "active function object" available in a Web IDL constructor's steps.  Those steps are handed the already-created to-be-initialized object and IDL values for the arguments; see https://heycam.github.io/webidl/#create-an-interface-object step 9.

Presumably this is trying to do something like what `[HTMLConstructor]` does, but note that this defined a separate extended attribute, with different semantics from IDL's `[Constructor]`.  

Is there a specific reason to disallow directly constructing `AudioWorkletProcessor` instances?

hoch:
Previously: https://github.com/WebAudio/web-audio-api/issues/1447

bzbarsky:
Yeah, so https://github.com/WebAudio/web-audio-api/issues/1447#issuecomment-344034925 was just not correct.  :(

hoch:
> Is there a specific reason to disallow directly constructing AudioWorkletProcessor instances?

Yes. AudioWorkletProcessor (render thread) is a counterpart of AudioWorkletNode (main thread). Without a node, a processor is meaningless because it can't be a part of the audio graph.

As we already have seen, the current text is based on https://github.com/WebAudio/web-audio-api/issues/1447#issuecomment-344034925, and we'll appreciate the guidance to fix this. Perhaps things have changed over 2 years...

bzbarsky:
> Without a node, a processor is meaningless because it can't be a part of the audio graph.

OK, but is that a reason to prevent it being constructed, at the cost of specification and implementation complexity?  If someone just constructs a processor, they can't do anything useful with it.  Fine.  That means no one will do that in practice...

But also, this approach doesn't actually solve that problem as described.   If we take that current spec at face value, and code in the worklet does:
```
    class Foo extends AudioWorkletProcessor {};
    let foo = new Foo;
```
that _is_ allowed by the current spec (because now `NewTarget` is `Foo` so https://webaudio.github.io/web-audio-api/#dom-audioworkletprocessor-audioworkletprocessor step 1 does not throw) and results in an `AudioWorkletProcessor` instance that is not hooked up to an `AudioWorkletNode`.

> the current text is based on #1447 (comment)

[...]

PR #2005, opened 2019-07-18
Split the instantiation algorithm of AWN/AWP into each constructor

hoch:
PTAL - I would like to tackle this over multiple commits. Some changes are coupled tightly, so it's hard to break down.

## DONE
- Initialize `AWP.port` explicitly (#1973)
- Remove `NewTarget` and `active function object` bit (#1963)
- Revise the usage of `parameterDescriptor` in `AudioWorkletNode` constructor. (#1972)

***
<a href="https://pr-preview.s3.amazonaws.com/hoch/web-audio-api/pull/2005.html" title="Last updated on Aug 8, 2019, 6:22 PM UTC (f5cf604)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/2005/7787d62...hoch:f5cf604.html" title="Last updated on Aug 8, 2019, 6:22 PM UTC (f5cf604)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -9853,13 +9853,6 @@ Every {{AudioWorkletProcessor}} has an associated <dfn>active source</dfn> flag,
 the node to be retained in memory and perform audio processing in
 the absence of any connected inputs.
 
-Once the construction of the node is completed,
-<dfn>processor construction data</dfn> will be created and
-transferred to the matching {{AudioWorkletProcessor}}'s
-{{AudioWorkletProcessor()|constructor}}. This data contains
-a set of objects and references that is required for
-{{AudioWorkletProcessor}}'s construction.
-
 <xmp class="idl">
 [Exposed=Window]
 interface AudioParamMap {
@@ -9894,6 +9887,14 @@ Constructors</h5>
 			name: A string that is a key for the {{BaseAudioContext}}’s <a>node name to parameter descriptor map</a>.
 			options: Optional initial parameters value for this {{AudioWorkletNode}}.
 		</pre>
+
+		Once the construction of the node is completed,
+		<dfn dfn>processor construction data</dfn> will be created and
+		transferred to the matching {{AudioWorkletProcessor}}'s
+		{{AudioWorkletProcessor()|constructor}}. This data contains
+		a set of objects and references that is required for
+		{{AudioWorkletProcessor}}'s construction. See the step 12 in
+		the algorithm below.
 
 		When the constructor was called, the user agent MUST perform the
 		following steps on the control thread:
--- a/index.bs
+++ b/index.bs
@@ -9969,7 +9969,7 @@ Constructors</h5>
 		1. <a>Queue a control message</a> to invoke the
 			{{AudioWorkletProcessor()|constructor}} of
 			the corresponding {{AudioWorkletProcessor}} with
-			the {{processor construction data}} that consists of:
+			the [=processor construction data=] that consists of:
 			<var>nodeName</var>,
 			<var>serializedProcessorPort</var>,
 			<var>serializedOptions</var>,
@@ -10158,14 +10158,14 @@ Constructors</h5>
 			This constructor is invoked by processing a control
 			message queued by the {{AudioWorkletNode()|constructor}}
 			of the associated {{AudioWorkletNode}}. This control
-			message carries the {{processor construction data}} that
+			message carries the [=processor construction data=] that
 			consists of:
 			<var>nodeName</var>,
 			<var>serializedProcessorPort</var>,
 			<var>serializedOptions</var>,
 			and <var>node</var>.
 
-			1. If there is no processor construction data
+			1. If there is no [=processor construction data=]
 				passed from the associated {{AudioWorkletNode}},
 				throw a {{TypeError}}.
 
--- a/index.bs
+++ b/index.bs
@@ -9855,7 +9855,7 @@ the absence of any connected inputs.
 
 Once the construction of the node is completed,
 <dfn>processor construction data</dfn> will be created and
-trasnferred to the matching {{AudioWorkletProcessor}}'s
+transferred to the matching {{AudioWorkletProcessor}}'s
 {{AudioWorkletProcessor()|constructor}}. This data contains
 a set of objects and references that is required for
 {{AudioWorkletProcessor}}'s construction.
@@ -10181,13 +10181,6 @@ Constructors</h5>
 		When the constructor for {{AudioWorkletProcessor}} is invoked,
 		the following steps are performed on the <a>rendering thread</a>.
 
-		If any of these steps throws an exception, abort the rest of
-		steps and <a>queue a task</a> to fire an
-		<a href="https://www.w3.org/TR/html50/webappapis.html#the-errorevent-interface">ErrorEvent</a>
-		named <code>processorerror</code> at the {{AudioWorkletNode}}
-		on the <a>control thread</a> with the relevant information about
-		the error.
-
 		<div algorithm="AudioWorkletProcessor()">
 			This constructor is invoked by processing a control
 			message queued by the {{AudioWorkletNode()|constructor}}
@@ -10198,6 +10191,13 @@ Constructors</h5>
 			<var>serializedProcessorPort</var>,
 			<var>serializedOptions</var>,
 			and <var>node</var>.
+
+			If any of these steps throws an exception, abort the rest of
+			steps and <a>queue a task</a> to fire an
+			<a href="https://www.w3.org/TR/html50/webappapis.html#the-errorevent-interface">ErrorEvent</a>
+			named <code>processorerror</code> at the {{AudioWorkletNode}}
+			on the <a>control thread</a> with the relevant information about
+			the error.
 
 			1. If there is no [=processor construction data=]
 				passed from the associated {{AudioWorkletNode}},
--- a/index.bs
+++ b/index.bs
@@ -9910,7 +9910,7 @@ Constructors</h5>
 				<var>parameterDescriptors</var>:
 
 				1. Let <var>paramName</var> be the value of
-					<var>name</var> member in
+					{{AudioParamDescriptor/name}} member in
 					<var>descriptor</var>.
 
 				1. Let <var>audioParam</var> be a new
--- a/index.bs
+++ b/index.bs
@@ -9950,19 +9950,14 @@ Constructors</h5>
 						of <var>audioParamInMap</var>
 						to <var>paramValue</var>.
 
-			1. For each key-value pair (<var>paramNameInOption</var> to
-				<var>value</var>) of <var>options</var>:
-				1. If there exists an entry with name member equal to
-					<var>paramNameInOption</var> inside
-					<var>audioParamMap</var>, set that {{AudioParam}}'s
-					value to <var>value</var>.
+			1. For each key-value pair <var>paramNameInOption</var>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -9844,19 +9844,105 @@ Constructors</h5>
...
 	: <dfn>AudioWorkletNode(context, name, options)</dfn>
 	::
-		Let <var>node</var> be a new {{AudioWorkletNode}} object.
-		<a href="#audionode-constructor-init">Initialize</a>
-		<var>node</var>. Perform the <a href="#instantiation-of-AudioWorkletNode-and-AudioWorkletProcessor">
-		construction procedure</a> of an
-		{{AudioWorkletNode}} and the corresponding
-		{{AudioWorkletProcessor}} object. Return
-		<var>node</var>.
+		When the constructor of {{AudioWorkletNode}} ot its subclass is
```
padenot: `s/ot/to/`. But I don't understand the "to its subclass" part.

index.bs
```diff
@@ -10008,35 +10099,62 @@ interface AudioWorkletProcessor {
...
+			message queued by the
+			{{AudioWorkletNode()|constructor}} of the associated
+			{{AudioWorkletNode}}. This control message carries
+			the processor construction data:
+			<var>nodeName</var>,
+			<var>serializedProcessorPort</var>,
+			<var>serializedOptions</var>,
+			and <var>node</var>.
+
+			1. If there is no processor construction data
```
padenot: This needs a link that says what are "processor construction data".

index.bs
```diff
@@ -10008,35 +10099,62 @@ interface AudioWorkletProcessor {
 
 <h5 id="AudioWorketProcessor-constructors">
 Constructors</h5>
+
 <dl dfn-type="constructor" dfn-for="AudioWorkletProcessor">
 	: <dfn>AudioWorkletProcessor(options)</dfn>
 	::
-		When the constructor for {{AudioWorkletProcessor}} is invoked, the following steps are performed:
+		When the constructor for {{AudioWorkletProcessor}} is invoked,
+		the following steps are performed steps on the
```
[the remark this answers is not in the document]
hoch: Oops. Done.

index.bs
```diff
@@ -10008,35 +10099,62 @@ interface AudioWorkletProcessor {
...
+		<a>rendering thread</a>.
+
+		If any of these steps throws an exception, abort the rest of
+		steps and <a>queue a task</a> to inform the associated
+		{{AudioWorkletNode}} on the <a>control thread</a> with the
+		relevant information about the error. In turn, the
+		{{AudioWorkletNode}} will fire an
+		<a href="https://www.w3.org/TR/html50/webappapis.html#the-errorevent-interface">ErrorEvent</a>
+		via the {{AudioWorkletNode/onprocessorerror}} event handler.
```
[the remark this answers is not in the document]

index.bs
hoch: The problem here is that `<dfn>` tag does not work in the algorithm section. The definition is at the l.10123 above.

index.bs
```diff
@@ -10015,35 +10135,60 @@ interface AudioWorkletProcessor {
 
 <h5 id="AudioWorketProcessor-constructors">
 Constructors</h5>
+
 <dl dfn-type="constructor" dfn-for="AudioWorkletProcessor">
 	: <dfn>AudioWorkletProcessor(options)</dfn>
 	::
-		When the constructor for {{AudioWorkletProcessor}} is invoked, the following steps are performed:
+		When the constructor for {{AudioWorkletProcessor}} is invoked,
+		the following steps are performed on the <a>rendering thread</a>.
```
padenot: Is this mandatory ? It would be best if this was not done on an RT thread, right?

index.bs
```diff
@@ -10015,35 +10135,60 @@ interface AudioWorkletProcessor {
...
+
+		If any of these steps throws an exception, abort the rest of
+		steps and <a>queue a task</a> to fire an
+		<a href="https://www.w3.org/TR/html50/webappapis.html#the-errorevent-interface">ErrorEvent</a>
+		named "processorerror" at the AudioWorkletNode on the
```
padenot: Link `AudioWorkletNode` and put code tag around `"processorerror"`.

index.bs
padenot: We can't leave it like that, surely there is a way?

index.bs
hoch: Hmm. AudioWorkletProcessor belongs to the rendering thread. How would you use this object if this is created in somewhere else? Are you creating this from the control and transfer it?

index.bs
hoch: I defined "processor construction data" in the AWN's constructor description.

index.bs
hoch: If you want to have the flexibility of "construction", you might want to create a follow-up PR after this one.

index.bs
```diff
@@ -9844,19 +9849,134 @@ Constructors</h5>
...
-		<var>node</var>. Perform the <a href="#instantiation-of-AudioWorkletNode-and-AudioWorkletProcessor">
-		construction procedure</a> of an
-		{{AudioWorkletNode}} and the corresponding
-		{{AudioWorkletProcessor}} object. Return
-		<var>node</var>.
+		When the constructor of {{AudioWorkletNode}} is invoked in the
+		main global scope, the following construction algorithm MUST be
+		performed.
+
+		During the construction of an AudioWorkletNode, a
```
rtoy: In line 9853 you say "algorithm".  It would be nice if this paragraph were restructured to be an algorithm with steps.  But it's ok with me if you don't want to do that. (I'm not sure how to make it an algorithm with steps.)

rtoy: Oh. Maybe move the argument block further up or down so that the algorithm steps are closer to this paragrph.

index.bs
```diff
@@ -9844,19 +9849,134 @@ Constructors</h5>
...
+		corresponding {{AudioWorkletProcessor}} instance is also
+		automatically created in the {{AudioWorkletGlobalScope}}.
+		Note that the instantiation of these object pairs spans the
+		control thread and the rendering thread.
```
rtoy: Maybe this is step 0 of the algorithm below?

index.bs
```diff
@@ -9844,19 +9849,134 @@ Constructors</h5>
...
+		{{AudioWorkletNode/AudioWorkletNode(context, name, options)/options}})
+		constructor is invoked, the user agent MUST perform the
+		following steps on the control thread, where the constructor was
+		called.
+
+		1. Perform the validity check on <var>options</var>. If the
+			check throws any exception, propagate the exception
+			and abort these steps.
+
+		1. If <var>nodeName</var> does not exists as a key in the
```
rtoy: Typo: "exists" -> "exist"

index.bs
```diff
@@ -10018,35 +10140,60 @@ interface AudioWorkletProcessor {
...
 <h5 id="AudioWorketProcessor-constructors">
 Constructors</h5>
+
 <dl dfn-type="constructor" dfn-for="AudioWorkletProcessor">
 	: <dfn>AudioWorkletProcessor(options)</dfn>
 	::
-		When the constructor for {{AudioWorkletProcessor}} is invoked, the following steps are performed:
+		When the constructor for {{AudioWorkletProcessor}} is invoked,
+		the following steps are performed on the <a>rendering thread</a>.
+
```
rtoy: This is basically a move of the existing algorithm from another section to this, right? No other changes?

hoch: It depends on what your "this" means here. This PR has few more changes integrated. I summarized them in the description.

But if you're asking about this L.10147 specifically - Yes. It's a simple move.

index.bs
hoch: It's already there at L.10198, but I wanted to emphasize the procedure because it is unconventional.

index.bs
```diff
@@ -9853,6 +9853,13 @@ Every {{AudioWorkletProcessor}} has an associated <dfn>active source</dfn> flag,
 the node to be retained in memory and perform audio processing in
 the absence of any connected inputs.
 
+Once the construction of the node is completed,
+<dfn>processor construction data</dfn> will be created and
+trasnferred to the matching {{AudioWorkletProcessor}}'s
```
rtoy: Typo: "trasnferred".

index.bs
```diff
@@ -9853,6 +9853,13 @@ Every {{AudioWorkletProcessor}} has an associated <dfn>active source</dfn> flag,
...
+{{AudioWorkletProcessor()|constructor}}. This data contains
```
rtoy: Maybe "The `<a>processor construction data</a> contains...`

index.bs
```diff
@@ -10056,35 +10174,60 @@ interface AudioWorkletProcessor {
...
 Constructors</h5>
+
 <dl dfn-type="constructor" dfn-for="AudioWorkletProcessor">
 	: <dfn>AudioWorkletProcessor(options)</dfn>
 	::
-		When the constructor for {{AudioWorkletProcessor}} is invoked, the following steps are performed:
+		When the constructor for {{AudioWorkletProcessor}} is invoked,
+		the following steps are performed on the <a>rendering thread</a>.
+
+		If any of these steps throws an exception, abort the rest of
```
rtoy: I think this paragraph should probably be placed inside the algorithm block, perhaps after the first paragraph of the algorithm.

index.bs
rtoy: Hmm.  This is ill-defined, because I can't tell what the data actually is and there's no link to what it might be.  But later on, in the AudioWorkletNode constructor algorithm, step 12, you say what the processor construction data actually is.  Thus, step 12 has actual definition of what the processor contruction data is.  Not sure if this is what you want, but this at least tells you what the data really is, instead of some abstract concept of data here.

index.bs
```diff
@@ -10056,35 +10174,60 @@ interface AudioWorkletProcessor {
...
+			message queued by the {{AudioWorkletNode()|constructor}}
+			of the associated {{AudioWorkletNode}}. This control
+			message carries the [=processor construction data=] that
+			consists of:
+			<var>nodeName</var>,
+			<var>serializedProcessorPort</var>,
+			<var>serializedOptions</var>,
+			and <var>node</var>.
+
+			1. If there is no [=processor construction data=]
```
rtoy: Is that actually possible? (No processor construction data)

index.bs
```diff
@@ -10056,35 +10174,60 @@ interface AudioWorkletProcessor {
...
+				Construct(<var>processorCtor</var>, « <var>options</var> »).
+
+			1. Set <var>processor</var>’s port to <var>processorPort</var>.
+
+			1. Set <var>processor</var>'s {{[[node reference]]}} to
+				<var>node</var>.
+
+			1. Set {{[[callable process]]}} to `true`.
+
+			1. Set <var>node</var>'s <a>processor reference</a> to
```
rtoy: Not relevant for this CL, but maybe we should make "processor reference" an internal slot like callable process, node reference, and so on.

index.bs
hoch: That was the original form, but the problem here is we can't define a thing in the middle of an algorithm. (a bikeshed rule) That's why the definition has to be moved to here.

index.bs

index.bs
hoch: This is the resolution from https://github.com/WebAudio/web-audio-api/issues/1963#issuecomment-510955630. This is a "magic" that prevents the user-supplied code from constructing an AudioWorkletProcessor.

index.bs

PR #2013, opened 2019-07-25
Remove 'nullalble' type and default 'null' value from AudioWorkletNodeOptions.processorOptions

### LINKED ISSUE #2011: AudioWorkletNodeOptions.processorOptions should not default to null.
**Describe the issue**
Because it is defaulting to `null`, the processor always get `null` for the field even the main scope option object is undefined. We should make it a plain object.
```webidl
object processorOptions;
```

**Where Is It**
https://webaudio.github.io/web-audio-api/#dictdef-audioworkletnodeoptions

hoch:
Fixes #2011, plus minor clean up.

***
<a href="https://pr-preview.s3.amazonaws.com/hoch/web-audio-api/pull/2013.html" title="Last updated on Jul 25, 2019, 10:46 PM UTC (4d83069)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/2013/31a8116...hoch:4d83069.html" title="Last updated on Jul 25, 2019, 10:46 PM UTC (4d83069)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -9951,6 +9951,7 @@ Dictionary {{AudioWorkletNodeOptions}} Members</h6>
 		{{AudioWorkletNode}}. If the string key of an entry in the
 		list does not match the name of any {{AudioParam}} objects in
 		the node, it is ignored.
+
 	: <dfn>processorOptions</dfn>
 	::
 		This holds any additional user-defined data that may be used to
--- a/index.bs
+++ b/index.bs
@@ -9913,7 +9913,7 @@ dictionary AudioWorkletNodeOptions : AudioNodeOptions {
 	unsigned long numberOfOutputs = 1;
 	sequence<unsigned long> outputChannelCount;
 	record<DOMString, double> parameterData;
-	object? processorOptions = null;
+	object processorOptions;
 };
 </xmp>
 
@@ -9953,7 +9953,9 @@ Dictionary {{AudioWorkletNodeOptions}} Members</h6>
 		the node, it is ignored.
 	: <dfn>processorOptions</dfn>
 	::
-		 This holds any additional user-defined data that may be used to initialize the corresponding {{AudioWorkletProcessor}} for this {{AudioWorkletNode}}.
+		This holds any additional user-defined data that may be used to
+		initialize the corresponding {{AudioWorkletProcessor}} for this
+		{{AudioWorkletNode}}.
 </dl>
 
 <h6 id="configuring-channels-with-audioworkletnodeoptions">


index.bs
```diff
@@ -9951,9 +9951,12 @@ Dictionary {{AudioWorkletNodeOptions}} Members</h6>
 		{{AudioWorkletNode}}. If the string key of an entry in the
 		list does not match the name of any {{AudioParam}} objects in
 		the node, it is ignored.
+
 	: <dfn>processorOptions</dfn>
 	::
-		 This holds any additional user-defined data that may be used to initialize the corresponding {{AudioWorkletProcessor}} for this {{AudioWorkletNode}}.
+		This holds any additional user-defined data that may be used to
+		initialize the corresponding {{AudioWorkletProcessor}} for this
+		{{AudioWorkletNode}}.
```
rtoy: What changed here?

hoch: There was a redundant space and it was a single long line. I think Bikeshed rendered this line incorrectly because of the space. Also I have been doing line-wrapping as much as possible.

rtoy: Please don't do that. It makes reviewing harder.  And also keep in mind:  https://github.com/tabatkins/bikeshed/issues/1470.  Until that's fixed, we should be very careful about wrapping lines.

hoch:
Sorry, but I have a different opinion. We should improve the work where we can. These kind of long lines make the readability significantly worse. The review might slightly be more difficult, but it's better to improve the code. In this case, it's only 3 lines. (not having a well-defined style guide was a big mistake in the first place)

Also I am not sure what the bug reported to Bikeshed was, but the preview looks good to me. I would like to merge as-is.

rtoy:
The bikeshed bug was that in some cases the source looked like
> foobar stuff
> new stuff

which gets rendered like

> foobar stuffnew stuff

which is pretty horrible if you should find it.

Reformatting lines is ok, but unless it's actually releveant to the change, we should refrain from doing it in the same PR.  It makes reviewing much harder.  And even if it is relevant, wrapping lines makes it much harder to review.  Don't make it even harder for the reviewers to review.

Reformat in a different PR.

Anyway, I already gave my lgtm earlier....

hoch:
> Reformatting lines is ok, but unless it's actually releveant to the change, we should refrain from doing it in the same PR.

The line-wrapping change is not random and relevant to the changed property at l.9916. Also I wouldn't say it is "much harder" to read 18 more words, but perhaps that's more of personal preference.

PR #2022, opened 2019-08-12
Revise AWN/AWP construction process

### LINKED ISSUE #2021: AudioWorkletProcessor constructor leads to infinite recursion
https://webaudio.github.io/web-audio-api/#dom-audioworkletprocessor-audioworkletprocessor step 5 constructs a subclass of `AudioWorkletProcessor`, which will call into the `AudioWorkletProcessor` constructor (via `super`), which will land in these steps again, etc.

This looks like it was introduced in https://github.com/WebAudio/web-audio-api/pull/2005 but the new setup looks completely broken to me...

hoch:
This is an attempt to fix #2021.

Key changes:
- A new algorithm is introduced: "invoking AWP constructor with processor construction data"
- The "processor construction data" is now defined clearly.
- The AWN constructor will now explicitly trigger the algorithm above.
- The AWP constructor does not have infinite loop anymore.

cc @bzbarsky

***
<a href="https://pr-preview.s3.amazonaws.com/hoch/web-audio-api/pull/2022.html" title="Last updated on Aug 15, 2019, 10:05 PM UTC (e9ce078)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/2022/6b1fd01...hoch:e9ce078.html" title="Last updated on Aug 15, 2019, 10:05 PM UTC (e9ce078)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -9827,10 +9827,6 @@ Methods</h5>
 <h5 id="AudioWorkletProcessor-instantiation">
 The instantiation of AudioWorkletProcessor</h5>
 
-The construction of an {{AudioWorkletNode}} is followed by the construction
-of a corresponding {{AudioWorkletProcessor}} instance in the
-{{AudioWorkletGlobalScope}}.
-
 At the end of the {{AudioWorkletNode}} construction,
 <dfn>processor construction data</dfn> will be prepared for the cross-thread
 transfer. This data contains:
@@ -9838,8 +9834,9 @@ transfer. This data contains:
 <dl>
 	: <dfn dfn>nodeName</dfn>
 	::
-		A {{DOMString}} of the {{AudioWorkletProcessor}} name that is to
-		be looked up <a>node name to processor constructor map</a>.
+		A {{DOMString}} which is the {{AudioWorkletProcessor}} name
+		that is to be looked up in the <a>node name to processor
+		constructor map</a>.
 	: <dfn dfn>nodePort</dfn>
 	::
 		A serialized {{MessagePort}} that is paired with
@@ -10237,24 +10234,24 @@ Constructors</h5>
 				invocation of the constructor throw a
 				{{TypeError}}.
 
-			1. Let <var>processor</var> be a new
+			1. Let <var>this</var> be a new
 				{{AudioWorkletProcessor}} instance.
 
-			1. Let <var>processorPort</var> be [$StructuredDeserializeWithTransfer$](
+			1. Let <var>processorPort</var> be
+				[$StructuredDeserializeWithTransfer$](
 				[=nodePort=], the current Realm).
 
-			1. Set <var>processor</var>’s {{AudioWorkletProcessor/port}}
+			1. Set <var>this</var>’s {{AudioWorkletProcessor/port}}
 				to <var>processorPort</var>.
 
-			1. Set <var>processor</var>'s {{[[node reference]]}} to
+			1. Set <var>this</var>'s {{[[node reference]]}} to
 				<var>node</var>.
 
-			1. Set <var>processor</var>'s {{[[callable process]]}} to `true`.
+			1. Set <var>this</var>'s {{[[callable process]]}} to
+				`true`.
 
 			1. Set <var>node</var>'s <a>processor reference</a> to
-				<var>processor</var>.
-
-			1. Return <var>processor</var>.
+				<var>this</var>.
 		</div>
 
 		<pre class="argumentdef" for="AudioWorkletProcessor/AudioWorkletProcessor()">
--- a/index.bs
+++ b/index.bs
@@ -9825,12 +9825,12 @@ Methods</h5>
 </dl>
 
 <h5 id="AudioWorkletProcessor-instantiation">
-The instantiation of AudioWorkletProcessor</h5>
+The instantiation of {{AudioWorkletProcessor}}</h5>
 
 At the end of the {{AudioWorkletNode}} construction,
 A [=struct=] named
 <dfn>processor construction data</dfn>
-will be prepared for the cross-thread transfer. This
+will be prepared for cross-thread transfer. This
 <a spec="infra" lt="struct">struct</a> contains the following
 <a spec="infra" for="struct" lt="item">items</a>:
 
@@ -9854,15 +9854,13 @@ the <a>rendering thread</a> will invoke the algorithm below:
 			[=processor construction data=] transferred from the
 			[=control thread=].
 
-		1. Let <var>processorName</var> be <var>constructionData</var>'s
-			[=processor construction data/name=].
+		1. Let <var>processorName</var>, <var>nodeReference</var> and
+			<var>serializedPort</var> be
+			<var>constructionData</var>'s
+			[=processor construction data/name=],
+			[=processor construction data/node=], and
+			[=processor construction data/port=] respectively.
 
-		1. Let <var>nodeReference</var> be <var>constructionData</var>'s
-			[=processor construction data/node=].
-
-		1. Let <var>serializedPort</var> be <var>constructionData</var>'s
-			[=processor construction data/port=].
-		
 		1. Let <var>serializedOptions</var> be
 			<var>constructionData</var>'s
 			[=processor construction data/options=].
@@ -9950,7 +9948,7 @@ Constructors</h5>
 			options: Optional initial parameters value for this {{AudioWorkletNode}}.
 		</pre>
 
-		When the constructor was called, the user agent MUST perform the
+		When the constructor is called, the user agent MUST perform the
 		following steps on the control thread:
 
 		<div algorithm="AudioWorkletNode()">
--- a/index.bs
+++ b/index.bs
@@ -9824,6 +9824,45 @@ Methods</h5>
 		</div>
 </dl>
 
+<h5 id="AudioWorkletProcessor-instantiation">
+Instantiation of AudioWorkletProcessor</h5>
+
+The construction of an {{AudioWorkletNode}} is followed by the construction
+of a corresponding {{AudioWorkletProcessor}} instance in the
+{{AudioWorkletGlobalScope}}.
+
+At the end of the {{AudioWorkletNode}} construction,
+<dfn>processor construction data</dfn> will be prepared for the cross-thread
+transfer. This data contains:
+
+<dl dfn-type=attribute dfn-for="processor-construction-data">
+	:<dfn>nodeName</dfn>
+	::
+		A {{DOMString}} of the {{AudioWorkletProcessr}} name that is to
+		be looked up <a>node name to processor constructor map</a>.
+	:<dfn>nodePort</dfn>
+	::
+		A serialized {{MessagePort}} that is paired with
+		the associated {{AudioWorkletNode}}'s {{AudioWorkletNode/port}}.
+	:<dfn>nodeOptions</dfn>
+	::
+		A serialzied {{AudioWorkletNodeOptions}} dictionary that is
+		given to the {{AudioWorkletNode}}
+		{{AudioWorkletNode()|constructor}}.
+</dl>
+
+This is transferred to
+the {{AudioWorkletGlobalScope}} to invoke {{AudioWorkletProcessor()|constructor}}.
+
+	<div algorithm="AudioWorkletProcessor-instantiation">
+	If the instantiation is invoked without a valid processor construction data
+	abort the following step, and fire error event on no.onprocessorerror
+
+		1. processor construction data -> nodeName
+		1. find a constructor from the AWGS
+		1. invoke constructor with processor construction data (port, options)
+	</div>
+
 <h4 interface lt="AudioWorkletNode">
 The {{AudioWorkletNode}} Interface</h4>
 
@@ -9887,14 +9926,6 @@ Constructors</h5>
 			name: A string that is a key for the {{BaseAudioContext}}’s <a>node name to parameter descriptor map</a>.
 			options: Optional initial parameters value for this {{AudioWorkletNode}}.
 		</pre>
-
-		Once the construction of the node is completed,
-		<dfn dfn>processor construction data</dfn> will be created and
-		transferred to the matching {{AudioWorkletProcessor}}'s
-		{{AudioWorkletProcessor()|constructor}}. This data contains
-		a set of objects and references that is required for
-		{{AudioWorkletProcessor}}'s construction. See the step 12 in
-		the algorithm below.
 
 		When the constructor was called, the user agent MUST perform the
 		following steps on the control thread:
@@ -10006,12 +10037,6 @@ Constructors</h5>
 
 			1. Return <var>node</var>.
 		</div>
-
-		During the construction of an {{AudioWorkletNode}}, a
-		corresponding {{AudioWorkletProcessor}} instance is also
-		automatically created in the {{AudioWorkletGlobalScope}}.
-		Note that the instantiation of these object pairs spans the
-		control thread and the rendering thread.
 </dl>
 
 <h5 id="AudioWorkletNode-attributes">
--- a/index.bs
+++ b/index.bs
@@ -9880,9 +9880,10 @@ the <a>rendering thread</a> will invoke the algorithm below:
 			{{AudioWorkletGlobalScope}}'s
 			<a>node name to processor constructor map</a>.
 		
-		1. Make <var>nodeReference</var>, <var>deserializedPort</var>,
-			<var>deserializedOptions</var> available for the
-			execution of <var>processorCtor</var>.
+		1. Make <dfn><var>nodeReference</var></dfn>,
+			<dfn><var>deserializedPort</var></dfn>, and
+			<var>deserializedOptions</var> available
+			for the execution of <var>processorCtor</var>.
 
 		1. Invoke the <var>processorCtor</var> with the argument of
 			<var>deserializedOptions</var>.
@@ -10248,17 +10249,17 @@ Constructors</h5>
 		the following steps are performed on the <a>rendering thread</a>.
 
 		<div algorithm="AudioWorkletProcessor()">
-			1. Find <var>nodeReference</var> that is made available
-				for this constructor. Throw a {{TypeError}}
-				exception if not found.
+			1. Find <var>[=nodeReference=]</var> that is made
+				available for this constructor. Throw a
+				{{TypeError}} exception if not found.
 			
-			If any of following steps throws any exception,
-			abort the rest of steps and <a>queue a task</a> to the
-			<a>control thread</a> to
-			<a spec="dom" lt="fire an event">fire</a> an
-			<a href="https://www.w3.org/TR/html50/webappapis.html#the-errorevent-interface">ErrorEvent</a>
-			named <code>processorerror</code> on
-			<var>nodeReference</var>.
+			1. If any of following steps throws any exception,
+				abort the rest of steps and <a>queue a task</a>
+				to the <a>control thread</a> to
+				<a spec="dom" lt="fire an event">fire</a> an
+				<a href="https://www.w3.org/TR/html50/webappapis.html#the-errorevent-interface">ErrorEvent</a>
+				named <code>processorerror</code> on
+				<var>nodeReference</var>.
 
 			1. Let <var>processor</var> be
 				<a spec="webidl" lt="this">this</a> value.
@@ -10269,7 +10270,7 @@ Constructors</h5>
 			1. Set <var>processor</var>'s {{[[callable process]]}}
 				to `true`.
 
-			1. Find <var>deserializedPort</var> that is made
+			1. Find <var>[=deserializedPort=]</var> that is made
 				available for this constructor. Throw
 				{{TypeError}} exception if not found.
 
--- a/index.bs
+++ b/index.bs
@@ -10237,6 +10237,9 @@ Constructors</h5>
 				invocation of the constructor throw a
 				{{TypeError}}.
 
+			1. Let <var>processor</var> be a new
+				{{AudioWorkletProcessor}} instance.
+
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -9824,6 +9824,63 @@ Methods</h5>
 		</div>
 </dl>
 
+<h5 id="AudioWorkletProcessor-instantiation">
+The instantiation of AudioWorkletProcessor</h5>
+
+The construction of an {{AudioWorkletNode}} is followed by the construction
```
bzbarsky: Is this text normative or informative?  Seems like it's informative, since the normative bit is just the last step in the AudioWorkletNode constructor?  If so, it would be best if this section were explicitly marked informative.

index.bs
```diff
@@ -9824,6 +9824,63 @@ Methods</h5>
...
+of a corresponding {{AudioWorkletProcessor}} instance in the
+{{AudioWorkletGlobalScope}}.
+
+At the end of the {{AudioWorkletNode}} construction,
+<dfn>processor construction data</dfn> will be prepared for the cross-thread
```
bzbarsky: This definition of "processor construction data" is normative, though.  Maybe this section should in fact be normative, remove the informative bits, and do the following:

1) Define what a "processor construction data" is.  
2) Define an algorithm like "create an AudioWorkletProcessor from processor construction data" which defines the steps to do that.

Then the step at the end of the AudioWorkletNode constructor will do something like "queue a control message to create an AudioWorkletProcessor from processor construction data consisting of nodeName as the nodeName, serializedProcessorPort as the nodePort, and serializedOptions as the nodeOptions" or so.  Might be worth looking in the infra spec to see how best to define the construction data and describe its creation.

index.bs
```diff
@@ -9824,6 +9824,63 @@ Methods</h5>
...
+transfer. This data contains:
+
+<dl>
+	: <dfn dfn>nodeName</dfn>
+	::
+		A {{DOMString}} of the {{AudioWorkletProcessor}} name that is to
```
bzbarsky: s/of the/which is the/

index.bs
```diff
@@ -9824,6 +9824,63 @@ Methods</h5>
...
+		be looked up <a>node name to processor constructor map</a>.
```
bzbarsky: "Looked up in the"

index.bs
```diff
@@ -9824,6 +9824,63 @@ Methods</h5>
...
+
+	<div id="invoking-processor-constructor" algorithm="invoking processor constructor">
+		Upon the arrival of <a>processor construction data</a>
+		<var>constructionData</var>, perform the following steps. If
+		any step throws an exception, abort the remaining steps and
+		<a>queue a task</a> to fire an
+		<a href="https://www.w3.org/TR/html50/webappapis.html#the-errorevent-interface">
+		ErrorEvent</a> named <code>processorerror</code> at the
+		{{AudioWorkletNode}} on the <a>control thread</a> with the
+		relevant information about the error.
```
bzbarsky: This should probably be more clear about how the ErrorEvent is initialized....

index.bs
```diff
@@ -9824,6 +9824,63 @@ Methods</h5>
...
+
+		1. Let <var>port</var> be the [=nodePort=] member of
+			<var>constructionData</var>.
+
+		1. Let <var>processorCtor</var> be the result of looking
+			up <var>name</var> on the {{AudioWorkletGlobalScope}}'s
+			<a>node name to processor constructor map</a>.
+
+		1. Perform Construct(<var>processorCtor</var>,
+			« <var>options</var> ») with the given <var>port</var>.
```
bzbarsky: What does "with the given port" mean, exactly?

index.bs
```diff
@@ -10183,44 +10226,25 @@ Constructors</h5>
...
 			the error.
 
-			1. If there is no [=processor construction data=]
-				passed from the associated {{AudioWorkletNode}},
-				throw a {{TypeError}}.
-
-			1. Let <var>processorPort</var> be
-				[$StructuredDeserializeWithTransfer$](<var>serializedProcessorPort</var>,
-				the current Realm).
+			1. If there is no provided [=nodePort=] at the
```
bzbarsky: This doesn't seem to be well-defined, unfortunately.  In particular, what does "invocation of the constructor" mean?  Keep in mind that we can have a subclass of `AudioWorkletProcessor` that invokes multiple `AudioWorkletProcessor` constructors, via `super()` or not.  And none of those invocations can provide a nodePort, note.

This part really needs to have a clear processing model spelled out.  Where is the port information coming from, and does the act of looking it up remove it from that place or not?

index.bs
```diff
@@ -10183,44 +10226,25 @@ Constructors</h5>
...
+				invocation of the constructor throw a
+				{{TypeError}}.
 
-			1. Let <var>options</var> be
-				[$StructuredDeserialize$](<var>serializedOptions</var>,
-				the current Realm).
+			1. Let <var>processor</var> be a new
+				{{AudioWorkletProcessor}} instance.
```
bzbarsky: You don't want to do this.  This part is handled for you by IDL already; see https://heycam.github.io/webidl/#create-an-interface-object step 1.8 and 1.9.  You can refer to that object here as "this".

index.bs
```diff
@@ -10229,6 +10253,8 @@ Constructors</h5>
 
 			1. Set <var>node</var>'s <a>processor reference</a> to
 				<var>processor</var>.
+
+			1. Return <var>processor</var>.
```
bzbarsky: This step is not needed or doable, since the caller of this algorithm (the IDL steps for constructors) already has the object and isn't expecting these steps to return anything.

index.bs
```diff
@@ -9824,6 +9824,63 @@ Methods</h5>
...
+Upon the arrival of the transferred data on the {{AudioWorkletGlobalScope}},
+the <a>rendering thread</a> will trigger the algorithm below:
+
+	<div id="invoking-processor-constructor" algorithm="invoking processor constructor">
+		Upon the arrival of <a>processor construction data</a>
+		<var>constructionData</var>, perform the following steps. If
+		any step throws an exception, abort the remaining steps and
+		<a>queue a task</a> to fire an
+		<a href="https://www.w3.org/TR/html50/webappapis.html#the-errorevent-interface">
+		ErrorEvent</a> named <code>processorerror</code> at the
```
rtoy: "processorerror" -> "processorError"?

hoch: This is an error event, whose name is usually all lowercase. See https://developer.mozilla.org/en-US/docs/Web/Events.

index.bs
hoch: This is something I wanted to ask, but you already pointed out the infra spec uses the similar pattern of using `construction data`. I'll take a look at it.

index.bs
hoch: I think both points are addressed by this PR. I just need to be more explicit about an AWP being created from the construction data. I'll address this over next few commits.

index.bs
hoch: Do you think the "relevant information" bit is too loose?

index.bs
hoch: The port is a part of the processor construction data. It is not an explicit argument from the constructor.

> Keep in mind that we can have a subclass of AudioWorkletProcessor that invokes multiple AudioWorkletProcessor constructors, via super() or not. And none of those invocations can provide a nodePort, note.

I think that's the intention? The AWP constructor cannot be invoked by the user code. So if there is no provided `port` object, the construction can't succeed. I think this is also closely related to how the processor construction data is passed and used. I'll take a look at the infra spec.

index.bs
hoch: Can you point where I can find such example? I looked https://infra.spec.whatwg.org/, but couldn't find examples that uses the similar pattern.

bzbarsky: The infra spec doesn't have existing "construction data" things.  But it does define https://infra.spec.whatwg.org/#structs which is basically what you are using here.  So you would say that the "construction data" is a [struct](https://infra.spec.whatwg.org/#structs) and the items in it have the names you want, etc.  This isn't strictly required, but it makes it clear that this "construction data" is a notational convenience spec data structure.

index.bs
bzbarsky: Well, what I meant is that it's really not clear what the message, filename, lineno, and colno attributes should _be_ in this case.   If the intent is that they are just `""`, `""`, `0`, and `0`, then we can just say "fire an event named `processorerror` using `ErrorEvent` at the AudioWorkletNode`, per <https://dom.spec.whatwg.org/#firing-events>.  If we want to initialize some of those attributes to specific values, we can do that similarly to the click event example given in that section.

index.bs
bzbarsky: No, the infra spec has nothing to say about this....

index.bs
bzbarsky: Your intention is that `super()` (which is an invocation of the AWP constructor in user code!) should be considered as having a port provided based on something about how the original subclass constructor was called, right?  I just think this part should be spelled out so it can be implemented interoperably.

hoch: > I just think this part should be spelled out so it can be implemented interoperably.

I agree but can we deal with that in a follow-up PR?

index.bs
hoch: Forming a struct sounds reasonable, but I am not sure if that'll solve any real problem because such struct will not be consumed by any exposed interface.

index.bs
hoch: Thanks for the pointer. I'll follow your suggestion to clarify this.

index.bs

index.bs
bzbarsky: We can if you would like, but dealing with it was the primary purpose of _this_ PR....  I feel like we're just going around in circles without getting closer to this key part actually being specified, and I am unfortunately running out of the time I have allocated to this.

bzbarsky: Though I guess if the goal here is rather to narrowly fix https://github.com/WebAudio/web-audio-api/issues/2021 then fine.  But we do still need to fix the port issue...

index.bs
bzbarsky: Infra structs are _never_ consumed by anything exposed.  They are just agreed-on ways of bundling several items together into a single package in a spec, so people reading the spec know what the terminology being used means.

index.bs
hoch: Yes. I think it's better to create a new PR that specifically addresses: the AudioWorkletProcessor's construction with the processor construction data.

Thanks for your time and effort on this. It's been a great help.

index.bs
hoch: Sure. As I mentioned in the comment below, the processor construction data issue will be addressed in a different PR.

index.bs
```diff
@@ -9824,6 +9824,71 @@ Methods</h5>
...
+			the current Realm).
+		
+		1. Let <var>processorCtor</var> be the result of looking
+			up  <var>processorName</var> on the
+			{{AudioWorkletGlobalScope}}'s
+			<a>node name to processor constructor map</a>.
+		
+		1. Make <dfn><var>nodeReference</var></dfn>,
+			<dfn><var>deserializedPort</var></dfn>, and
+			<var>deserializedOptions</var> available
```
hoch: This is a new concept of "making things available" for the AWP constructor.

index.bs
```diff
@@ -10200,53 +10249,34 @@ Constructors</h5>
...
-				up <var>nodeName</var> on the
-				{{AudioWorkletGlobalScope}}'s
-				<a>node name to processor constructor map</a>.
-
-			1. Let <var>processor</var> be the result of
-				Construct(<var>processorCtor</var>, « <var>options</var> »).
-
-			1. Set <var>processor</var>’s port to <var>processorPort</var>.
-
+			1. Find <var>[=nodeReference=]</var> that is made
```
hoch: This is how the AWP constructor retrieves the data required for the construction.

index.bs
```diff
@@ -10068,7 +10071,8 @@ Attributes</h5>
 		When an unhandled exception is thrown from the processor's
 		<code>constructor</code>, <code>process</code> method,
 		or any user-defined class method, the processor will
-		<a>queue a task</a> to fire an
+		<a>queue a task</a> to
+		<a spec="dom" lt="fire an event">fire</a> an
```
bzbarsky: I was pretty serious about the DOM spec's wording here: it's "fire an event named whatever using ErrorEvent" to fire an instance of ErrorEvent.

What _shoould_ the event fired here be named?  "error" or "processorerror"?

index.bs
```diff
@@ -10068,7 +10071,8 @@ Attributes</h5>
...
 		<a href="https://html.spec.whatwg.org/multipage/webappapis.html#the-errorevent-interface">
 		ErrorEvent</a> to the associated {{AudioWorkletNode}}.
```
bzbarsky: s/to/at/

index.bs
```diff
@@ -10245,34 +10249,34 @@ Constructors</h5>
...
-
-			1. Set <var>processor</var>’s {{AudioWorkletProcessor/port}}
-				to <var>processorPort</var>.
-
+			1. Find <var>[=nodeReference=]</var> that is made
+				available for this constructor. Throw a
+				{{TypeError}} exception if not found.
+			
+			1. If any of following steps throws any exception,
+				abort the rest of steps and <a>queue a task</a>
```
bzbarsky: "of the steps"

index.bs
```diff
@@ -10245,34 +10249,34 @@ Constructors</h5>
...
+				to the <a>control thread</a> to
+				<a spec="dom" lt="fire an event">fire</a> an
```
bzbarsky: Again, "fire an event named processorerror using ErrorEvent"

index.bs
```diff
@@ -10245,34 +10249,34 @@ Constructors</h5>
...
+				<a href="https://www.w3.org/TR/html50/webappapis.html#the-errorevent-interface">ErrorEvent</a>
+				named <code>processorerror</code> on
```
bzbarsky: "at nodeReference"

index.bs
```diff
@@ -10245,34 +10249,34 @@ Constructors</h5>
...
+				<var>nodeReference</var>.
+
+			1. Let <var>processor</var> be
+				<a spec="webidl" lt="this">this</a> value.
```
bzbarsky: "be the this value"

index.bs
```diff
@@ -9827,58 +9827,66 @@ Methods</h5>
...
-
-		1. Perform Construct(<var>processorCtor</var>,
-			« <var>options</var> ») with the given <var>port</var>.
+		
+		1. Make <dfn><var>nodeReference</var></dfn>,
+			<dfn><var>deserializedPort</var></dfn>, and
+			<var>deserializedOptions</var> available
+			for the execution of <var>processorCtor</var>.
+
+		1. Invoke the <var>processorCtor</var> with the argument of
```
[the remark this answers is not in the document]
hoch: Perhaps it's better to say `Perform Construct(processorCtor, deserializedOptions)` like the step 9 of https://heycam.github.io/webidl/#construct-a-callback-function.

I am not sure `deserializedOptions` is a valid ES value. Probably it needs a conversion?

index.bs
hoch: Understood. It's `processorerror`.

index.bs
hoch: Sure. I would like to merge the progress in this PR, and follow-up on this part later.

index.bs
```diff
@@ -9824,6 +9824,71 @@ Methods</h5>
...
+		1. Let <var>serializedPort</var> be <var>constructionData</var>'s
+			[=processor construction data/port=].
+		
+		1. Let <var>serializedOptions</var> be
+			<var>constructionData</var>'s
+			[=processor construction data/options=].
+		
+		1. Let <var>deserializedPort</var> be the result of
+			[$StructuredDeserialize$](<var>serializedPort</var>,
+			the current Realm).
```
bzbarsky: What current Realm?  There is no current Realm here.

index.bs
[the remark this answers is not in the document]
bzbarsky: Why would we prefer `Construct` over "construct a callback function", especially since what we have is in fact a callback function?  I mean, we _can_ do a manual `Construct` but then we will need to also manually duplicate a lot of the bookkeeping "construct a callback function" does for us, no?

`deserializedOptions` is the return value of a `StructuredDeserialize` call, right?  That would be an ES value, as defined at https://html.spec.whatwg.org/multipage/structured-data.html#structureddeserialize

index.bs
hoch: Okay, then how do we deserialize the transferred data without a `realm`? I took the text for granted when I got some help from @domenic. Perhaps the `realm` should be given to the algorithm from the call site?

index.bs
```diff
@@ -9824,6 +9824,71 @@ Methods</h5>
 		</div>
 </dl>
 
+<h5 id="AudioWorkletProcessor-instantiation">
+The instantiation of AudioWorkletProcessor</h5>
```
rtoy: Maybe `of {{AudioWorkletProcessor}}`? Not sure about that though.

index.bs
```diff
@@ -9824,6 +9824,71 @@ Methods</h5>
...
+
+At the end of the {{AudioWorkletNode}} construction,
+A [=struct=] named
+<dfn>processor construction data</dfn>
+will be prepared for the cross-thread transfer. This
```
rtoy: "for the cross-thread" -> "for cross-thread"

index.bs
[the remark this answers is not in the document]
hoch: I don't have any preference here. I am merely focusing two things: your suggestion and minimizing changes.

> I mean, we can do a manual Construct but then we will need to also manually duplicate a lot of the bookkeeping "construct a callback function" does for us, no?

It sounds like you're constructing a callback function, but not executing the constructor itself. I was mistaken. If this is the correct way of performing the constructor (processorCtor), then I am all for it.

Also the algorithm has some interesting bits:
"Let relevant settings be realm’s settings object."

Is this "setting object" similar to "processor construction data"? Then this algorithm might be a solution for our situation.

> deserializedOptions is the return value of a StructuredDeserialize call, right? That would be an ES value

Understood.

index.bs
```diff
@@ -9824,6 +9824,71 @@ Methods</h5>
...
+		1. Let <var>processorName</var> be <var>constructionData</var>'s
+			[=processor construction data/name=].
+
+		1. Let <var>nodeReference</var> be <var>constructionData</var>'s
+			[=processor construction data/node=].
+
+		1. Let <var>serializedPort</var> be <var>constructionData</var>'s
+			[=processor construction data/port=].
+		
+		1. Let <var>serializedOptions</var> be
```
rtoy: Can steps 2-5 be placed all in one step? The order isn't really important here, right?  But ok if you want to leave it as is.

index.bs
```diff
@@ -9888,14 +9949,6 @@ Constructors</h5>
...
 
-		Once the construction of the node is completed,
-		<dfn dfn>processor construction data</dfn> will be created and
-		transferred to the matching {{AudioWorkletProcessor}}'s
-		{{AudioWorkletProcessor()|constructor}}. This data contains
-		a set of objects and references that is required for
-		{{AudioWorkletProcessor}}'s construction. See the step 12 in
-		the algorithm below.
-
 		When the constructor was called, the user agent MUST perform the
```
rtoy: Nit: "was called" -> "is called".  But since this isn't part of this PR, you can ignore this.

index.bs
```diff
@@ -10200,53 +10249,34 @@ Constructors</h5>
...
+			1. If any of following steps throws any exception,
+				abort the rest of steps and <a>queue a task</a>
+				to the <a>control thread</a> to
+				<a spec="dom" lt="fire an event">fire</a> an
+				<a href="https://www.w3.org/TR/html50/webappapis.html#the-errorevent-interface">ErrorEvent</a>
+				named <code>processorerror</code> on
+				<var>nodeReference</var>.
+
+			1. Let <var>processor</var> be
+				<a spec="webidl" lt="this">this</a> value.
```
rtoy: Doesn't `{{this}}` work?  (You don't have to check.)

hoch: Nope! (just checked)

index.bs
hoch: Also this algorithm is performed in AudioWorkletGlobalScope so there is a defined "current realm", I think.
- https://tc39.es/ecma262/#current-realm
- https://www.w3.org/TR/worklets-1/#creating-a-workletglobalscope

bzbarsky: You want the realm of the AudioWorkletGlobalScope, but it's not the current realm at that point.  In order for it to be the current realm, you'd have to have a running execution context, which can only happen if something pushes it on the execution context stack, etc.  None of that is happening here.

What you probably want to say is that deserialization is passed the realm of the AudioWorkletGlobalScope involved.  It doesn't look like ES really defines getting the realm of a global; I guess you can say you use the realm whose  [[GlobalObject]]  is the AudioWorkletGlobalScope that corresponds to the BaseAudioContext of the AudioWorkletNode that we're in the middle of creating....

bzbarsky: I guess maybe https://html.spec.whatwg.org/multipage/webappapis.html#concept-global-object-realm offers a shortcut, so you can just talk about taking the realm of the AudioWorkletGlobalScope etc.

index.bs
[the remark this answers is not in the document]
bzbarsky: > If this is the correct way of performing the constructor

Yes, it is.

> Is this "setting object" similar to "processor construction data"?

No, it's the thing defined at https://html.spec.whatwg.org/multipage/webappapis.html#settings-object

index.bs
hoch: Thanks for the pointer, but this is really out of scope of this PR. I file a new issue and work on a new PR for it.

index.bs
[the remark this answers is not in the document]

hoch:
Merging after an approval. I opened #2027 for some issues raised in this PR.

PR #2023, opened 2019-08-13
Add an algorithm to validate an AudioWorkletNodeOptions

### LINKED ISSUE #1969: It would be clearer if the various validation of AudioWorkletNodeOptions were an explicit step or set of steps
https://webaudio.github.io/web-audio-api/#dictionary-audioworkletnodeoptions-members defines various validity constraints on `AudioWorkletNodeOptions` members that are supposed to lead to exceptions.

https://webaudio.github.io/web-audio-api/#configuring-channels-with-audioworkletnodeoptions defines some more validity constraints.

It's not 100% clear to me whether these constraints should be enforced in just the `AudioWorkletNode` constructor, or also in the `AudioWorkletProcessor` constructor (even though that claims to ignore its options arg), or something else.

It would be much clearer if there were a "validate a `AudioWorkletNodeOptions`" algorithm defined that does the relevant checks and if that algorithm were invoked in the relevant places explicitly.

hoch:
Fixes #1969.

- A new algorithm for configuring channels with AudioWorkletNodeOptions
- Change the AWN constructor algorithm to include a new validation check.

cc @bzbarsky

***
<a href="https://pr-preview.s3.amazonaws.com/hoch/web-audio-api/pull/2023.html" title="Last updated on Aug 14, 2019, 10:05 PM UTC (c9717b4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/2023/8387895...hoch:c9717b4.html" title="Last updated on Aug 14, 2019, 10:05 PM UTC (c9717b4)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -10137,10 +10137,11 @@ used to configure various channel configurations.
 		1. If both {{AudioWorkletNodeOptions/numberOfInputs}} and
 			{{AudioWorkletNodeOptions/numberOfOutputs}} are 1,
 			set the channel count of the <var>node</var> output to
-			the value and return.
+			the one value in
+			{{AudioWorkletNodeOptions/outputChannelCount}}.
 
-		1. Othewise set the channel count of each output of the
-			<var>node</var> to each value in the
+		1. Otherwise set the channel count of each output of the
+			<var>node</var> to the corresponding value in the
 			{{AudioWorkletNodeOptions/outputChannelCount}} sequence
 			and return.
 
@@ -10154,10 +10155,10 @@ used to configure various channel configurations.
 
 			NOTE: For this case, the output chanel count will
 			change to <a>computedNumberOfChannels</a> dynamically
-			based on the input in runtime.
+			based on the input at runtime.
 
-		1. Otherwise set all the outputs of the <var>node</var> to 1
-			and return.
+		1. Otherwise set the channel counts of the outputs of the
+			<var>node</var> to 1 and return.
 </div>
 
 <h4 interface lt="AudioWorkletProcessor">
--- a/index.bs
+++ b/index.bs
@@ -10061,11 +10061,6 @@ The {{AudioWorkletNodeOptions}} dictionary can be used
 to initialize attibutes in the instance of an {{AudioWorkletNode}} and
 an {{AudioWorkletProcessor}}.
 
-The base implementation of {{AudioWorkletProcessor}} does not examine the values
-in this dictionary. However, an user-defined subclass of the
-{{AudioWorkletProcessor}} can use this dictionary to initialize custom
-properties in its instance.
-
 <xmp class="idl">
 dictionary AudioWorkletNodeOptions : AudioNodeOptions {
 	unsigned long numberOfInputs = 1;
@@ -10118,30 +10113,49 @@ Dictionary {{AudioWorkletNodeOptions}} Members</h6>
 		{{AudioWorkletNode}}.
 </dl>
 
-<h6 id="configuring-channels-with-audioworkletnodeoptions">
-Configuring Channels with {{AudioWorkletNodeOptions}}</h6>
+<h6>Configuring Channels with {{AudioWorkletNodeOptions}}</h6>
 
-With a combination of {{AudioWorkletNodeOptions/numberOfInputs}},
-{{AudioWorkletNodeOptions/numberOfOutputs}},
-and {{AudioWorkletNodeOptions/outputChannelCount}},
-various channel configurations can be achieved.
+The following algorithm describes how an {{AudioWorkletNodeOptions}} can be
+used to configure various channel configurations.
 
-: {{AudioWorkletNodeOptions/numberOfInputs}} = 0,
-	{{AudioWorkletNodeOptions/numberOfOutputs}} = 0
-:: {{NotSupportedError}} MUST be thrown by the constructor.
+<div id="configuring-channels-with-audioworkletnodeoptions" algorithm="configuring channels with AudioWorkletNodeOptions">
+	1. Let <var>node</var> be an {{AudioWorkletNode}} instance that is
+		given to this algorithm.
 
-: {{AudioWorkletNodeOptions/numberOfInputs}} = 1,
-	{{AudioWorkletNodeOptions/numberOfOutputs}} = 1
-:: If {{AudioWorkletNodeOptions/outputChannelCount}} is
-	unspecified, the output channel count will match
-	<a>computedNumberOfChannels</a> from the input. If
-	{{AudioWorkletNodeOptions/outputChannelCount}} is specified
-	with one value, the output channel count will match the given value.
+	1. If both {{AudioWorkletNodeOptions/numberOfInputs}} and
+		{{AudioWorkletNodeOptions/numberOfOutputs}} are zero,
+		throw a {{NotSupportedError}} and abort the remaining steps.
 
-: All other cases
-:: If {{AudioWorkletNodeOptions/outputChannelCount}} is
-	unspecified, it will be mono for all outputs. If specified,
-	the values will be used as given.
+	1. If the length of {{AudioWorkletNodeOptions/outputChannelCount}}
+		does not match {{AudioWorkletNodeOptions/numberOfOutputs}},
+		throw an {{IndexSizeError}} and abort the remaining steps.
+
+	1. If any value in {{AudioWorkletNodeOptions/outputChannelCount}}
+		is zero or greater than the implementation’s maximum number of
+		channels, throw a {{NotSupportedError}} and abort the remaining
+		steps.
+
+	1. If both {{AudioWorkletNodeOptions/numberOfInputs}} and
+		{{AudioWorkletNodeOptions/numberOfOutputs}} are 1,
+
+		1. If {{AudioWorkletNodeOptions/outputChannelCount}} is
+			given and it contains a value, the channel count of the
+			<var>node</var> output will be set to the value and
+			return.
+
+		1. If {{AudioWorkletNodeOptions/outputChannelCount}} is
+			unspecified, the output channel will match
+			<a>computedNumberOfChannels</a> based on the input and
+			return.
+
+	1. If {{AudioWorkletNodeOptions/outputChannelCount}} is unspecified,
+		all the outputs of the <var>node</var> will be set to mono and
+		return.
+
+	1. If {{AudioWorkletNodeOptions/outputChannelCount}} is specified,
+		each value in the sequence will be used to set the channel count
+		of each output of the <var>node</var>.
+</div>
 
 <h4 interface lt="AudioWorkletProcessor">
 The {{AudioWorkletProcessor}} Interface</h4>
--- a/index.bs
+++ b/index.bs
@@ -9903,10 +9903,6 @@ Constructors</h5>
 			When the {{AudioWorkletNode()|AudioWorkletNode}} constructor
 			is invoked with <var ignore>context</var>, <var>nodeName</var>, <var>options</var>:
 
-			1. Perform the validity check on <var>options</var>. If the
-				check throws any exception, propagate the exception
-				and abort these steps.
-
 			1. If <var>nodeName</var> does not exist as a key in the
 				{{BaseAudioContext}}’s <a>node name to parameter
 				descriptor map</a>, throw a {{InvalidStateError}}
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -9915,6 +9911,12 @@ Constructors</h5>
 			1. Let <var>node</var> be the instance being created by the
```
bzbarsky: Pre-existing, but you should just be able to reference `this` here, per https://heycam.github.io/webidl/#create-an-interface-object step 1.9

index.bs
```diff
@@ -10093,55 +10090,65 @@ Dictionary {{AudioWorkletNodeOptions}} Members</h6>
...
+
+<div id="configure-with-audioworkletnodeoptions" algorithm="configure with AudioWorkletNodeOptions">
+	1. Let <var>node</var> be an {{AudioWorkletNode}} instance that is
+		given to this algorithm.
+
+	1. If both {{AudioWorkletNodeOptions/numberOfInputs}} and
+		{{AudioWorkletNodeOptions/numberOfOutputs}} are zero,
+		throw a {{NotSupportedError}} and abort the remaining steps.
+
+	1. If the length of {{AudioWorkletNodeOptions/outputChannelCount}}
```
bzbarsky: This needs to handle `outputChannelCount` not being present, right?

index.bs
```diff
@@ -10093,55 +10090,65 @@ Dictionary {{AudioWorkletNodeOptions}} Members</h6>
...
+		does not match {{AudioWorkletNodeOptions/numberOfOutputs}},
+		throw an {{IndexSizeError}} and abort the remaining steps.
+
+	1. If any value in {{AudioWorkletNodeOptions/outputChannelCount}}
```
bzbarsky: Again, needs to handle `outputChannelCount` not being present.  By the way, should this sequence member be names `outputChannelCounts`?

index.bs
```diff
@@ -10093,55 +10090,65 @@ Dictionary {{AudioWorkletNodeOptions}} Members</h6>
...
+		is zero or greater than the implementation’s maximum number of
+		channels, throw a {{NotSupportedError}} and abort the remaining
+		steps.
+
+	1. If both {{AudioWorkletNodeOptions/numberOfInputs}} and
+		{{AudioWorkletNodeOptions/numberOfOutputs}} are 1,
+
+		1. If {{AudioWorkletNodeOptions/outputChannelCount}} is
+			given and it contains a value, the channel count of the
```
bzbarsky: The term IDL defines is "present".  See https://heycam.github.io/webidl/#dfn-present

bzbarsky: I don't understand what "it contains a value" means.  Since `numberOfOutputs` is 1 and `outputChannelCount` is present, we already threw if the length of `outputChannelCount` is not 1.... so how can it not "contain a value"?

index.bs
```diff
@@ -10093,55 +10090,65 @@ Dictionary {{AudioWorkletNodeOptions}} Members</h6>
...
+			<var>node</var> output will be set to the value and
+			return.
+
+		1. If {{AudioWorkletNodeOptions/outputChannelCount}} is
+			unspecified, the output channel will match
```
bzbarsky: Does "unspecified" mean "not present"?

index.bs
```diff
@@ -10093,55 +10090,65 @@ Dictionary {{AudioWorkletNodeOptions}} Members</h6>
...
+			<a>computedNumberOfChannels</a> based on the input and
+			return.
+
+	1. If {{AudioWorkletNodeOptions/outputChannelCount}} is unspecified,
```
bzbarsky: "not present"?

index.bs
```diff
@@ -10093,55 +10090,65 @@ Dictionary {{AudioWorkletNodeOptions}} Members</h6>
...
+		all the outputs of the <var>node</var> will be set to mono and
+		return.
+
+	1. If {{AudioWorkletNodeOptions/outputChannelCount}} is specified,
```
bzbarsky: "present"

index.bs
```diff
@@ -10093,55 +10090,65 @@ Dictionary {{AudioWorkletNodeOptions}} Members</h6>
...
+		is zero or greater than the implementation’s maximum number of
+		channels, throw a {{NotSupportedError}} and abort the remaining
+		steps.
+
+	1. If both {{AudioWorkletNodeOptions/numberOfInputs}} and
+		{{AudioWorkletNodeOptions/numberOfOutputs}} are 1,
+
+		1. If {{AudioWorkletNodeOptions/outputChannelCount}} is
+			given and it contains a value, the channel count of the
+			<var>node</var> output will be set to the value and
```
bzbarsky: The "will be set to ABC and return" phrasing on all of these is a little weird.  It would be clearer to say things like "set the channel count of the node output to ABC and return".

index.bs
hoch: Do you propose to replace `node` with `this`? or do you mean I can just use `this` and remove this step?

index.bs
hoch: Yes. I fixed it.

index.bs
bzbarsky: You can just say "Let _node_ be the **this** value" with a link to https://heycam.github.io/webidl/#this if desired.

index.bs
hoch: This is now addressed.

index.bs
hoch: I'll bring it up in the next teleconference.

index.bs
hoch: The new commit rearranged these step to handle this first.

index.bs
hoch: Good to know. I'll address this in a different PR.

index.bs
```diff
@@ -10093,55 +10090,75 @@ Dictionary {{AudioWorkletNodeOptions}} Members</h6>
...
+		1. If the length of
+			{{AudioWorkletNodeOptions/outputChannelCount}} does not
+			match {{AudioWorkletNodeOptions/numberOfOutputs}},
+			throw an {{IndexSizeError}} and abort the remaining
+			steps.
+
+		1. If both {{AudioWorkletNodeOptions/numberOfInputs}} and
+			{{AudioWorkletNodeOptions/numberOfOutputs}} are 1,
+			set the channel count of the <var>node</var> output to
+			the value and return.
```
bzbarsky: What is "the value" here?  Presumably it should be "the one value in {{AudioWorkletNodeOptions/outputChannelCount}}" or something?

index.bs
```diff
@@ -10093,55 +10090,75 @@ Dictionary {{AudioWorkletNodeOptions}} Members</h6>
...
+
+		1. Othewise set the channel count of each output of the
+			<var>node</var> to each value in the
```
bzbarsky: s/each/the corresponding/?

I assume the outputs have a well-defined ordering on them that's obvious?

index.bs
```diff
@@ -10093,55 +10090,75 @@ Dictionary {{AudioWorkletNodeOptions}} Members</h6>
...
+		<a spec="webidl" lt="present">not present</a>,
+
+		1. If both {{AudioWorkletNodeOptions/numberOfInputs}} and
+			{{AudioWorkletNodeOptions/numberOfOutputs}} are 1,
+			set the channel count of the <var>node</var> output to
+			1 and return.
+
+			NOTE: For this case, the output chanel count will
+			change to <a>computedNumberOfChannels</a> dynamically
+			based on the input in runtime.
```
bzbarsky: "at runtime"?

index.bs
```diff
@@ -10093,55 +10090,75 @@ Dictionary {{AudioWorkletNodeOptions}} Members</h6>
...
+
+		1. Otherwise set all the outputs of the <var>node</var> to 1
```
bzbarsky: Set the outputs, or set the channel counts of the outputs?

index.bs
hoch: Yes.

index.bs
```diff
@@ -10093,55 +10088,76 @@ Dictionary {{AudioWorkletNodeOptions}} Members</h6>
...
+		<a spec="webidl" lt="present">not present</a>,
+
+		1. If both {{AudioWorkletNodeOptions/numberOfInputs}} and
+			{{AudioWorkletNodeOptions/numberOfOutputs}} are 1,
+			set the channel count of the <var>node</var> output to
+			1 and return.
+
+			NOTE: For this case, the output chanel count will
+			change to <a>computedNumberOfChannels</a> dynamically
+			based on the input at runtime.
```
rtoy: Nit: based on the input and the channelCountMode.

index.bs
```diff
@@ -10093,55 +10088,76 @@ Dictionary {{AudioWorkletNodeOptions}} Members</h6>
...
+
+	1. If {{AudioWorkletNodeOptions/outputChannelCount}} is
+		<a spec="webidl" lt="present">not present</a>,
+
+		1. If both {{AudioWorkletNodeOptions/numberOfInputs}} and
+			{{AudioWorkletNodeOptions/numberOfOutputs}} are 1,
+			set the channel count of the <var>node</var> output to
+			1 and return.
+
+			NOTE: For this case, the output chanel count will
```
rtoy: Can't do this because it makes it highlighted as non-normative, but it is normative.

index.bs
```diff
@@ -10093,55 +10088,76 @@ Dictionary {{AudioWorkletNodeOptions}} Members</h6>
...
+			change to <a>computedNumberOfChannels</a> dynamically
+			based on the input at runtime.
+
+		1. Otherwise set the channel counts of the outputs of the
```
rtoy: Perhaps better phrased as "set the channel count of each output of the node to 1..."

index.bs
```diff
@@ -10093,55 +10088,76 @@ Dictionary {{AudioWorkletNodeOptions}} Members</h6>
...
+
+		1. If any value in
+			{{AudioWorkletNodeOptions/outputChannelCount}} is zero
+			or greater than the implementation’s maximum number
+			of channels, throw a {{NotSupportedError}} and abort
+			the remaining steps.
+
+		1. If the length of
+			{{AudioWorkletNodeOptions/outputChannelCount}} does not
+			match {{AudioWorkletNodeOptions/numberOfOutputs}},
```
rtoy: Nit: "does not match" -> "does not equal"

index.bs
```diff
@@ -10093,55 +10088,76 @@ Dictionary {{AudioWorkletNodeOptions}} Members</h6>
...
+			throw an {{IndexSizeError}} and abort the remaining
+			steps.
+
+		1. If both {{AudioWorkletNodeOptions/numberOfInputs}} and
+			{{AudioWorkletNodeOptions/numberOfOutputs}} are 1,
+			set the channel count of the <var>node</var> output to
+			the one value in
+			{{AudioWorkletNodeOptions/outputChannelCount}}.
+
+		1. Otherwise set the channel count of each output of the
```
rtoy: Maybe it's clearer if you say something like:

set the channel count of the k'th output of the node to the k'th element of outputChannelCount.

index.bs
hoch: I rearranged the position of the note.

index.bs
```diff
@@ -10151,14 +10151,14 @@ used to configure various channel configurations.
...
-			NOTE: For this case, the output chanel count will
-			change to <a>computedNumberOfChannels</a> dynamically
-			based on the input at runtime.
-
-		1. Otherwise set the channel counts of the outputs of the
+		1. Otherwise set the channel count of each output of the
 			<var>node</var> to 1 and return.
 </div>
 
+NOTE: For the case of the step 4.1 above, the output chanel count will
```
rtoy: This doesn't really help because this is actually normative.  The node MUST do this.  If you can't figure out a bikeshed way to make a note that isn't non-normative, just leave out "NOTE:" (and put it back where it was).

hoch: No. I think this is informative note.

This algorithm describes what should happen when we validate the option. What happens at runtime is out of scope. That's why I used the "note" notation.

This has been a big problem in the spec. When a certain behavior is out of the algorithm scope, it should not be a step in the algorithm. Or make it clear that it is out of its scope.

Also having a note within an algorithm is not an illegal thing to do:
https://html.spec.whatwg.org/multipage/workers.html#worker-processing-model

If this were up to me, I would just put the note back into where it was.

rtoy: My point is that the way the algorithm is currently phrased the count is 1 and nothing else.  The note is that it can be changed from 1.  And if you leave the note out I take it to mean exactly one.  With the comment, then I see that it's not fixed to one.

But if you rephrase to say "initial count" is one, then you can leave the note in (or just remove it) because then I don't infer it to mean a fixed value of 1.

hoch: That's a good idea. I'll use "initial count" and keep the note as is.

bzbarsky: Sounds good to me, fwiw.  I assume there is an explicit algorithm step somewhere else that changes the count to be not-1, right?

hoch: Yes. The algorithm `computedNumberOfChannels` and mixing rules are rather well-defined.

rtoy: Sorry for not being clearer in my original comment!

PR #2034, opened 2019-08-20
Handles the return value of AWP.process() in the rendering loop algorithm

### LINKED ISSUE #1970: "active source" flag handling is a weird monkeypatch
https://webaudio.github.io/web-audio-api/#dom-audioworkletprocessor-process says:

> At the conclusion of each call to the process() method, ToBoolean (described in [ECMASCRIPT]) is applied to the return value and the result is assigned to the associated AudioWorkletProcessor’s active source flag.

This is a pretty weird place to put that handling, imo, assuming this is meant to be a normative requirement (which I assume it is, since nothing else really sets the flag).  It would be a lot clearer if it happened at https://webaudio.github.io/web-audio-api/#rendering-a-graph step 3.4.4.2.3 or so, where we actually make the call.  Otherwise it's pretty easy to miss this requirement while implementing the processing model steps.

Having an informative note about what the function's return value means does make some sense.

hoch:
Fixes #1970.

***
<a href="https://pr-preview.s3.amazonaws.com/hoch/web-audio-api/pull/2034.html" title="Last updated on Aug 23, 2019, 5:31 PM UTC (ed3f019)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/2034/1aa738f...hoch:ed3f019.html" title="Last updated on Aug 23, 2019, 5:31 PM UTC (ed3f019)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -10317,16 +10317,13 @@ of {{AudioParamDescriptor}}s.
 		every <a>render quantum</a>, if {{AudioWorkletNode}} is
 		[=actively processing=].
 
-		The return value of this method controls the lifetime of the
-		{{AudioWorkletProcessor}}'s associated
-		{{AudioWorkletNode}}. At the conclusion of each call to the
-		{{process()}} method, <a href="https://tc39.github.io/ecma262/#sec-toboolean"><code>ToBoolean</code></a> (described in [[!ECMASCRIPT]])
-		is applied to the return value and the result is assigned to the associated AudioWorkletProcessor's <a>active source</a> flag. This in turn affects whether subsequent
-		invocations of {{process()}} occur, and has an impact on the lifetime of the node.
-
 		<div class="note">
-			This lifetime policy can support a variety of approaches found in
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -10317,16 +10317,13 @@ of {{AudioParamDescriptor}}s.
...
-
 		<div class="note">
-			This lifetime policy can support a variety of approaches found in
-			built-in nodes, including the following:
+			The return value of this method controls the lifetime
+			of the {{AudioWorkletProcessor}}'s associated
+			{{AudioWorkletNode}}.
+
+			This lifetime policy can support a variety of approaches
+			found in built-in nodes, including the following:
```
rtoy: If you do this, then where is the normative description of what the return value means?

hoch: Then perhaps we want to put l.10321-10323 back to where it was?

rtoy: Oops. I missed the other part.  lgtm.

PR #2040, opened 2019-08-21
Remove `options` argument from AWP constructor

### LINKED ISSUE #2036: Remove `options` parameter of `AudioWorkletProcessor` constructor
https://github.com/WebAudio/web-audio-api/issues/1965#issuecomment-522750816 says

> > step 2 says that the AudioWorkletProcessor is initialized using options. At the moment this is a no-op, as the following note says. That's fine, but in that case why have this step listed at all?
> 
> This is not relevant anymore.
> 
> > though nothing actually forces them to, since the argument is optional, right? And at the moment the superclass (and indeed UA code in general) does nothing with the object, as far as I can tell.
> 
> This is also changed. Now the construction relies on "processor construction data", so the user code does not need to use "super(options)" to initialize the internals correctly.

hoch:
Fixes #2036 per https://github.com/WebAudio/web-audio-api/issues/2036#issuecomment-523601202.

In short, the base class constructor does not need `options` passed from the node.

***
<a href="https://pr-preview.s3.amazonaws.com/hoch/web-audio-api/pull/2040.html" title="Last updated on Aug 21, 2019, 7:30 PM UTC (8534ca8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/2040/a73dab6...hoch:8534ca8.html" title="Last updated on Aug 21, 2019, 7:30 PM UTC (8534ca8)">Diff</a>

### DIFF
--- a/expected-errs.txt
+++ b/expected-errs.txt
@@ -47,6 +47,5 @@ LINK ERROR: No 'idl-name' refs found for 'sequence<unsigned long>'.
 <a data-link-type="idl-name" data-lt="sequence&lt;unsigned long&gt;">sequence&lt;unsigned long&gt;</a>
 LINK ERROR: No 'idl-name' refs found for 'record<DOMString, double>'.
 <a data-link-type="idl-name" data-lt="record&lt;DOMString, double&gt;">record&lt;DOMString, double&gt;</a>
-LINE 9993: No 'method' refs found for 'AudioWorkletProcessor()'.
 LINE 10077: No 'method' refs found for 'process(inputs, outputs, parameters))'.
 YAY Successfully generated, but fatal errors were suppressed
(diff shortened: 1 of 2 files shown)


pm100:
well I just spent a happy few hours working out that I cannot pass options into the constructor (I only needed options at worklet startup so this seemed a better option that processing via process method args). All the docs say its supported. The typescript def files all have it.

padenot:
You can, and it works:

```html
<script type="[worklet]()">
  registerProcessor('test-param', class param extends AudioWorkletProcessor {
    constructor(options) {
      console.log("ctor", options.processorOptions);
      super(options)
    }
    process(input, output, parameters) {
    }
  });
</script>
<script>
  var ac = new AudioContext;
  var e = document.querySelector("script[type=worklet]")
  var text = e.innerText;
  const blob = new Blob([text], {type: "application/javascript"});

  var url = URL.createObjectURL(blob);
  var option = { key: "An option" };
  ac.audioWorklet.addModule(url).then(() => {
      nodea = new AudioWorkletNode(ac, 'test-param', {processorOptions: option});
      nodea.connect(ac.destination);
  });
</script>
```

PR #2092, opened 2019-10-28
Refactor the constructor and factory method text

### LINKED ISSUE #2073: Inconsistencies in constructor descriptions and factory method initialization
The Web Audio API specifies an algorithm [to create an AudioNode of a particular type using its constructor](https://webaudio.github.io/web-audio-api/#AudioNode-creation).

Many descriptions of constructors more or less repeat that algorithm, but without explicitly passing arguments to the algorithm for [Initializing an object that inherits from AudioNode](https://webaudio.github.io/web-audio-api/#audionode-constructor-init).
e.g. [the AudioBufferSourceNode constructor](https://webaudio.github.io/web-audio-api/#dom-audiobuffersourcenode-audiobuffersourcenode).

Some descriptions of constructors use a similar algorithm, but extended to perform additional initialization (e.g. [the MediaStreamAudioSourceNode ](https://webaudio.github.io/web-audio-api/#dom-audiobuffersourcenode-audiobuffersourcenode)).

The [AudioWorkletNode constructor description](https://webaudio.github.io/web-audio-api/#AudioWorkletNode-constructors) is a little different in that the algorithm initializes `this` instead of creating and returning a new object.

The approach for AudioWorkletNode seems more consistent with the WebIDL spec which says that
[The prose definition of a constructor operation must either initialize the value passed as `this`, or throw an exception.](https://heycam.github.io/webidl/#idl-constructors)
[The WebIDL algorithm for creation of an interface object with the ECMAScript binding](https://heycam.github.io/webidl/#create-an-interface-object) will "Perform the actions listed in the description of constructor with values as the argument values and object as this" and so it would not seem appropriate for the description of the constructor to return a new object.

Given that each particular type describes the steps for the constructor and WebIDL describes the steps for creation from ECMAScript and that the generic algorithm is not complete for all "particular type"s, I guess it would be best to remove the generic algorithm "to create an `AudioNode` of a particular type _n_ using its constructor".

Also, the algorithm "to create a new `AudioNode` of a particular type _n_ using its factory method" does not invoke the constructor, and so does not contain any type-specific additional initialization steps, such as those in the `MediaStreamAudioSourceNode` constructor.  Perhaps this should invoke the constructor instead of the "Intialize" algorithm.

padenot:
This fixes #2073, by doing what is spelled out at https://github.com/WebAudio/web-audio-api/issues/2073#issuecomment-540651254.

I caught a couple mistakes while doing this, in the first commit.

***
<a href="https://pr-preview.s3.amazonaws.com/padenot/web-audio-api/pull/2092.html" title="Last updated on Nov 12, 2019, 3:04 PM UTC (18b95bd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/2092/ef3262d...padenot:18b95bd.html" title="Last updated on Nov 12, 2019, 3:04 PM UTC (18b95bd)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2631,41 +2631,29 @@ is called on.
 
 <div algorithm="initialize an AudioNode">
 	<dfn id="audionode-constructor-init">Initializing</dfn> an object
-	<var>o</var> of interface <var>n</var> that inherits from
-	{{AudioNode}} means executing the following steps, given the
-	arguments <var>context</var> and <var>dict</var> passed to the
-	constructor of this interface.
+	<var>o</var> that inherits from {{AudioNode}} means executing the following
+	steps, given the arguments <var>context</var> and <var>dict</var> passed to
+	the constructor of this interface.
 
 	1. Set <var>o</var>'s associated {{BaseAudioContext}} to <var>context</var>.
 
-	2. Set its value for {{AudioNode/numberOfInputs}}, {{AudioNode/numberOfOutputs}}, {{AudioNode/channelCount}}, {{AudioNode/channelCountMode}}, {{AudioNode/channelInterpretation}} to the default value for this
+	2. Set its value for {{AudioNode/numberOfInputs}},
+		{{AudioNode/numberOfOutputs}}, {{AudioNode/channelCount}},
+		{{AudioNode/channelCountMode}}, {{AudioNode/channelInterpretation}} to the
+		default value for this
 		specific interface outlined in the section for each {{AudioNode}}.
 
-	3. If the {{AudioNode}} being constructed is a
-		{{ConvolverNode}}, set its {{ConvolverNode/normalize}} attribute with the inverse of the
-		value of the {{ConvolverOptions/disableNormalization}} in <var>dict</var>, and
-		then set its {{ConvolverNode/buffer}} attribute
-		to the value of the {{ConvolverOptions/buffer}}
-		in <var>dict</var> member, in this order, and jump to the last step
-		of this algorithm.
+	3. For each member of <var>dict</var> passed in, execute these steps, with
+		<var>k</var> the key of the member, and <var>v</var> its value. If any
+		exceptions is thrown when executing these steps, abort the iteration and
+		propagate the exception to the caller of the algorithm (constructor or
+		factory method).
 
-		Note: This means that the buffer will be normalized according to the
-		value of the {{ConvolverNode/normalize}}
-		attribute.
-
-	4. For each member of <var>dict</var> passed in, execute these
-		steps, with <var>k</var> the key of the member, and <var>v</var>
-		its value:
-
-		1. If <var>k</var> is {{ConvolverOptions/disableNormalization}} or
-			{{ConvolverOptions/buffer}} and <var>n</var> is {{ConvolverNode}},
-			jump to the beginning of this loop.
-
-		2. If <var>k</var> is the name of an {{AudioParam}} on this
+		1. If <var>k</var> is the name of an {{AudioParam}} on this
 			interface, set the {{AudioParam/value}}
 			attribute of this {{AudioParam}} to <var>v</var>.
 
-		3. Else if <var>k</var> is the name of an attribute on this
+		2. Else if <var>k</var> is the name of an attribute on this
 			interface, set the object associated with this attribute to
 			<var>v</var>.
 </div>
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -6576,11 +6560,22 @@ Constructors</h4>
...
+			value of {{ConvolverOptions/disableNormalization}}.
+		2. If {{ConvolverNode/buffer}} is
+			<a href="https://heycam.github.io/webidl/#dfn-present">present</a>, set the
+			{{ConvolverNode/buffer}} attribute to its value.
+
+			Note: This means that the buffer will be normalized according to the
+			value of the {{ConvolverNode/normalize}}
+			attribute.
+
+		3. Let <var>o</var> be an empty dictionnary.
```
karlt: We still want to process the `AudioNodeOptions` members on `options`.  One way to do that without the `ConvolverNode` special case in the "initialize" algorithm would be to let `o` be a new `AudioNodeOptions` dictionary with each dictionary member set to the corresponding value in `options`, if present.

index.bs
```diff
@@ -7905,14 +7911,14 @@ Constructors</h4>
 			4. Sort the elements in <var>tracks</var> based on their <code>id</code>
 				attribute using lexicographic ordering on sequences of code unit
 				values.
-			5. Let <var>node</var> be a new {{MediaStreamAudioSourceNode}}
-				object. <a href="#audionode-constructor-init">Initialize</a>
-				<var>node</var>.
-			6. Set an internal slot <dfn attribute for="MediaStreamAudioSourceNode">[[input track]]</dfn> on this
+			5. Let <var>o</var> be an empty dictionnary.
+			6. <a href="#audionode-constructor-init">Initialize the AudioNode</a>
+				<var>this</var>, with <var>c</var> and <var>o</var> as arguments.
```
karlt: Here, I assume `options` can be passed directly to the "initialize" algorithm, as the mediaStream attribute should be set on the node?

index.bs
```diff
@@ -8003,9 +8009,9 @@ Constructors</h4>
 		2. If the {{MediaStreamTrackAudioSourceOptions/mediaStreamTrack}}'s
 			<code>kind</code> attribute is not <code>"audio"</code>, throw an
 			{{InvalidStateError}} and abort these steps.
-		3. Let <var>node</var> be a new {{MediaStreamTrackAudioSourceNode}} object.
-			<a href="#audionode-constructor-init">Initialize</a> <var>node</var>, and
-			return <var>node</var>.
+		3. Let <var>o</var> be an empty dictionnary.
+		4. <a href="#audionode-constructor-init">Initialize the AudioNode</a>
+			<var>this</var>, with <var>c</var> and <var>option</var> as arguments.
```
karlt: Here, `options` can be passed directly to the "initialize" algorithm, as the mediaStreamTrack member will be ignored.

index.bs
padenot: That's right, I totally forgot that this was something.

padenot:
I've address the comments from @karlt in the last patch.

index.bs
```diff
@@ -4608,9 +4585,11 @@ Constructors</h4>
 <dl dfn-type=method dfn-for="AudioBufferSourceNode/constructor">
 	: <dfn>AudioBufferSourceNode(context, options)</dfn>
 	::
-		Let <var>node</var> be a new {{AudioBufferSourceNode}}
-		object. <a href="#audionode-constructor-init">Initialize</a>
-		<var>node</var>, and return <var>node</var>.
+
+		When the constructor is called with a {{BaseAudioContext}} <var>c</var> and
```
rtoy: Nit: since the parameters of the constructor are `context` and `options`, maybe use `<var>context</var>` and `<var>options</var>` instead?

index.bs
```diff
@@ -4152,6 +4124,11 @@ Constructors</h4>
 <dl dfn-type=constructor dfn-for="AnalyserNode">
 	: <dfn>AnalyserNode(context, options)</dfn>
 	::
+		When the constructor is called with a {{BaseAudioContext}} <var>c</var> and
+		an option object <var>option</var>, the user agent MUST <a
+		href="#audionode-constructor-init">initialize the AudioNode</a>
```
rtoy: Isn't there a bikeshed way to reference a section of the doc instead of using href?

index.bs
```diff
@@ -4152,6 +4124,11 @@ Constructors</h4>
 <dl dfn-type=constructor dfn-for="AnalyserNode">
 	: <dfn>AnalyserNode(context, options)</dfn>
 	::
+		When the constructor is called with a {{BaseAudioContext}} <var>c</var> and
```
rtoy: I didn't check everywhere, but it looks like this paragraph is the same for every node.  Perhaps a bikeshed macro would be useful here so that there's only one place to fix things and it will get updated everywhere.

index.bs
padenot: Non, only sections for some reason.

index.bs
padenot: It is exactly the same, I've factored it out in a macro.

index.bs
padenot: Yes, fixed (in the macro).

padenot:
The last commit addresses comment from @rtoy.

PR #2093, opened 2019-10-29
 Fix inconsistencies between the prose and algorithm for the offset parameter of the AudioBufferSourceNode start method

padenot:
This fixes the first part of #2072.

***
<a href="https://pr-preview.s3.amazonaws.com/padenot/web-audio-api/pull/2093.html" title="Last updated on Oct 29, 2019, 4:13 PM UTC (73e1644)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/2093/ef3262d...padenot:73e1644.html" title="Last updated on Oct 29, 2019, 4:13 PM UTC (73e1644)">Diff</a>

Issue #2072, opened 2019-10-03
Clarification on `AudioBufferSourceNode` looping, and loop points


padenot:
The spec is weird (emphasis mine):

> The offset parameter supplies a playhead position where playback will begin. If 0 is passed in for this value, then playback will start from the beginning of the buffer. A RangeError exception MUST be thrown if offset is negative. **If offset is greater than loopEnd, playback will begin at loopEnd (and immediately loop to loopStart)**. offset is silently clamped to [0, duration], when startTime is reached, where duration is the value of the duration attribute of the AudioBuffer set to the buffer attribute of this AudioBufferSourceNode. 

There are a number of issues and questions here, all answered by the algorithm:

- This is true only when `playbackRate` is > 0. If no, then it will go backward in the loop. The normative algorithm says yes, only when looping is enabled. This could be fixed by changing the prose.
- Does this need to happen **only** when looping is enabled? The algorithm, which is normative says says yes. This is fine, but we need to change the prose.
- Does the opposite happen when `playbackRate` is < 0, meaning the offset should be clamped in `[loopStart, loopEnd]` ? The normative algorithm says no. This seems inconsistent, but is the current normative requirement.
- In this case, is it allowed to start after the looped region, and then enter the looped region like it is for the positive `playbackRate` case? The normative algorithm says no. his seems inconsistent, but is the current normative requirement.

# Examples:

## Normal case
With a buffer of `3.0` seconds: starting, say, at `0.0`, with a `playbackRate` of `1.0`, and end/start loop points of respectively `1.0` and `2.0`, will play `[0.0, 1.0]`, then looping continuously in `[1.0, 2.0]`.

## Backward case
With a buffer of `3.0` second: starting, say, at `3.0`, with a `playbackRate` of `-1.0`, and end/start loop points of respectively `1.0` and `2.0`, will play `[2.0, 3.0]` backwards, then looping continuously (also backwards) in `[1.0, 2.0]`.

padenot:
Call resolution:
- At least fix the first two bullet point, by considering the algorithm is right, and adjusting the prose
- Have a look to see if fixing the two other points is easy. It probably is. There is an implementation in webkit iirc.

Issue #2031, opened 2019-08-20
AudioBufferSourceNode: Clarify offset truncation for negative playbackRates


collares:
**Describe the issue**
The spec says "If offset is greater than loopEnd, playback will begin at loopEnd (and immediately loop to loopStart)". If I understand the intent of the spec correctly, then this applies only when playbackRate is non-negative. The similar instruction "For negative playbackRate, if offset is less than loopStart, playback will begin at loopStart (and immediately loop to loopEnd)" applies in the other case.

**Where Is It**
In the offset argument for https://webaudio.github.io/web-audio-api/#dom-audiobuffersourcenode-start

padenot:
This is #2033 but backward isn't it? It makes sense to fix it in the same way (but in the opposite direction of course).

collares:
This issue is about the following asymmetry: 

- When `playbackRate` is positive, it is possible to start before `loopStart` and play audio until you reach the loop for the first time.
- Therefore, when `playbackRate` is negative, I would expect the behavior to be symmetric: It should be possible to start after `loopEnd` and let the audio play until reaching the loop. But, currently, the spec says that `offset > loopEnd` results in immediate truncation. I believe this makes sense for the positive playbackRate case, but not for the negative one.

In other words, #2033 deals with "positive `playbackRate`, playhead after `loopEnd`". The symmetric version of #2033 would be "negative `playbackRate`, playhead before `loopStart"`. The current issue refers to "negative `playbackRate`, playhead after `loopEnd`". 

Solutions for #2033 that remove the (currently specified) `offset` parameter truncation should fix this issue as well, but #2033 at the moment refers to dynamically-changing loop bounds only, for which the truncation behavior is currently unspecified.

padenot:
I had opened #2072 that was related, and opened two PRs for this.

PR #2094, opened 2019-10-29
Generalize the algorithm and the prose for AudioBufferSourceNode to have symmetric behaviour between forward and reverse playback

padenot:
This is based on #2093, and fixes the second part of #2072, which is also #2031.

***
<a href="https://pr-preview.s3.amazonaws.com/padenot/web-audio-api/pull/2094.html" title="Last updated on Nov 14, 2019, 1:17 PM UTC (23a2173)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/2094/1669e4c...padenot:23a2173.html" title="Last updated on Nov 14, 2019, 1:17 PM UTC (23a2173)">Diff</a>


index.bs
```diff
@@ -4750,7 +4750,7 @@ Methods</h4>
...
...eter supplies a <a>playhead position</a> where playback will begin. If 0 is passed in for this value, then playback will start from the beginning of the buffer. <span class="synchronous">A {{RangeError}} exception MUST be thrown if <code>offset</code> is negative.</span> If <code>offset</code> is greater than {{AudioBufferSourceNode/loopEnd}}, {{AudioBufferSourceNode/playbackRate}} is positive or zero, and {{AudioBufferSourceNode/loop}} is <code>true</code>, playback will begin at {{AudioBufferSourceNode/loopEnd}} (and immediately loop to {{AudioBufferSourceNode/loopStart}}).  If <code>offset</code> is greater than {{AudioBufferSourceNode/loopStart}}, {{AudioBufferSourceNode/playbackRate}} is negative, and {{AudioBufferSourceNode/loop}} is <code>true</code>, playback will begin at {{AudioBufferSourceNode/loopStart}} (and immediately loop to {{AudioBufferSourceNode/loopEnd}}). <code>offset</code> is silently clamped to [0, <code>duration</code>], when <code>startTime</code> is reached, where <code>duration</code> is the value of the <code>duration</code> attribute of the {{AudioBuffer}} set to the {{AudioBufferSourceNode/buffer}} attribute of this <code>AudioBufferSourceNode</code>.
```
rtoy: If offset > loopEnd, and  playbackRate is 0, how would it immediately loop to loopStart?

Really, I think we should give up trying to describe these complicated things and just reference the algorithm.

index.bs
```diff
@@ -5008,6 +5008,12 @@ function process(numberOfFrames) {
 		if (!started) {
 			// Take note that buffer has started playing and get initial
 			// playhead position.
+			if (computedPlaybackRate &gt;= 0 &amp;&amp; loop &amp;&amp; offset &gt;= actualLoopEnd) {
```
rtoy: Maybe it would be good to replace `<pre>` with `<xmp>` here so that we don't have to escape > and & and so on.

Maybe put `loop` first to emphasize this (and below) is all about looping?

index.bs
padenot: I can do that in another PR if we feel like it's what we want.

index.bs
padenot: Fixed both. Much better indeed, thanks.

index.bs
padenot: My concern here is that this is describing the method calls, and the algorithm the rendering.

padenot:
@rtoy I addressed you comments.

PR #2105, opened 2019-11-22
Empty the pending data slot when an exception is thrown in the AWP construction

### LINKED ISSUE #2096: Lifetime of pending processor construction data and exceptions in instantiation of AudioWorkletProcessor
[The instantiation of AudioWorkletProcessor algorithm](https://webaudio.github.io/web-audio-api/#AudioWorkletProcessor-instantiation) sets the node reference and transferred port of the AudioWorkletGlobalScope's pending processor construction data.  AFAICS this data is cleared only in the [AudioWorkletProcessor() constructor](https://webaudio.github.io/web-audio-api/#dom-audioworkletprocessor-audioworkletprocessor).

If an exception is thrown between setting the pending processor construction data and AudioWorkletProcessor construction, then that leaves the pending processor constructor data available after the instantiation of AudioWorkletProcessor algorithm completes.  The pending processor construction data can then be used at another time for construction of an AudioWorkletProcessor and to obtain a reference to the MessagePort entangled with the MessagePort on the AudioWorkletNode.  Is this intentional?

The [rendering loop](https://webaudio.github.io/web-audio-api/#rendering-loop) can run before the AudioWorkletProcessor is constructed in this way, in which case "Let processor be the associated AudioWorkletProcessor instance of AudioWorkletNode" can find there is no associated AudioWorkletProcessor.  Would `process()` be called on subsequent iterations when the AudioWorkletProcessor is eventually constructed?

Should the pending processor construction data instead be cleared at the completion of the instantiation of AudioWorkletProcessor algorithm, even when an exception is thrown during "Construct a callback function from _processorCtor_"?
Should a ["processorerror"](https://webaudio.github.io/web-audio-api/#dom-audioworkletnode-onprocessorerror) also be fired after an exception during the instantiation of AudioWorkletProcessor algorithm?

hoch:
Fixes #2096.

cc @karlt

***
<a href="https://pr-preview.s3.amazonaws.com/hoch/web-audio-api/pull/2105.html" title="Last updated on Nov 22, 2019, 10:37 PM UTC (b547463)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/2105/2f03431...hoch:b547463.html" title="Last updated on Nov 22, 2019, 10:37 PM UTC (b547463)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -10300,14 +10300,20 @@ Constructors</h5>
 				Throw a {{TypeError}} exception if the slot is
 				empty.
 
-			1. If any of following steps throws any exception,
-				abort the rest of the steps and
-				<a>queue a task</a> to the <a>control thread</a>
-				to <a spec="dom" lt="fire an event">fire an
-				event</a> named <code>processorerror</code>
-				using
-				<a href="https://www.w3.org/TR/html50/webappapis.html#the-errorevent-interface">ErrorEvent</a>
-				at <var>nodeReference</var>.
+			1. If any of following steps throws any exception:
+
+				1. Empty the
+					[=pending processor construction data=]
+					slot.
+
+				2. Abort the rest of the steps and
+					<a>queue a task</a> to the
+					<a>control thread</a>
+					to <a spec="dom" lt="fire an event">fire
+					an event</a> named
+					<code>processorerror</code> using
+					<a href="https://www.w3.org/TR/html50/webappapis.html#the-errorevent-interface">ErrorEvent</a>
+					at <var>nodeReference</var>.
 
 			1. Let <var>processor</var> be the
 				<a spec="webidl" lt="this">this</a> value.
--- a/index.bs
+++ b/index.bs
@@ -10301,7 +10301,7 @@ Constructors</h5>
 				empty.
 
 			1. If any of following steps throws any exception,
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -10300,14 +10300,20 @@ Constructors</h5>
...
 
-			1. If any of following steps throws any exception,
-				abort the rest of the steps and
-				<a>queue a task</a> to the <a>control thread</a>
-				to <a spec="dom" lt="fire an event">fire an
-				event</a> named <code>processorerror</code>
-				using
-				<a href="https://www.w3.org/TR/html50/webappapis.html#the-errorevent-interface">ErrorEvent</a>
-				at <var>nodeReference</var>.
+			1. If any of following steps throws any exception:
```
rtoy: Phrasing here is a bit awkward.  Perhaps it would sound better if you said:
> If any of following steps throws any exception, perform these steps to handle the exception

Can't think of a good way to phrase this because "following steps" is a bit ambiguous. Is "following steps" the new steps you've added as a sub step of this one, or is it the other steps in this algorithm at the same level as this item?

hoch: I rephrased a bit. PTAL again.

PR #2104, opened 2019-11-22
Remove `IsCallable` check on process() method in `registerProcess()` call

hoch:
Fixes #2045.

***
<a href="https://pr-preview.s3.amazonaws.com/hoch/web-audio-api/pull/2104.html" title="Last updated on Nov 22, 2019, 9:51 PM UTC (20d6249)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/2104/2f03431...hoch:20d6249.html" title="Last updated on Nov 22, 2019, 9:51 PM UTC (20d6249)">Diff</a>

PR #2106, opened 2019-12-03
Create PULL_REQUEST_TEMPLATE.md

marcoscaceres:
Recent situation come up where a change was made to the spec, but no corresponding web platform test was added. This ended up causing interop issues in Gecko. To prevent this, I'll encourage the WG to adopt this template. It helps guard against changes to being made to the spec without having both a Web Platform Test and public implementation commitments from the various browser engines.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,11 @@
+Closes #???
+
+The following tasks have been completed:
+
+ * [ ] Modified Web platform tests (link to pull request)
+
+Implementation commitment:
+
+ * [ ] WebKit (https://bugs.webkit.org/show_bug.cgi?id=)
+ * [ ] Chromium (https://bugs.chromium.org/p/chromium/issues/detail?id=)
+ * [ ] Gecko (https://bugzilla.mozilla.org/show_bug.cgi?id=)


rtoy:
What was the interop issue?

marcoscaceres:
@bzbarsky can probably provide details.

bzbarsky:
There wasn't an interop issue, because a Gecko engineer happened to notice the spec change and applied the corresponding change to Gecko.  But the lack of tests for the new behavior is still a problem...

rtoy:
It would certainly be helpful to know what the change in the spec was.

bzbarsky:
I don't really like putting people on the spot like that, but since you exist: https://github.com/WebAudio/web-audio-api/pull/2104

rtoy:
Thank you.  We all make mistakes.

marcoscaceres:
Note that HTML has also adopted a similar template:
https://github.com/whatwg/html/blob/master/PULL_REQUEST_TEMPLATE.md

And well as various working groups, including WebApps WG, Web Payments WG, and DAS. It’s quite helpful at preventing these kinds of mistakes.

rtoy:
The Audio WG is fairly small, and basically every PR has support from at least two vendors.  PRs generally also link to the issue which generally contains all of the relevant discussions. (This could be improved.)

marcoscaceres:
Sure, feel free to adapt to suit the group’s needs. It’s minimal for that reason.

rtoy:
Teleconf:  Add this for V2, after V1 recommendation.

rtoy:
I've filed WebAudio/web-audio-api-v2#60 to track this and to link to this PR so we can migrate it to v2.

As mentioned in https://github.com/WebAudio/web-audio-api/pull/2106#issuecomment-565109608, we're not going to do that for this repo, but will for the v2 repo.

Closing the PR, but we'll migrate it later.

Issue #2107, opened 2019-12-06
Update references to HTML user activation behavior


rtoy:
**Describe the issue**
The link for "triggered by user activation" no longer works.

**Where Is It**
In https://webaudio.github.io/web-audio-api/#AudioContext, see the paragraph just below the context IDL block.

**Additional Information**
This was updated in whatwg/html#5129 and https://github.com/whatwg/html/issues/5129#issuecomment-562210730 has some guidelines on what to do.

Without fixing this, the doc has errors in it. (Well a new error.)

PR #2109, opened 2019-12-06
Fix #2071: cancelScheduledValues with setValueCurveAtTime

### LINKED ISSUE #2071: cancelScheduledValues with setValueCurveAtTime
**Describe the issue**
Consider the following bit of code:
```
time = ctx.currentTime;
gain.gain.setValueCurveAtTime(curve, time + 0.1, 1);
gain.gain.cancelScheduledValues(time + 0.2);
gain.gain.setValueCurveAtTime(curve, time + 0.3, 1);
```
The second call to `setValueCurveAtTime` will throw an error because it overlaps the first.  Basically `cancelScheduledValues` didn't remove the first `setValueCurveAtTime`.

A strict reading of the spec says this is the correct behavior because the event time of the first setValueCurve is `time + 0.1`, which is before the cancel time of `time + 0.2`.  Hence it isn't removed.

The result is that you can't actually remove a `setValueCurve`; you have to wait until after the `setValueCurve` has ended before scheduling anything else.

The spec is pretty clear on what the behavior is, but I think we do really want to be able to remove the `setValueCurve` event if it ends some time after the cancel time.

`setTargetAtTime` kind of has the same issue, but you can always schedule events after it without errors, so perhaps we don't need to do anything here for that.

**Where Is It**
See [cancelScheduledValues](https://webaudio.github.io/web-audio-api/#dom-audioparam-cancelscheduledvalues)

Also see [StackOverflow question on this](https://stackoverflow.com/questions/58204307/cancelscheduledvaluesx-followed-by-setvaluecurveattimex-throws-error).

### LINKED ISSUE #2095: Specify which task source is used
The [current specification](https://webaudio.github.io/web-audio-api/) doesn't specify which task source is used whenever queuing a task.

It seems like we may want to use the [media element event task source](https://html.spec.whatwg.org/multipage/media.html#media-element-event-task-source) here?

Related: https://github.com/WebAudio/web-audio-api/issues/2008

rtoy:
If the cancelTime overlaps the setValueCurveAtTime event, remove the
event too.

***
<a href="https://pr-preview.s3.amazonaws.com/rtoy/web-audio-api/pull/2109.html" title="Last updated on Jan 13, 2020, 4:22 PM UTC (e4f203d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/2109/e139bd4...rtoy:e4f203d.html" title="Last updated on Jan 13, 2020, 4:22 PM UTC (e4f203d)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3555,6 +3555,12 @@ Methods</h4>
 		are also removed if the hold time occurs after
 		{{AudioParam/cancelScheduledValues()/cancelTime!!argument}}.
 
+		For a {{AudioParam/setValueCurveAtTime()}}, let \(T_0\) and \(T_D\) be the corresponding 
+		{{AudioParam/setValueCurveAtTime()/startTime!!argument}} and {{AudioParam/setValueCurveAtTime()/duration!!argument}}, respectively of this event.
+		Then if {{AudioParam/cancelScheduledValues()/cancelTime!!argument}}
+		is in the range \([T_0, T_0 + T_D]\), the event is
+		removed from the timeline.
+
 		<pre class=argumentdef for="AudioParam/cancelScheduledValues()">
 			cancelTime: The time after which any previously scheduled parameter changes will be cancelled. It is a time in the same time coordinate system as the {{AudioContext}}'s {{BaseAudioContext/currentTime}} attribute. <span class="synchronous">A {{RangeError}} exception MUST be thrown if <code>cancelTime</code> is negative or is not a finite number.</span> If <code>cancelTime</code> is less than {{BaseAudioContext/currentTime}}, it is clamped to {{BaseAudioContext/currentTime}}.
 		</pre>


rtoy:
This is blocked on #2107

rtoy:
#2107 is fixed now and merging the changes fixes the build error.

rtoy:
I explicitly left that undefined because we don't say what happens if you cancel events that are currently running.  If we want to define that, then that's a different issue.

But a quick test with Chrome shows that if a linear ramp is already running and then cancel events between current time and end of the ramp, then Chrome holds the ramp value at the cancel time. (If the ramp hasn't started, then the output is different.)

padenot:
> I explicitly left that undefined because we don't say what happens if you cancel events that are currently running. If we want to define that, then that's a different issue.

It would be best to fix that.

PR #2116, opened 2019-12-19
Add a note about MessagePort listener and garbage collection

### LINKED ISSUE #2060: Would it be helpful to restrict use of `AudioWorkletProcessor.port().postMessage()` in order to facilitate garbage collection?
Consider the [VU Meter Node example](https://webaudio.github.io/web-audio-api/#vu-meter-mode), but followed by `oscillator.stop()` and `cancelAnimationFrame(drawMeter)`.  The main global scope has no references to `vuMeterNode` or to `vuMeterNode.port`, the input to the associated processor falls silent, and the `process()` method returns `false`.  `process()` will not be called again, and the `AudioWorkletGlobalScope` has no references to the processor or its `port`.

Theoretically, an implementation could determine that `vuMeterNode`, its processor, and their `MessagePort`s will not be accessed again, and so could recover all resources associated with `vuMeterNode`.  Currently, however, no browsers have the capability to do this.

If an entangled port `portA` can be accessed from script and its [[RemotePort]] `portB` has a listener, then `portA` must keep `portB` alive to receive messages.  If `portB` may receive messages, then the message listener is able to send messages back.  If `portA` has a listener, then `portB` must keep `portA` alive to receive messages.  There is an ownership cycle.

To my knowledge, no browser currently has the ability to detect when such cycles are garbage if `portA` and `portB` are on different threads, as would be the case with AudioWorkletNode.  Detection of otherwise unreferenced cycles of this nature may require significant architectural changes, one complication being that an approach that required stopping both threads to analyse ownership graphs would not be acceptable when a realtime audio thread is involved.

One alternative option might be to non-normatively document that a client which listens on `AudioWorkletNode.port` should call `close()` on either of the ports to allow resources to be recovered before page unload.  I feel like this would be passing the buck to the client to find a solution.  I'd like to consider whether there is a better option.  (The second note at https://html.spec.whatwg.org/multipage/web-messaging.html#ports-and-garbage-collection is similar to this option but misleads that lack of a `close()` call can lead to only transient memory usage.) 

The proposal to consider is whether `AudioWorkletProcessor.port().postMessage()` would throw when invoked if neither during the _processFunction_ invocation for the associated AudioWorkletNode in the [rendering loop](https://webaudio.github.io/web-audio-api/#rendering-loop) nor during `message` or `messageerror` event handlers for this `AudioWorkletProcess.port`.
When an AudioWorkletNode is not actively processing, and there are no messages in flight for its ports, this restriction would allow an implementation to more easily determine that the `AudioWorkletProcessor.port` could not send messages and so would not need to keep its entangled `AudioWorkletNode.port` alive.  When all ownership is in the direction from main global scope to AudioWorkletGlobalScope, there are no inter-thread cycles, main global scope objects can be detected as garbage without consideration of associated AudioWorkletGlobalScope objects, AudioWorkletNodes can be released in the same way as other AudioNodes, and their associated AudioWorkletGlobalScope objects can then also be released. 

What are people's thoughts?

An implementation cannot assume that `AudioWorkletProcessor.port.postMessage()` would only be called during a callback on this processor or its port because a client can share a reference to the port via the shared `AudioWorkletGlobalScope`.

padenot:
This fixes #2060.

***
<a href="https://pr-preview.s3.amazonaws.com/padenot/web-audio-api/pull/2116.html" title="Last updated on Jan 7, 2020, 11:21 AM UTC (baaa070)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/2116/e3bed78...padenot:baaa070.html" title="Last updated on Jan 7, 2020, 11:21 AM UTC (baaa070)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -10163,6 +10163,11 @@ Attributes</h5>
 		bidirectional communication between the
 		{{AudioWorkletNode}} and its {{AudioWorkletProcessor}}.
 
+		Note: Authors that register a event listener on the <code>"message"</code>
+		event of this {{AudioWorkletNode/port}} should call {{MessagePort/close}} on
+		either end of the {{MessageChannel}} (either in the
+		{{AudioWorkletProcessor}} or the {{AudioWorkletNode}} side) to allow for
+		ressources to be [[html#ports-and-garbage-collection|collected]].
 </dl>
 
 <h5 dictionary lt="AudioWorkletNodeOptions">
@@ -10369,6 +10374,12 @@ Constructors</h5>
 		the port on the corresponding {{AudioWorkletNode}} object
 		allowing bidirectional communication between an
 		{{AudioWorkletNode}} and its {{AudioWorkletProcessor}}.
+
+		Note: Authors that register a event listener on the <code>"message"</code>
+		event of this {{AudioWorkletProcessor/port}} should call
+		{{MessagePort/close}} on either end of the {{MessageChannel}} (either in the
+		{{AudioWorkletProcessor}} or the {{AudioWorkletNode}} side) to allow for
+		ressources to be [[html#ports-and-garbage-collection|collected]].
 </dl>
 
 <h5 id="AudioWorkletProcessor-methods">
--- a/index.bs
+++ b/index.bs
@@ -10167,7 +10167,7 @@ Attributes</h5>
 		event of this {{AudioWorkletNode/port}} should call {{MessagePort/close}} on
 		either end of the {{MessageChannel}} (either in the
 		{{AudioWorkletProcessor}} or the {{AudioWorkletNode}} side) to allow for
-		ressources to be [[html#ports-and-garbage-collection|collected]].
+		resources to be [[html#ports-and-garbage-collection|collected]].
 </dl>
 
 <h5 dictionary lt="AudioWorkletNodeOptions">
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -10138,6 +10138,11 @@ Attributes</h5>
 		bidirectional communication between the
 		{{AudioWorkletNode}} and its {{AudioWorkletProcessor}}.
 
+		Note: Authors that register a event listener on the <code>"message"</code>
+		event of this {{AudioWorkletNode/port}} should call {{MessagePort/close}} on
+		either end of the {{MessageChannel}} (either in the
+		{{AudioWorkletProcessor}} or the {{AudioWorkletNode}} side) to allow for
+		ressources to be [[html#ports-and-garbage-collection|collected]].
```
hoch: ressources => resources

index.bs
```diff
@@ -10344,6 +10349,12 @@ Constructors</h5>
 		the port on the corresponding {{AudioWorkletNode}} object
 		allowing bidirectional communication between an
 		{{AudioWorkletNode}} and its {{AudioWorkletProcessor}}.
+
+		Note: Authors that register a event listener on the <code>"message"</code>
+		event of this {{AudioWorkletProcessor/port}} should call
+		{{MessagePort/close}} on either end of the {{MessageChannel}} (either in the
+		{{AudioWorkletProcessor}} or the {{AudioWorkletNode}} side) to allow for
+		ressources to be [[html#ports-and-garbage-collection|collected]].
```
hoch: ditto

PR #2117, opened 2019-12-19
Fix #2113:  BiquadFilter gain lower bound is lower

### LINKED ISSUE #2113: BiquadFilter gain lower bound can be lower.
In #2087, we changed the lower bound for the gain AudioParam for a `BiquadFilterNode`.  It's about -1541 now.  However, any values less than this should just silently underflow to produce a linear gain value of 0, so maybe we should make the lower bound the most negative single float?

rtoy:
Since the filter gain is in dB, any finite negative value is fine
since the value should just underflow to a linear gain value of
0. (Unlike the upper bound which would cause overflow if it were
higher.)

***
<a href="https://pr-preview.s3.amazonaws.com/rtoy/web-audio-api/pull/2117.html" title="Last updated on Dec 19, 2019, 6:06 PM UTC (ae27e4d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/2117/73726d1...rtoy:ae27e4d.html" title="Last updated on Dec 19, 2019, 6:06 PM UTC (ae27e4d)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5949,8 +5949,8 @@ Attributes</h4>
 		path: audioparam.include
 		macros:
 			default: 0
-			min: \(\approx -1541\)
-			min-notes:
+			min: <a>most-negative-single-float</a>
+			min-notes: Approximately 3.4028235e38
 			max: \(\approx 1541\)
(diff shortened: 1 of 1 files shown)


rtoy:
Note:  the preview won't show the fix because the preview doesn't include the audioparam include file that contains the change.

PR #2166, opened 2020-02-18
Rebase the document unloading steps

padenot:
This version converts to bikeshed markup, harmonizes the names of the
slots (`[[pending promises]]` and `[[pending resume promises]]`), and
hooks the slots of the interface that makes the most sense
(respectively, `BaseAudioContext` and `AudioContext`). The commit message for the first commit explains the original commit.

There is no provision for doing anything with the `addModule` promise,
this should be in the worklet text.

Another commit fixes the trailing white-spaces.

***
<a href="https://pr-preview.s3.amazonaws.com/padenot/web-audio-api/pull/2166.html" title="Last updated on Feb 21, 2020, 1:11 PM UTC (e4c4876)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/2166/3ff9eea...padenot:e4c4876.html" title="Last updated on Feb 21, 2020, 1:11 PM UTC (e4c4876)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -644,6 +644,11 @@ but is instead extended by the concrete interfaces
 {{AudioContext}} (for real-time rendering) and
 {{OfflineAudioContext}} (for offline rendering).
 
+{{BaseAudioContext}} are created with an internal slot 
+<dfn attribute for="BaseAudioContext">[[pending promises]]</dfn>  that is an
+initialy empty ordered list of promises.
+
+
 <pre class="idl">
 enum AudioContextState {
 	"suspended",
@@ -1100,23 +1105,27 @@ Methods</h4>
 			called, the following steps MUST be performed on the control
 			thread:</span>
 
-			1. Let <var>promise</var> be a new promise.
+			1. Let <var>promise</var> be a new Promise.
 
 			2. If the operation <a href="https://tc39.github.io/ecma262/#sec-isdetachedbuffer"><code>IsDetachedBuffer</code></a>
 				(described in [[!ECMASCRIPT]]) on {{BaseAudioContext/decodeAudioData(audioData, successCallback, errorCallback)/audioData!!argument}} is
 				<code>false</code>, execute the following steps:
 
-				1. <a href="https://tc39.github.io/ecma262/#sec-detacharraybuffer">Detach</a>
-					the {{BaseAudioContext/decodeAudioData(audioData, successCallback,
-					errorCallback)/audioData!!argument}} {{ArrayBuffer}}. This operation is
-					described in [[!ECMASCRIPT]]. If this operations throws, jump to
-					the step 3.
-				2. Queue a decoding operation to be performed on another thread.
+				1. Append <var>promise</var> to {{BaseAudioContext/[[pending promises]]}}.
+
+				2. <a href="https://tc39.github.io/ecma262/#sec-detacharraybuffer">
+					Detach</a> the {{BaseAudioContext/decodeAudioData(audioData,
+					successCallback, errorCallback)/audioData!!argument}} {{ArrayBuffer}}.
+					This operation is described in [[!ECMASCRIPT]]. If this operations
+					throws, jump to the step 3.
+
+				3. Queue a decoding operation to be performed on another thread.
 
 			3. Else, execute the following error steps:
 
 				1. Let <var>error</var> be a {{DataCloneError}}.
-				2. Reject <var>promise</var> with <var>error</var>.
+				2. Reject <var>promise</var> with <var>error</var>, and remove it from
+					{{BaseAudioContext/[[pending promises]]}}.
 				3. <a>Queue a task</a> to invoke {{BaseAudioContext/decodeAudioData()/errorCallback!!argument}} with <var>error</var>.
 
 			4. Return <var>promise</var>.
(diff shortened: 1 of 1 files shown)


padenot:
Raymond, if you can have a quick look? It's functionally equivalent to the previous one you reviewed a while back.

index.bs
```diff
@@ -1683,14 +1718,18 @@ Methods</h4>
 			4. In case of failure, queue a task on the <a>control thread</a> to execute the following,
 				and abort these steps:
 
-				1. Reject all promises from <a>pendingResumePromises</a> in order, then clear <a>pendingResumePromises</a>.
+				1. Reject all promises from {{AudioContext/[[pending resume promises]]}} in order, then
+					clear {{AudioContext/[[pending resume promises]]}}. Additionaly, remove those promises
+					from {{BaseAudioContext/[[pending promises]]}}.
 
 				2. Reject <em>promise</em>.
```
rtoy: I think this promise can be rejected twice.  Once because it might be a member of `[[pending resume promises]]` and then once again in this step.  Is that ok?

index.bs
```diff
@@ -2088,22 +2142,28 @@ Methods</h4>
 			3. In case of failure, queue a task on the <a>control thread</a> to execute the following,
 				and abort these steps:
 
-				1. Reject all promises from <a>pendingResumePromises</a> in order, then clear <a>pendingResumePromises</a>.
+				1. Reject all promises from {{AudioContext/[[pending resume promises]]}}
```
rtoy: An `OfflineAudioContext` isn't an `AudioContext`, so shouldn't we define `[[pending resume promises]]` slot of an offline context?  And update links accordingly.

Also, should the promise be appended to `[[pending resume promises]]` in earlier steps?  Otherwise this is empty.

index.bs
```diff
@@ -2088,22 +2142,28 @@ Methods</h4>
...
 
 			4. Queue a task on the <a>control thread</a>'s event loop, to
 				execute these steps:
 
-				1. Resolve all promises from <a>pendingResumePromises</a> in order, then clear <a>pendingResumePromises</a>.
+				1. Resolve all promises from {{AudioContext/[[pending resume
+					promises]]}} in order, then clear {{AudioContext/[[pending resume
+					promises]]}}.
 
 				2. Resolve <em>promise</em>.
```
rtoy: This seems to resolve the promise twice.  Once in the previous step and then again here.

index.bs
```diff
@@ -11236,6 +11296,20 @@ running the algorithm for an {{AudioNode}}, using an <a>input
 buffer</a> and the value(s) of the {{AudioParam}}(s) of this
 {{AudioNode}} as the input for this algorithm.
 
+<h3 id="unloading-a-document">Unloading a document</h3>
+	Additional <a href=
+	"https://html.spec.whatwg.org/#unloading-document-cleanup-steps">unloading
+	document cleanup steps</a> are defined for document that uses
```
rtoy: Typo?  "document that uses" -> "documents that use"

index.bs
```diff
@@ -11236,6 +11296,20 @@ running the algorithm for an {{AudioNode}}, using an <a>input
...
+	{{BaseAudioContext}}:
+
+1. Reject all the promises of {{BaseAudioContext/[[pending promises]]}} with
+	<code>InvalidStateError</code>, for each {{AudioContext}} and
+	{{OfflineAudioContext}} whose relevant global object is the same as
+	the document's associated Window.
+2. Stop all {{decoding thread}}s.
+3. <a>Queue a control message</a> to close the
```
rtoy: "close" should link to the appropriate close algorithm?  Now it's a bit unclear on what "close" means.

index.bs
```diff
@@ -2088,22 +2142,28 @@ Methods</h4>
...
+				1. Reject all promises from {{AudioContext/[[pending resume promises]]}}
+					in order, then clear {{AudioContext/[[pending resume promises]]}}.
 
 				2. Reject <em>promise</em>.
 
 			4. Queue a task on the <a>control thread</a>'s event loop, to
 				execute these steps:
 
-				1. Resolve all promises from <a>pendingResumePromises</a> in order, then clear <a>pendingResumePromises</a>.
+				1. Resolve all promises from {{AudioContext/[[pending resume
```
rtoy: This seems inconsistent with how we handle `AudioContext`.  There, we clear `[[pending resume promises]]`.  Here we resolve them all in order.

index.bs
padenot: You're right, I've removed step 2, and split the first step, that was in fact multiple steps, in two.

index.bs
padenot: This is a mistake, I'm removing it. There is no `[[pending resume promises]]` on `OfflineAudioContext` as you note.

index.bs
padenot: Same mistake, I'm removing this.

index.bs
padenot: It's explained in a section in `close`:

>		Running a <a>control message</a> to close an
>		{{AudioContext}} means running these steps on the
>		<a>rendering thread</a>:
>              ...

and the same for `OfflineAudioContext`.

index.bs
```diff
@@ -1718,17 +1718,17 @@ Methods</h4>
...
 				and abort these steps:
 
-				1. Reject all promises from {{AudioContext/[[pending resume promises]]}} in order, then
-					clear {{AudioContext/[[pending resume promises]]}}. Additionaly, remove those promises
-					from {{BaseAudioContext/[[pending promises]]}}.
+				1. Reject all promises from {{AudioContext/[[pending resume promises]]}}
+					in order, then clear {{AudioContext/[[pending resume promises]]}}.
 
-				2. Reject <em>promise</em>.
+				2. Additionaly, remove those promises from {{BaseAudioContext/[[pending
```
rtoy: Typo: Additionaly -> Additionally.

PR #2168, opened 2020-02-18
Fix #2147: Clean up PeriodicWave constructor

### LINKED ISSUE #2147: PeriodicWave constructor examines lengths of things that might not be there
https://webaudio.github.io/web-audio-api/#PeriodicWave-constructors says (step 2):

> If the real and imag parameters of the PeriodicWaveOptions are not of the same length, an IndexSizeError exception MUST be thrown.

but those parameters might not be present at all.  Is this meaning to only compare the lengths when both are present?  Or does being not present correspond to a length of 0?  The text here needs to be a lot more explicit about what should happen.

Then step 4 says:

> Otherwise, let [[real]] and [[imag]] be two internal slots of type Float32Array, both of length equal to the maximum of the lengths of the real and imag attributes of the PeriodicWaveOptions passed in.

but the two things passed in must be the same length per step 2.  Again, unless being missing means skipping the length check.  But then what does the "maximum" mean if one of the lengths does not exist?  If both lengths do not exist?

There's something in step 3 about the "options has not been passed in" case, which can't happen, because `options` has a default value, so it's _always_ passed in.  It may or may not have `real` and `imag` present in it, of course, whether it's the default value or an explicit value passed from script.

rtoy:
Cleans up the constructor for PeriodicWave based on the algorithm given in
https://github.com/WebAudio/web-audio-api/issues/2147#issuecomment-586520459.

A few other minor cleanups in the algorithm description were also done:
*  Move declaration of `[[real]]`, `[[imag]]`, and `[[normalize]]` to the beginning of the algorithm.
* Remove extra text for the members of the `PeriodicWaveOptions` to the algorithm so that everthing is all in one place.

***
<a href="https://pr-preview.s3.amazonaws.com/rtoy/web-audio-api/pull/2168.html" title="Last updated on Feb 19, 2020, 5:50 PM UTC (33332d6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/2168/7bbfb8f...rtoy:33332d6.html" title="Last updated on Feb 19, 2020, 5:50 PM UTC (33332d6)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -8953,35 +8953,29 @@ Constructors</h4>
 	: <dfn>PeriodicWave(context, options)</dfn>
 	::
 		<div algorithm="construct periodic wave">
-			1. Let <var>p</var> be a new {{PeriodicWave}} object. Let
-				<dfn attribute for="PeriodicWave">[[associated context]]</dfn> be a reference to the
-				{{BaseAudioContext}} passed as first argument of this
-				constructor.
+			1. Let <var>p</var> be a new {{PeriodicWave}} object. Let <dfn attribute for="PeriodicWave">\[[real]]</dfn> and <dfn attribute for="PeriodicWave">\[[imag]]</dfn> be two internal slots of type {{Float32Array}}, and let <dfn attribute for="PeriodicWave">\[[normalize]]</dfn> be an internal slot.
 
-			2. <span class="synchronous">If the {{PeriodicWaveOptions/real}} and {{PeriodicWaveOptions/imag}} parameters of
-				the {{PeriodicWaveOptions}} are not of the same length, an
-				{{IndexSizeError}} exception MUST be thrown.</span>
+			1. Process {{PeriodicWave/PeriodicWave(context, options)/options!!argument}} according to one of the following cases:
+				1. If both {{PeriodicWaveOptions/real|options.real}} and {{PeriodicWaveOptions/imag|options.imag}} are given
+					1. <span class="synchronous">If the lengths are different or if either length is less than 2, throw an {{IndexSizeError}}error and abort this algorithm.</span>
+					1. Set {{[[real]]}} and {{[[imag]]}} to new arrays with a length of {{PeriodicWaveOptions/real|options.real}}
+					1. Copy all elements from {{PeriodicWaveOptions/real|options.real}} to {{[[real]]}} and {{PeriodicWaveOptions/imag|options.imag}} to {{[[imag]]}}
+				1. If only {{PeriodicWaveOptions/real|options.real}} is given
+					1. <span class="synchronous">If length is less than 2, throw an {{IndexSizeError}} and abort this algorithm.</span>
+					1. Set {{[[real]]}} and {{[[imag]]}} to arrays with the same length as {{PeriodicWaveOptions/real|options.real}}.
+					1. Copy {{PeriodicWaveOptions/real|options.real}} to {{[[real]]}} and set {{[[imag]]}} to all zeros.
+				1. If only {{PeriodicWaveOptions/imag|options.imag}} is given
+					1. <span class="synchronous">If length is less than 2, throw an {{IndexSizeError}} and abort this algorithm.</span>
+					1. Set {{[[real]]}} and {{[[imag]]}} to arrays with the same length as {{PeriodicWaveOptions/real|options.imag}}.
+					1. Copy {{PeriodicWaveOptions/imag|options.imag}} to {{[[imag]]}} and set {{[[real]]}} to all zeros.
+				1. Otherwise
+					1. Set {{[[real]]}} and {{[[imag]]}} to zero-filled arrays of length 2
+					1. Set element 1 of {{[[imag]]}} to 1.
 
-			3. If {{PeriodicWave/PeriodicWave(context, options)/options!!argument}} has not been passed in, let
-				<dfn attribute for="PeriodicWave">\[[real]]</dfn> and <dfn attribute for="PeriodicWave">\[[imag]]</dfn> be two internal
-				slots of type {{Float32Array}} and length 2. Set the
-				second element of the {{[[imag]]}} array to 1.
-
-				Note: When setting this {{PeriodicWave}} on an
-				{{OscillatorNode}}, this is equivalent to using the
-				built-in type <code>"sine"</code>.
-
-			4. Otherwise, let {{[[real]]}} and
-				{{[[imag]]}} be two internal slots of type
-				{{Float32Array}}, both of length equal to the maximum of the
-				lengths of the {{PeriodicWaveOptions/real}} and {{PeriodicWaveOptions/imag}} attributes of
-				the {{PeriodicWaveOptions}} passed in.
-				<a href="https://heycam.github.io/webidl/#dfn-get-buffer-source-copy">
-				Make a copy</a> of those arrays into their respective
-				internal slots.
-
-			5. Let <dfn attribute for="PeriodicWave">\[[normalize]]</dfn> be an internal slot of the
-				{{PeriodicWave}} that is initialized to the inverse of the
+					Note: When setting this {{PeriodicWave}} on an {{OscillatorNode}}, this is equivalent to using the built-in type "{{OscillatorType/sine}}".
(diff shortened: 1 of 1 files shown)


rtoy:
@bzbarsky PTAL to see if I missed anything.

index.bs
```diff
@@ -8953,35 +8953,29 @@ Constructors</h4>
...
-				internal slots.
-
-			5. Let <dfn attribute for="PeriodicWave">\[[normalize]]</dfn> be an internal slot of the
-				{{PeriodicWave}} that is initialized to the inverse of the
+			1. Let <var>p</var> be a new {{PeriodicWave}} object. Let <dfn attribute for="PeriodicWave">\[[real]]</dfn> and <dfn attribute for="PeriodicWave">\[[imag]]</dfn> be two internal slots of type {{Float32Array}}, and let <dfn attribute for="PeriodicWave">\[[normalize]]</dfn> be an internal slot.
+
+			1. Process {{PeriodicWave/PeriodicWave(context, options)/options!!argument}} according to one of the following cases:
+				1. If both {{PeriodicWaveOptions/real|options.real}} and {{PeriodicWaveOptions/imag|options.imag}} are given
+					1. <span class="synchronous">If the lengths are different or if either length is less than 2, throw an {{IndexSizeError}} and abort this algorithm.</span>
+					1. Set {{[[real]]}} and {{[[imag]]}} to new arrays with a length of {{PeriodicWaveOptions/real|options.real}}.
```
bzbarsky: Are these objects observable later as actual objects in JS?  That is, does is matter what Realm is used for these allocations?

Also "with a length of" might be better as "with the same length as", like the other cases.

index.bs
```diff
@@ -8953,35 +8953,29 @@ Constructors</h4>
...
-				Make a copy</a> of those arrays into their respective
-				internal slots.
-
-			5. Let <dfn attribute for="PeriodicWave">\[[normalize]]</dfn> be an internal slot of the
-				{{PeriodicWave}} that is initialized to the inverse of the
+			1. Let <var>p</var> be a new {{PeriodicWave}} object. Let <dfn attribute for="PeriodicWave">\[[real]]</dfn> and <dfn attribute for="PeriodicWave">\[[imag]]</dfn> be two internal slots of type {{Float32Array}}, and let <dfn attribute for="PeriodicWave">\[[normalize]]</dfn> be an internal slot.
+
+			1. Process {{PeriodicWave/PeriodicWave(context, options)/options!!argument}} according to one of the following cases:
+				1. If both {{PeriodicWaveOptions/real|options.real}} and {{PeriodicWaveOptions/imag|options.imag}} are given
+					1. <span class="synchronous">If the lengths are different or if either length is less than 2, throw an {{IndexSizeError}} and abort this algorithm.</span>
```
bzbarsky: "If the lengths of options.real and options.imag are differen or if ... "?

index.bs
```diff
@@ -8953,35 +8953,29 @@ Constructors</h4>
...
+					1. Set {{[[real]]}} and {{[[imag]]}} to new arrays with a length of {{PeriodicWaveOptions/real|options.real}}.
+					1. Copy all elements from {{PeriodicWaveOptions/real|options.real}} to {{[[real]]}} and {{PeriodicWaveOptions/imag|options.imag}} to {{[[imag]]}}.
+				1. If only {{PeriodicWaveOptions/real|options.real}} is given
+					1. <span class="synchronous">If length is less than 2, throw an {{IndexSizeError}} and abort this algorithm.</span>
```
bzbarsky: "If the length of options.real ..."

index.bs
```diff
@@ -8953,35 +8953,29 @@ Constructors</h4>
...
+					1. Set {{[[real]]}} and {{[[imag]]}} to arrays with the same length as {{PeriodicWaveOptions/real|options.real}}.
+					1. Copy {{PeriodicWaveOptions/real|options.real}} to {{[[real]]}} and set {{[[imag]]}} to all zeros.
+				1. If only {{PeriodicWaveOptions/imag|options.imag}} is given
+					1. <span class="synchronous">If length is less than 2, throw an {{IndexSizeError}} and abort this algorithm.</span>
```
bzbarsky: "if the length of options.imag"

index.bs
```diff
@@ -8953,35 +8953,29 @@ Constructors</h4>
...
-				<a href="https://heycam.github.io/webidl/#dfn-get-buffer-source-copy">
-				Make a copy</a> of those arrays into their respective
-				internal slots.
-
-			5. Let <dfn attribute for="PeriodicWave">\[[normalize]]</dfn> be an internal slot of the
-				{{PeriodicWave}} that is initialized to the inverse of the
+			1. Let <var>p</var> be a new {{PeriodicWave}} object. Let <dfn attribute for="PeriodicWave">\[[real]]</dfn> and <dfn attribute for="PeriodicWave">\[[imag]]</dfn> be two internal slots of type {{Float32Array}}, and let <dfn attribute for="PeriodicWave">\[[normalize]]</dfn> be an internal slot.
+
+			1. Process {{PeriodicWave/PeriodicWave(context, options)/options!!argument}} according to one of the following cases:
+				1. If both {{PeriodicWaveOptions/real|options.real}} and {{PeriodicWaveOptions/imag|options.imag}} are given
```
bzbarsky: s/given/present/.  See https://heycam.github.io/webidl/#dfn-present

index.bs
```diff
@@ -8953,35 +8953,29 @@ Constructors</h4>
...
+					1. <span class="synchronous">If the lengths are different or if either length is less than 2, throw an {{IndexSizeError}} and abort this algorithm.</span>
+					1. Set {{[[real]]}} and {{[[imag]]}} to new arrays with a length of {{PeriodicWaveOptions/real|options.real}}.
+					1. Copy all elements from {{PeriodicWaveOptions/real|options.real}} to {{[[real]]}} and {{PeriodicWaveOptions/imag|options.imag}} to {{[[imag]]}}.
+				1. If only {{PeriodicWaveOptions/real|options.real}} is given
```
bzbarsky: s/given/present/

index.bs
```diff
@@ -8953,35 +8953,29 @@ Constructors</h4>
...
+					1. <span class="synchronous">If length is less than 2, throw an {{IndexSizeError}} and abort this algorithm.</span>
+					1. Set {{[[real]]}} and {{[[imag]]}} to arrays with the same length as {{PeriodicWaveOptions/real|options.real}}.
+					1. Copy {{PeriodicWaveOptions/real|options.real}} to {{[[real]]}} and set {{[[imag]]}} to all zeros.
+				1. If only {{PeriodicWaveOptions/imag|options.imag}} is given
```
bzbarsky: s/given/present/

index.bs
```diff
@@ -8953,35 +8953,29 @@ Constructors</h4>
...
+					1. <span class="synchronous">If length is less than 2, throw an {{IndexSizeError}} and abort this algorithm.</span>
+					1. Set {{[[real]]}} and {{[[imag]]}} to arrays with the same length as {{PeriodicWaveOptions/real|options.imag}}.
+					1. Copy {{PeriodicWaveOptions/imag|options.imag}} to {{[[imag]]}} and set {{[[real]]}} to all zeros.
+				1. Otherwise
+					1. Set {{[[real]]}} and {{[[imag]]}} to zero-filled arrays of length 2.
+					1. Set element 1 of {{[[imag]]}} to 1.
```
bzbarsky: I think this might be clearer as "Set the element at index 1 ..."

index.bs
```diff
@@ -8953,35 +8953,29 @@ Constructors</h4>
...
+
+					Note: When setting this {{PeriodicWave}} on an {{OscillatorNode}}, this is equivalent to using the built-in type "{{OscillatorType/sine}}".
+			1. Set element 0 of both {{[[real]]}} and {{[[imag]]}} to 0.  (This sets the DC component to 0.)
```
bzbarsky: And here, "Set the element at index 0"

index.bs
rtoy: The internal slots are internal and aren't ever exposed to JS in any way.

rtoy:
Thanks for your review @bzbarsky. I've applied all of your suggestions.

rtoy:
Thanks for the review!

PR #2172, opened 2020-03-03
Allow not having get parameterDescriptors in an AudioWorkletProcessor

### LINKED ISSUE #2171: AudioWorkletProcessor definition without `get ParameterDescriptor`
Should this be allowed? Per spec, it shouldn't:

https://heycam.github.io/webidl/#es-sequence throws if the input is not iterable, and there is particular handling for when there is no `get parameterDescriptors`.

I see WPTs and examples in the spec that don't have `get parameterDescriptors`.

@hoch, I believe your intent there was to allow it, is that right?

padenot:
This allows omitting it if a particular Processor is not going to use
AudioParam. However, if the getter is specified, then it MUST return an
iterable.

This fixes #2171.

***
<a href="https://pr-preview.s3.amazonaws.com/padenot/web-audio-api/pull/2172.html" title="Last updated on Mar 3, 2020, 3:35 PM UTC (fe0a930)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/2172/d43886d...padenot:fe0a930.html" title="Last updated on Mar 3, 2020, 3:35 PM UTC (fe0a930)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -9847,49 +9847,52 @@ Methods</h5>
 				result of
 				<code><a href="http://www.ecma-international.org/ecma-262/6.0/#sec-get-o-p">Get</a>(O=<i>processorCtor</i>, P="parameterDescriptors")</code>.
 
-			1. Let <var>parameterDescriptorSequence</var>
-				be the result of
-				<a href="https://heycam.github.io/webidl/#es-to-sequence">
-				the conversion</a> from
-				<var>parameterDescriptorsValue</var>
-				to an IDL value of type
-				<code>sequence&lt;AudioParamDescriptor&gt;</code>.
+			1. If <var>parameterDescriptorsValue</var> is not <code>undefined</code>,
+				execute the following steps:
 
-			1. Let <var>paramNames</var> be an empty Array.
+				1. Let <var>parameterDescriptorSequence</var>
+					be the result of
+					<a href="https://heycam.github.io/webidl/#es-to-sequence">
+					the conversion</a> from
+					<var>parameterDescriptorsValue</var>
+					to an IDL value of type
+					<code>sequence&lt;AudioParamDescriptor&gt;</code>.
 
-			1. For each <var>descriptor</var> of
-				<var>parameterDescriptorSequence</var>:
+				1. Let <var>paramNames</var> be an empty Array.
 
-				1. Let <var>paramName</var> be the value of
-					the member {{AudioParamDescriptor/name}}
-					in <var>descriptor</var>. Throw
-					a {{NotSupportedError}} if
-					<var>paramNames</var> already
-					contains <var>paramName</var> value.
+				1. For each <var>descriptor</var> of
+					<var>parameterDescriptorSequence</var>:
 
-				1. Append <var>paramName</var> to
-					the <var>paramNames</var> array.
+					1. Let <var>paramName</var> be the value of
+						the member {{AudioParamDescriptor/name}}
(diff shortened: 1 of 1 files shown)


rtoy:
This looks good.  But you mention that the getter must return an iterable.  But I didn't see any change that makes this so. Is something missing?

padenot:
My patch doesn't say anything because it's already the case: https://heycam.github.io/webidl/#es-to-sequence throws if the parameter `parameterDescriptorsValue` is not iterable, and it's being just right after what I added.

rtoy:
Ok.  Your comment sounded like you were going to add some additional text.

LGTM.

index.bs
```diff
@@ -9847,49 +9847,52 @@ Methods</h5>
...
-					the member
-					{{AudioParamDescriptor/maxValue}}
-					in <var>descriptor</var>.
-
-				1. If <var>defaultValue</var> is less than
-					<var>minValue</var> or greater than
-					<var>maxValue</var>,
-					<span class="synchronous">throw a
-					{{InvalidStateError}}</span>.
+			1. If <var>parameterDescriptorsValue</var> is not <code>undefined</code>,
```
hoch: Can we not use double negative?

How about "exists"?  WebIDL seems to use this term when it checks the existence of things.

hoch:
Sorry for the late response, @padenot!

index.bs
karlt: `parameterDescriptorsValue` is a JS value and so WebIDL language is not helpful before conversion to a sequence, and `undefined` cannot be converted to a sequence.  ECMAScript has [`HasProperty()`](http://www.ecma-international.org/ecma-262/6.0/#sec-hasproperty), but that doesn't win us much because the property may be exist and still have the `undefined` value.

`undefined` is a JS value.  If that's what we choose to accept, then I assume we need to use that term.

padenot: Not using a `not` here forces to use a jump, so I prefer to keep it.

I've had another look at this:

1. If the property doesn't exist, it's fine, then the value is `undefined`, the `AudioWorkletProcessor` simply has no parameter (see [0])
2. If it exists and returns `undefined` then it is accepted, and the `AudioWorkletProcessor` similarly has no parameter (see [1])
3. If something else than `undefined` is returned, an attempt to convert it to an iterable happens. In case of success, the other checks are performed. Otherwise and exception is thrown as part of the normal iterable-to-sequence algorithm. 

There is a difference between 1. and 2. as Karl notes, but distinguishing here is not particularly helpful, is it?

[0]: http://www.ecma-international.org/ecma-262/6.0/#sec-get-o-p Assert 1 passes, assert 2 passes, http://www.ecma-international.org/ecma-262/6.0/#sec-ordinary-object-internal-methods-and-internal-slots-get-p-receiver is called, assert 1 still passes, then it goes 2, skip 3, goes into 4, and then depending, returns `undefined` now or later, depending on the prototype chain
[1]: http://www.ecma-international.org/ecma-262/6.0/#sec-get-o-p Assert 1 passes, assert 2 passes, http://www.ecma-international.org/ecma-262/6.0/#sec-ordinary-object-internal-methods-and-internal-slots-get-p-receiver is called, assert 1 still passes, then it goes 2, skip 3, skip 4, skip 5, execute 6, skip 7, execute 8, this returns `undefined`.

hoch: Yes. It makes sense. Thanks for the explanation. LGTM

PR #2174, opened 2020-03-05
Check that `minValue` is smaller `maxValue` than  when checking  after conversion to a sequence

### LINKED ISSUE #2173: Check that `minValue` is smaller than `maxValue` when checking `parameterDescriptors`
Otherwise this allows unbounded for the default value.

padenot:
This fixes #2173.

***
<a href="https://pr-preview.s3.amazonaws.com/padenot/web-audio-api/pull/2174.html" title="Last updated on Mar 5, 2020, 2:21 PM UTC (c42e050)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/2174/d43886d...padenot:c42e050.html" title="Last updated on Mar 5, 2020, 2:21 PM UTC (c42e050)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -9885,6 +9885,11 @@ Methods</h5>
 					{{AudioParamDescriptor/maxValue}}
 					in <var>descriptor</var>.
 
+				1. If  <var>minValue</var> is greater than
+					<var>maxValue</var>,
+					<span class="synchronous">throw a
+					{{InvalidStateError}}</span>.
+
 				1. If <var>defaultValue</var> is less than
 					<var>minValue</var> or greater than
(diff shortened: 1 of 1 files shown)


karlt:
If `minValue` is greater than `maxValue`, then we'll already have either "`defaultValue` is less than `minValue` or greater than `maxValue`" (or both) below and an InvalidStateError anyway, so we don't need a comparison of `minValue` directly with `maxValue`.

padenot:
Yes I got it backward. The point is that it's an error to do this, it's best signaled early, when `get parameterDescriptors` is read.

PR #2219, opened 2020-07-15
Update more terminology in images and markdown documents

padenot:
A few follow-ups after #2217. I also made changes to `.travis.yml` and the repo config on github. There is still a mention of `master` in an URL in `webaudio-CR-transition.md` but it's a direct quote so it makes sense to leave it as is.


rtoy:
Does this mean that we should now be committing to the main branch instead of master?  Should we just delete the master branch now?  Will git know what to do then?

padenot:
GitHub will now always try to merge to `main`. Your local clone will get a `main` branch after `git fetch`. The branch name is just a convention, locally.

PR #2222, opened 2020-07-17
Fix #2169: Remove redundant constraints in AudioParamDescriptor

### LINKED ISSUE #2169: AudioParamDescriptor has member constraints that are redundant
https://webaudio.github.io/web-audio-api/#dictionary-audioparamdescriptor-members describes various constraints on its members.  For example, for `defaultValue` it says:

> If this value is out of the range of float data type or the range defined by minValue and maxValue, a NotSupportedError exception MUST be thrown

That first constraint is redundant with the fact that it's declared as `float` in the IDL: the algorithm at https://heycam.github.io/webidl/#es-to-float will already throw on out-of-range values, before any of this spec's logic happens.  So it can just be removed.

The second constraint is redundant with https://webaudio.github.io/web-audio-api/#dom-audioworkletglobalscope-registerprocessor step 9 (though I took a look at the Blink code and I don't see it being enforced either in their version of step 7 or in step 9; I might have missed it, of course).

Similar for the other constraints here, including the one on "name"; that's enforced by https://webaudio.github.io/web-audio-api/#dom-audioworkletglobalscope-registerprocessor step 9 too.

@padenot

rtoy:
The constraints in the members of the `AudioParamDescriptor` are removed.  We replace that with a link to the steps in `registerProcessor` algorithm that describes how to handle the descriptor.

Updated the algorithm to include a step to throw an error if `minValue` is greater than `maxValue`.

***
<a href="https://pr-preview.s3.amazonaws.com/rtoy/web-audio-api/pull/2222.html" title="Last updated on Jul 20, 2020, 2:28 PM UTC (d5c6a6a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/2222/ca76b67...rtoy:d5c6a6a.html" title="Last updated on Jul 20, 2020, 2:28 PM UTC (d5c6a6a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -9883,13 +9883,13 @@ Methods</h5>
 
 				1. Let <var>paramNames</var> be an empty Array.
 
-				1. For each <var>descriptor</var> of
+				1. <div id="steps-parameterDescriptorSequence"></div>
+					For each <var>descriptor</var> of
 					<var>parameterDescriptorSequence</var>:
-
 					1. Let <var>paramName</var> be the value of
 						the member {{AudioParamDescriptor/name}}
-						in <var>descriptor</var>. Throw
-						a {{NotSupportedError}} if
+						in <var>descriptor</var>. <span class="synchronous">Throw
+						a {{NotSupportedError}}</span> if
 						<var>paramNames</var> already
 						contains <var>paramName</var> value.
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -9911,10 +9911,15 @@ Methods</h5>
 						{{AudioParamDescriptor/maxValue}}
 						in <var>descriptor</var>.
 
+					1. If <var>minValue</var> is
+						greater than or equal <var>maxValue</var>,
+						<span class="synchronous">throw an
+						{{InvalidStateError}}</span>.
```
padenot, line 9917: Per https://github.com/WebAudio/web-audio-api/pull/2174#issuecomment-595461280 this is unnecessary, but it's fine to keep it.

rtoy: Good point; I forgot about that. I think it's much clearer to remove this and adjust the next step to say that if the relationship minValue <= defaultValue <= maxValue is false, throw an error.

PR #2223, opened 2020-07-17
Fix #1457:  Privacy for sample rates

### LINKED ISSUE #1457: [privacy] Exposing data to an origin: fingerprinting
>> Does this specification expose any other data to an origin that it
doesn’t currently have access to?

>> Yes. When giving various information on available AudioNodes, the Web
Audio API potentially exposes information on characteristic features
of the client (such as audio hardware sample-rate) to any page that
makes use of the AudioNode interface. Additionally, timing information
can be collected through the AnalyserNode or ScriptProcessorNode
interface. The information could subsequently be used to create a
fingerprint of the client.

from Tom Ritter:
https://lists.w3.org/Archives/Public/public-privacy/2017OctDec/0017.html

> If a UA wanted to report generic information for a user, effectively
lying about audio hardware sample-rate and the rest) so that every
user presented the same data to prevent fingerprinting - are you able
to recommend sensible defaults in the draft? Or what a UA should
consider when choosing sensible defaults?

rtoy:
Basically takes @cwilso's suggested [text](https://github.com/WebAudio/web-audio-api/issues/1457#issuecomment-640767256) with slight editorial modifications to handle privacy issues with sample rates being exposed.

***
<a href="https://pr-preview.s3.amazonaws.com/rtoy/web-audio-api/pull/2223.html" title="Last updated on Jul 24, 2020, 3:53 PM UTC (7a5b15b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/2223/ca76b67...rtoy:7a5b15b.html" title="Last updated on Jul 24, 2020, 3:53 PM UTC (7a5b15b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -12297,7 +12297,7 @@ to Consider</a>
 	1. 44.1 kHz and 48 kHz are allowed as default rates; the system will choose
 		between them for best applicability. (Obviously, if the audio device is
 		natively 44.1, 44.1 will be chosen, etc., but also the system may choose
-		the most "compatible" rat&mdash;e.g. if the system is natively 96kHz,
+		the most "compatible" rate&mdash;e.g. if the system is natively 96kHz,
 		48kHz would likely be chosen, not 44.1kHz.
 	1. The system should resample to one of those two rates for devices that are
 		natively at different rates, despite the fact that this may cause extra
@@ -12308,14 +12308,14 @@ to Consider</a>
 		affordance to force use of the native rate&mdash;e.g. by setting a flag in
 		the browser on the device. This setting would not be exposed in the API.
 	1. It is also expected behavior that a different rate could be explicitly
-		requested in the constructor for AudioContext (this is already in the
+		requested in the constructor for {{AudioContext}} (this is already in the
 		specification; it normally causes the audio rendering to be done at the
-		requested sampleRate, and then up- or down-scaled to the device output),
+		requested sampleRate, and then up- or down-sampled to the device output),
 		and if that rate is natively supported, the rendering could be passed
 		straight through. This would enable apps to render to higher rates without
 		user intervention (although it's not observable from Web Audio that the
 		audio output is not downsampled on output)&mdash;for example, if
-		MediaDevices capabilities were read (with user intervention) and indicated
+		{{MediaDevices}} capabilities were read (with user intervention) and indicated
 		a higher rate was supported.
 
 6. Does this specification enable new script execution/loading
--- a/index.bs
+++ b/index.bs
@@ -12291,6 +12291,33 @@ to Consider</a>
 	recording or music making environment). Excessive latency decreases usability and may be
 	an accessibility issue.
 
+    Fingerprining via the sample rate of the {{AudioContext}} is also possible.  We
+    recommend the following steps to be taken to minimize this:
+
+	1. 44.1 kHz and 48 kHz are allowed as default rates; the system will choose
+		between them for best applicability. (Obviously, if the audio device is
+		natively 44.1, 44.1 will be chosen, etc., but also the system may choose
+		the most "compatible" rat&mdash;e.g. if the system is natively 96kHz,
+		48kHz would likely be chosen, not 44.1kHz.
+	1. The system should resample to one of those two rates for devices that are
+		natively at different rates, despite the fact that this may cause extra
+		battery drain due to resampled audio. (Again, the system will choose the
+		most compatible rate&mdash;e.g. if the native system is 16kHz, it's
+		expected that 48kHz would be chosen.)
+	1. It is expected (though not mandated) that browsers would offer a user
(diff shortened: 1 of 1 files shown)


rtoy:
Not sure if this is the best place to put this, but at least it's there in the security section.

index.bs
```diff
@@ -12291,6 +12291,33 @@ to Consider</a>
 	recording or music making environment). Excessive latency decreases usability and may be
 	an accessibility issue.
 
+    Fingerprining via the sample rate of the {{AudioContext}} is also possible.  We
+    recommend the following steps to be taken to minimize this:
+
+	1. 44.1 kHz and 48 kHz are allowed as default rates; the system will choose
+		between them for best applicability. (Obviously, if the audio device is
+		natively 44.1, 44.1 will be chosen, etc., but also the system may choose
+		the most "compatible" rat&mdash;e.g. if the system is natively 96kHz,
```
padenot, line 12300: missing `e`.

index.bs
```diff
@@ -12291,6 +12291,33 @@ to Consider</a>
...
+		battery drain due to resampled audio. (Again, the system will choose the
+		most compatible rate&mdash;e.g. if the native system is 16kHz, it's
+		expected that 48kHz would be chosen.)
+	1. It is expected (though not mandated) that browsers would offer a user
+		affordance to force use of the native rate&mdash;e.g. by setting a flag in
+		the browser on the device. This setting would not be exposed in the API.
+	1. It is also expected behavior that a different rate could be explicitly
+		requested in the constructor for AudioContext (this is already in the
+		specification; it normally causes the audio rendering to be done at the
+		requested sampleRate, and then up- or down-scaled to the device output),
```
padenot, line 12313: `s/scaled/sampled/`

index.bs
```diff
@@ -12291,6 +12291,33 @@ to Consider</a>
...
+		and if that rate is natively supported, the rendering could be passed
+		straight through. This would enable apps to render to higher rates without
+		user intervention (although it's not observable from Web Audio that the
+		audio output is not downsampled on output)&mdash;for example, if
+		MediaDevices capabilities were read (with user intervention) and indicated
```
padenot, line 12318: `{{MediaDevices}}` should auto-link.

index.bs
```diff
@@ -12291,6 +12291,33 @@ to Consider</a>
...
+	1. The system should resample to one of those two rates for devices that are
+		natively at different rates, despite the fact that this may cause extra
+		battery drain due to resampled audio. (Again, the system will choose the
+		most compatible rate&mdash;e.g. if the native system is 16kHz, it's
+		expected that 48kHz would be chosen.)
+	1. It is expected (though not mandated) that browsers would offer a user
+		affordance to force use of the native rate&mdash;e.g. by setting a flag in
+		the browser on the device. This setting would not be exposed in the API.
+	1. It is also expected behavior that a different rate could be explicitly
+		requested in the constructor for AudioContext (this is already in the
```
padenot, line 12311: `{{AudioContext}}` should auto-link.

index.bs
```diff
@@ -12291,6 +12291,33 @@ to Consider</a>
 	recording or music making environment). Excessive latency decreases usability and may be
 	an accessibility issue.
 
+    Fingerprining via the sample rate of the {{AudioContext}} is also possible.  We
```
padenot, line 12294: If we're doing it for the rate, we should do it for `outputLatency` and `maxChannelCount`.

rtoy: `maxChannelCount` should clearly be fixed to 2.  But what about `outputLatency`?  Not sure what value to use for that.  10 ms?

padenot: Gecko does this https://searchfox.org/mozilla-central/source/dom/media/webaudio/AudioContext.cpp#557-566, those are common numbers gathered from looking at a bunch of machines, and only make sense for Gecko, I assume: it's not buffering at all and it's up to date, because we know what OS API/buffer sizes we use underneath, and what a good "fake" latency figure should be.

rtoy: I guess, based on your link, that we just want to say that the `outputLatency` needs to be handled but in a browser-specific way, allowing browsers to choose good default values for each supported OS as needed.

padenot: I agree, that's the best solution.

rtoy: Fingerprinting of `outputLatency` is covered in the previous paragraph: 
> Fingerprinting via latency is also possible; it might be possible to
> deduce this from {{baseLatency}} and {{outputLatency}}. Mitigation
> strategies include adding jitter (dithering) and quantization so
> that the exact skew is incorrectly reported.

So I don't think we need to add anything else about this in this PR.

Just need to add a sentence about maxChannelCount.

rtoy:
Please take another look.

index.bs
rtoy: maxChannelCount set to 2.

PR #2224, opened 2020-07-18
Fix #2061: Mandate anti-fingerprinting of audio outputs

### LINKED ISSUE #2061: Privacy re-review of latest changes
I [raised an issue for privacy re-review](https://github.com/w3cping/tracking-issues/issues/13) for updated CR of v1.

rtoy:
Mandates browsers take action against fingerprinting of audio outputs, but leave the approach unspecified.  We mention one possible approach is dithering the output of audio nodes as described in the referenced paper.

***
<a href="https://pr-preview.s3.amazonaws.com/rtoy/web-audio-api/pull/2224.html" title="Last updated on Jul 22, 2020, 2:37 PM UTC (39d44d7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/2224/ca76b67...rtoy:39d44d7.html" title="Last updated on Jul 22, 2020, 2:37 PM UTC (39d44d7)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -12269,7 +12269,7 @@ to Consider</a>
 
 	In practice however, this merely allows deduction of information
 	already readily available by easier means (User Agent string),
-	such as "this is Chrome running on an x86".  However, to reduce the
+	such as "this is browser X running on platform Y".  However, to reduce the
 	possibility of additional fingerprinting, we mandate browsers take
 	action to mitigate fingerprinting issues that might be possible from the
 	output of any node.
--- a/index.bs
+++ b/index.bs
@@ -12269,7 +12269,16 @@ to Consider</a>
 
 	In practice however, this merely allows deduction of information
 	already readily available by easier means (User Agent string),
-	such as "this is Chrome running on an x86".
+	such as "this is Chrome running on an x86".  However, to reduce the
+	possibility of additional fingerprinting, we mandate browsers take
+	action to mitigate fingerprinting issues.  The exact approach is not
+	prescribed, but one possible approach is described in the paper by
+	Pierre Laperdrix, Benoit Baudry, Vikas Mishra. <a
+	href="https://hal.inria.fr/hal-01527580/document">"FPRandom: Randomizing
+	core browser objects to break advanced device fingerprinting
+	techniques"</a>, ESSoS 2017 - 9th International Symposium on Engineering
+	Secure Software and Systems, Jul 2017, Bonn,
+	Germany. pp.17. hal-01527580.
 
 	Fingerprinting via clock skew
 	<a href="http://sec.cs.ucl.ac.uk/users/smurdoch/talks/eurobsdcon07hotornot.pdf">has
--- a/index.bs
+++ b/index.bs
@@ -12271,14 +12271,8 @@ to Consider</a>
 	already readily available by easier means (User Agent string),
 	such as "this is Chrome running on an x86".  However, to reduce the
 	possibility of additional fingerprinting, we mandate browsers take
-	action to mitigate fingerprinting issues.  The exact approach is not
-	prescribed, but one possible approach is described in the paper by
-	Pierre Laperdrix, Benoit Baudry, Vikas Mishra. <a
-	href="https://hal.inria.fr/hal-01527580/document">"FPRandom: Randomizing
-	core browser objects to break advanced device fingerprinting
-	techniques"</a>, ESSoS 2017 - 9th International Symposium on Engineering
-	Secure Software and Systems, Jul 2017, Bonn,
-	Germany. pp.17. hal-01527580.
+	action to mitigate fingerprinting issues that might be possible from the
+	output of any node.
 
 	Fingerprinting via clock skew
 	<a href="http://sec.cs.ucl.ac.uk/users/smurdoch/talks/eurobsdcon07hotornot.pdf">has


index.bs
```diff
@@ -12269,7 +12269,16 @@ to Consider</a>
...
+	such as "this is Chrome running on an x86".  However, to reduce the
+	possibility of additional fingerprinting, we mandate browsers take
+	action to mitigate fingerprinting issues.  The exact approach is not
+	prescribed, but one possible approach is described in the paper by
+	Pierre Laperdrix, Benoit Baudry, Vikas Mishra. <a
+	href="https://hal.inria.fr/hal-01527580/document">"FPRandom: Randomizing
+	core browser objects to break advanced device fingerprinting
+	techniques"</a>, ESSoS 2017 - 9th International Symposium on Engineering
+	Secure Software and Systems, Jul 2017, Bonn,
+	Germany. pp.17. hal-01527580.
```
padenot, line 12281: I'll note that this approach, and the variant Brave ships, are both trivially defeated (this is not theoretical). I'd feel more comfortable not linking to this, because it doesn't work. The spec should not promote techniques that don't work.

rtoy: Removing the link is acceptable to me.

samuelweiler: @pes10k any comment?  @padenot, any pointers to work re: defeating farbling?

pes10k: I have no opinion about whether there is a link in the text. I especially do not want to give any impression of pushing for Brave's approach to be in the standard (i don't think think it should be).

I do think that "trivially defeated" (whether we agree or not in this particular case) are unrated on the web, since 1) making something "difficult but doable" is a deterrent in many cases, 2) making it less pleasant to do perf-wise really does deter some 1ps from doing it, 3) (prob most important) it helps put a bigger gap between benign and malicious API uses, which allows for other web-compat interventions.

rtoy: I note that if the referenced paper is removed, then the sentence in lines 12273-12274 basically implies all the other things that follow including handling the sample rate in PR #2223.  So I think we should either not do #2223 or not do this. Or refine the sentence to require a mitigation if a narrower scope.   (Then I'm not sure what to say here, though.)

padenot: As [previously discussed](https://github.com/w3cping/tracking-issues/issues/53#issuecomment-653541639), the fingerprinting offered by the Web Audio API and that we see in use today is just math. Essentially what is being fingerprinted is the implementation of the DSP algorithm of browsers and their environment, and not a particular the user or unique instance of a browser. 

What would be rational however is to encourage implementers to be careful at what math function are being used in a Web Audio API implementation and where they are from to avoid exposing the math library of the OS, that can (and does) change. This kind of prevent fingerprinting via the math library, which is kind of the same as finding the OS and its version, but some OSes can update their `libm` without a major version change, and so is probably not terribly useful.

This would only be useful if both:
- The math function that comes from the OS can be used to fingerprint
- The input of this function is controllable enough from the Web Audio API, where the values are restricted rather often, if controllable at all

In practice: Firefox (I tested different versions) on different computers with macOS 10.14 and macOS 10.15 and Chrome (I also tested different version) on different computers with the same OS pair return the same hash each time, regardless of the version of the "Web Audio Fingerprinting" I use, showing that this hash is function of the browser version.

More data point would be useful.

pes10k: @padenot i linked to some existing papers / data on web-audio fingerprinting here: https://github.com/WebAudio/web-audio-api/issues/2061#issuecomment-632372677

Specifically, 
> The most recent panopticlick data sees ~44 different web audio fingerprints; the Online Tracking: A 1-million-site Measurement and Analysis sees 713 different audio fingerprints during just the single study period

I dont know if thats compatible with the the "its just libm / implementation" conclusion (im not arguing it isn't right now) but is that the kind of data you're looking for?

padenot: This is known, yes. The kind of data I'm looking for is the opposite: get a bunch of machines in a controlled environment and check what makes the fingerprint differ. In my own testing, it's the browser, its version and the OS, and sometimes its version.

index.bs
```diff
@@ -12269,7 +12269,10 @@ to Consider</a>
 
 	In practice however, this merely allows deduction of information
 	already readily available by easier means (User Agent string),
-	such as "this is Chrome running on an x86".
+	such as "this is Chrome running on an x86".  However, to reduce the
```
padenot, line 12272: I'd prefer if we do not name products in an open specification.

> "this is browser X running on platform Y"

rtoy: I didn't actually change that, but I can change it to browser x on y.

index.bs
padenot: Also yes it's very much compatible at first glance, but it would be useful to have a repartition.

rtoy:
Teleconf:  Please respond by Aug 6.

pes10k:
@rtoy apologies if its obvious to others, but who are you waiting for responses from, and on what question(s)?

padenot:
This https://github.com/WebAudio/web-audio-api/pull/2224#discussion_r458618777

pes10k:
@padenot I've shared the data I have.  I'm not aware of any numbers on what you're asking for

padenot:
> The most recent panopticlick data sees ~44 different web audio fingerprints; the Online Tracking: A 1-million-site Measurement and Analysis sees 713 different audio fingerprints during just the single study period

The measured characteristics and what I expect to be common occurrences:

OS: macOS/Windows/Linux (and lots of variation there)/Android/iOS
Architecture: ARM (Windows/Linux/Android) x86 (macOS/Linux/Windows/Android)
Implementation: Chrome/Firefox/Safari/Edge/Nothing
Possible rate: 44100/48000/88200/96000/192000
Max Channel Count: mono/stereo/5.1/7.1/9.1

This quite a big parameter space. 713 fingerprints is very small compared to this parameter space, supporting the theory that this is what those fingerprinting methods are measuring, and that adding a noise floor doesn't help.

pes10k:
Howdy @padenot.

From reading through the "Online Tracking: A 1-million-site Measurement and Analysis", section 13 lists their methodology.  From what i can understand from the description of their implementation (and their [demo site](https://audiofingerprint.openwpm.com/)) I don't believe that channel count would increase the parameter space, since their implementation only considers the first channel in all cases.

So if we assume between 1-10 diff linux's (hard to say how many desktop linux users there are likely to be here w/o knowing more about the folks participating in the study), 2 archs, 4 different implementations, and 5 sample rates, that accounts for between 200 and 560 fingerprints; so that doesn't quite get to 713, but i see your point, that might be all thats going on here.

This also assumes that each of these parameters are independent, which is def not the case (no Safari on Arm, no Edge on MacOS or Linux, all iOS browsers will have the same Web Audio implementation, no Safari on Linux, etc), so the data isn't as dimensional as it might appear at first glance.

But maybe not enough to change the point.  Really it all hangs on how many different linux distros are in the mix, and I really have no guess at all about that; i could buy anything from none-to-a-lot, and don't know how to get a better answer than that. Your parameter space didn't include versions, which probably also add more space.

Anyway, TL;DR, It's still makes me *extremely* nervous that the folks who have the strongest commercial interest in tracking people seem to think this is a useful way to track people (the paper alone points to several independent tracking scripts that have unrelated implementations of web audio based fingerprinting), but maybe we're at the point that it doesn't matter. 

If the WG is certain that there is no fingerprinting surface here exposed beyond whats exposed by `Math.*` and similar in JS, then i'm fine to just note the concern in the spec and call it a day, and not trouble you all any further :)

padenot:
> If the WG is certain that there is no fingerprinting surface here exposed beyond whats exposed by `Math.*` and similar in JS, then i'm fine to just note the concern in the spec and call it a day, and not trouble you all any further :)

Because we don't have data for this, I'm doing it as I go, and noting the fingerprint of all the machines I have physically access to, where I can know the configuration, and more or less poke at it to understand what's up. For now, my theory is correct 100% of the time, but it's a slow process, and I don't have access to as many machines as usual because the Mozilla office is still closed.

pes10k:
Its not a fundamental solution or anything, but if you have a script or similar you're using, i'm happy to also pass it around Brave too, if more data points would be useful

rtoy:
Teleconf:  Land this and if there are still issues, file new ones.

padenot:
> Its not a fundamental solution or anything, but if you have a script or similar you're using, i'm happy to also pass it around Brave too, if more data points would be useful

Unfortunately this is very much a manual process.

Issue #2185, opened 2020-03-27
Many Bikeshed errors when building index.html (broken IDL?)


sideshowbarker:
I get many Bikeshed errors when building `index.html` from `index.bs` using `compile.sh`.

The errors seem mostly (or completely) related to IDL problems. I’m not sure if that’s because there’s a build step I’m missing, or what. I just run `compile.sh` directly:

```
$ ./compile.sh
+ BSLOG=bs.log
+ ERRLOG=actual-errs.txt
+ trap 'rm actual-errs.txt' 0
+ bikeshed --print=plain -f spec
+ tee bs.log
WARNING: Multiple elements have the same ID 'dom-decodeerrorcallback-error'.
Deduping, but this ID may not be stable across revisions.
WARNING: Multiple elements have the same ID 'dom-decodesuccesscallback-decodeddata'.
Deduping, but this ID may not be stable across revisions.
LINE 1961: Can't find the 'contextOptions' argument of method 'OfflineAudioContext/constructor(numberOfChannels, length, sampleRate)' in the argumentdef block.
LINE 3056: Can't find the 'destinationNode' argument of method 'AudioNode/connect(destinationParam, output)' in the argumentdef block.
LINE 3056: Can't find the 'input' argument of method 'AudioNode/connect(destinationParam, output)' in the argumentdef block.
LINE 3157: Can't find the 'destinationNode' argument of method 'AudioNode/disconnect(destinationParam, output)' in the argumentdef block.
LINE 3170: Can't find the 'destinationNode' argument of method 'AudioNode/disconnect(destinationParam, output)' in the argumentdef block.
LINE 3185: Can't find the 'destinationNode' argument of method 'AudioNode/disconnect(destinationParam, output)' in the argumentdef block.
LINE 3185: Can't find the 'input' argument of method 'AudioNode/disconnect(destinationParam, output)' in the argumentdef block.
LINE 10472: Can't find method '['AudioWorkletProcessor/process(inputs, outputs, parameters))']'.
LINE 1435: No 'method' refs found for 'constructor(contextOptions)'.
LINE 1961: No 'method' refs found for 'constructor(contextOptions)'.
LINE 1985: No 'method' refs found for 'constructor(numberOfChannels, length, sampleRate)'.
LINE 2346: No 'method' refs found for 'constructor(options)'.
LINE 4210: No 'method' refs found for 'constructor(context, options)'.
LINE 4672: No 'method' refs found for 'constructor(context, options)'.
[...]

PR #2236, opened 2020-08-03
Address #2185: Update expected errors

rtoy:
The latest veresion of bikeshed has fixed the errors printed for `<sequence<unsigned long>` and `record<DOMString, double>`, so we can remove these lines.


rtoy:
Just updates the expected results so I'm just going to land it.

PR #2245, opened 2020-08-26
Fix #2244: update implementation report

### LINKED ISSUE #2244: Update implementation report
**Describe the Issue**
The implementation report is outdated, having been updated last time in 2018.

**Expected Result**
The report should be recent.

rtoy:
Update the implementation report to show the current status of the implementation of the API.

I removed the section about the syntax-level DOM tests and replaced that with a link to the WebAudio IDL tests.

Also restored tidyconf.txt that we were using when the spec was respec so we can make the html report indented and wrapped nicely.

### DIFF
--- a/implementation-report.html
+++ b/implementation-report.html
@@ -6,33 +6,34 @@
       Web Audio 1.0 Implementation report
     </title>
     <style type="text/css">
-            a {
-                                text-decoration: none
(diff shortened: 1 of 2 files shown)


svgeesus (approved):
Great, LGTM

PR #2249, opened 2020-09-17
Fix #2227: Mention ABSN.start acquires the contents of the buffer

### LINKED ISSUE #2227: Clarify buffer.copyToChannel() must be called before source.buffer = buffer else nothing is played
**Describe the issue**

The current language in the specification does not make it clear that `buffer.copyToChannel()` must be called before `source.buffer = buffer` else audio will not play. 

Nightly 80 appears to implement that algorithm, where the only observable part of the algorithm employed is "nothing is played" even though `ended` event is dispatched.

At Chromium 86 audio is played when `buffer.copyToChannel()` is called after `source.buffer = buffer`. Though since the current language in the specification is not immediately clear if file an issue at Chromium bug tracker cannot now cite specific language relevant to the mandated order of calling `copyToChannel()` and setting `source.buffer = buffer`. 

Firefox bug https://bugzilla.mozilla.org/show_bug.cgi?id=1655336

**Where Is It**

1.4.3. Methods https://www.w3.org/TR/webaudio/#AudioBuffer-methods

> The acquire the contents of an AudioBuffer operation is invoked in the following cases:
> 
>     - When AudioBufferSourceNode.start is called, it acquires the contents of the node’s buffer. If the operation fails, nothing is played.
> 
>     - When the buffer of an AudioBufferSourceNode is set and AudioBufferSourceNode.start has been previously called, the setter acquires the content of the AudioBuffer. If the operation fails, nothing is played.
> 
>     - When a ConvolverNode's buffer is set to an AudioBuffer it acquires the content of the AudioBuffer.
> 
>     - When the dispatch of an AudioProcessingEvent completes, it acquires the contents of its outputBuffer.
> 
> Note: This means that copyToChannel() cannot be used to change the content of an AudioBuffer currently in use by an AudioNode that has acquired the content of an AudioBuffer since the AudioNode will continue to use the data previously acquired.

**Additional Information**

The language can be improved to state clearly what is intended to occur based on the necessary order of operations

> copyToChannel() MUST be called before source.buffer = buffer else nothing is played

which also means that in the case of `source.buffer = buffer` being executed before `buffer.copyToChannel()` "data" at

> AudioNode will continue to use the data previously acquired.

can be a `Float32Array` containing `0`s, effectively silence.

rtoy:
Just make it more explicit that calling `start()` will acquire the
contents of the buffer.  Without this, you kind of have to know that
this is described in the algorithm for acquiring the contents.

Also missed a minor issue in the start algorithm to abort the steps if
any exceptions are thrown because of parameter issues, as we say for
`AudioScheduledSourceNode.start()`.

***
<a href="https://pr-preview.s3.amazonaws.com/rtoy/web-audio-api/pull/2249.html" title="Last updated on Sep 18, 2020, 4:17 PM UTC (6644e1d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/2249/908dcc7...rtoy:6644e1d.html" title="Last updated on Sep 18, 2020, 4:17 PM UTC (6644e1d)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4808,13 +4808,19 @@ Methods</h4>
 				{{InvalidStateError}} exception MUST be thrown.</span>
 
 			2. Check for any errors that must be thrown due to parameter
-				constraints described below.
+				constraints described below.  If any
+				exception is thrown during this step,
+				abort those steps.
 
 			3. <a>Queue a control message</a> to start the
 				{{AudioBufferSourceNode}}, including the parameter values
 				in the messsage.
 
-			4. Send a <a>control message</a> to the associated {{AudioContext}} to
+			4. <a>Acquire the contents</a> of the
+			{{AudioBufferSourceNode/buffer}} if the
+			{{AudioBufferSourceNode/buffer}} has been set.
+
+			5. Send a <a>control message</a> to the associated {{AudioContext}} to
 				<a href="#context-resume">run it in the rendering thread</a> only when
 				the following conditions are met:
 				1. The context's <a>control thread state</a> is
--- a/index.bs
+++ b/index.bs
@@ -4817,8 +4817,8 @@ Methods</h4>
 				in the messsage.
 
 			4. <a>Acquire the contents</a> of the
-			{{AudioBufferSourceNode/buffer}} if the
-			{{AudioBufferSourceNode/buffer}} has been set.
+				{{AudioBufferSourceNode/buffer}} if the
+				{{AudioBufferSourceNode/buffer}} has been set.
 
 			5. Send a <a>control message</a> to the associated {{AudioContext}} to
 				<a href="#context-resume">run it in the rendering thread</a> only when


index.bs
```diff
@@ -4808,13 +4808,19 @@ Methods</h4>
...
+				abort those steps.
 
 			3. <a>Queue a control message</a> to start the
 				{{AudioBufferSourceNode}}, including the parameter values
 				in the messsage.
 
-			4. Send a <a>control message</a> to the associated {{AudioContext}} to
+			4. <a>Acquire the contents</a> of the
+			{{AudioBufferSourceNode/buffer}} if the
+			{{AudioBufferSourceNode/buffer}} has been set.
```
padenot, line 4821: You need to indent those two.

padenot (approved):
Good to merge once the build is fixed.

Issue #2047, opened 2019-08-25
AudioBufferSourceNode: Allow for (small) negative offsets in Start() when subsampling


collares:
This refers to the discussion at https://github.com/WebAudio/web-audio-api/issues/2032#issuecomment-524111689. In my opinion, @karlt makes a very good point in his comment: Due to subsampling, "interpolating" before the first sample provides (in some cases) meaningful audio content to be played. Consider the case, extracted from buffer-resampling.html, where the buffer's sample rate is 8000Hz and the context's sample rate is 48000Hz. Interpolating, even linearly, before the first sample would give us 5 non-silent samples which improve the audio quality in stitching cases. 

For backwards compatibility, the AudioBufferSourceNode can't start playing before its start time, so I am suggesting an opt-in alternative: Allowing for negative offsets (at least those falling between "buffer sample -1" (silence) and buffer sample 0) would have the desired outcome and address @karlt's audio quality concern. In this way, buffer stitching (as in sub-sample-buffer-stitching.html) could be achieved in a robust manner without linear extrapolation at the endpoints. The spec already allows for offsets a bit before after the last buffer sample (see #2032; this behavior is tested in buffer-resampling.html), so this would merely make the playback symmetric with regards to endpoints.

karlt:
I suspect there may be another way to address this.

> This also begs the question of what does `start(t)` then really mean if there is non-zero output before `t`?

I think that is a key part of the issue here.

Half the energy of the first sample in the buffer is before the time of the first sample and half is after, so centering the first buffer sample on time `t` is always placing half the energy of the first sample before the time `t`.

When the sample rate of the buffer matches that of the AudioContext and `t = n/sampleRate` for integer n, we don't notice any complications because we interpret the first ABSN output sample with its energy centered around the time `t`.  The differences in interpretation show up when "Resampling of the buffer may be performed arbitrarily by the UA at any desired point".

I've been assuming that the first sample in the buffer corresponds to buffer playhead position 0 and that the last sample corresponds to playhead position `buffer.duration - 1/buffer.sampleRate`.  However, it would be just as reasonable to assume that the first sample corresponds to playhead position `1/buffer.sampleRate` and that the last sample corresponds to position `buffer.duration`.  So the question arises whether symmetry would be better. i.e. first and last samples centred at `0.5/buffer.sampleRate` from each end.

[...]

PR #2256, opened 2020-10-02
Fix #2032:  ABSN extrapolates the last output

### LINKED ISSUE #2032: AudioBufferSourceNode: How to interpolate after last sample?
**Describe the issue**
The spec requires interpolation for playhead positions not corresponding to sampled times. Therefore, after each sample, there is an interval of duration 1/buffer.sampleRate corresponding to valid playhead positions which must be interpolated. This is also true for the interval after the last sample, but the spec doesn't specify how to produce the interpolated values for this region (since there is no next point to interpolate with).

This affects existing WPT tests: buffer-resampling.html seems to assume that the value of the last sample should be used for the whole interval, but another option is to interpolate with silence.

**Where Is It**
https://webaudio.github.io/web-audio-api/#playback-AudioBufferSourceNode -- more precisely, "Sub-sample start offsets or loop points may require additional interpolation between sample frames" does not cover the last interval described above, because it is not between sample frames. 

**Additional Information**
This is not related to looping, and in fact only applies when looping is disabled.

rtoy:
Update `playbackSignal` to mention that extrapolation is used to compute the output sample when the buffer is not looping and we're at the end of the buffer but need output a sample after the end of the buffer but before the next output sample frame.

***
<a href="https://pr-preview.s3.amazonaws.com/rtoy/web-audio-api/pull/2256.html" title="Last updated on Nov 6, 2020, 7:08 PM UTC (d911d4b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/2256/d50c68f...rtoy:d911d4b.html" title="Last updated on Nov 6, 2020, 7:08 PM UTC (d911d4b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5040,6 +5040,11 @@ function playbackSignal(position) {
 		algorithm that interpolates between sample frames in the neighborhood of
 		position.
 
+		If the buffer is not looping and if |position| is not on an exact sample
+		frame in the buffer and is past the end of the buffer (or specified
+		duration), extrapolation using the previous values in the buffer is used
+		to compute the output.  The extrapolation method is UA-specific.
+
 		If position is greater than or equal to loopEnd and there is no subsequent
 		sample frame in buffer, then interpolation should be based on the sequence
 		of subsequent frames beginning at loopStart.
--- a/index.bs
+++ b/index.bs
@@ -5056,11 +5056,11 @@ function playbackSignal(position) {
 		location of an exact sample frame in the buffer, this function returns
 		that frame. Otherwise, its return value is determined by a UA-supplied
 		algorithm that interpolates sample frames in the neighborhood of
-		position.
+		|position|.
 
-		If position is greater than or equal to loopEnd and there is no subsequent
+		If |position| is greater than or equal to |loopEnd| and there is no subsequent
 		sample frame in buffer, then interpolation should be based on the sequence
-		of subsequent frames beginning at loopStart.
+		of subsequent frames beginning at |loopStart|.
 	 */
 	 ...
 }
--- a/index.bs
+++ b/index.bs
@@ -5055,13 +5055,8 @@ function playbackSignal(position) {
 		values, one for each output channel. If |position| corresponds to the
 		location of an exact sample frame in the buffer, this function returns
 		that frame. Otherwise, its return value is determined by a UA-supplied
-		algorithm that interpolates between sample frames in the neighborhood of
+		algorithm that interpolates sample frames in the neighborhood of
 		position.
-
-		If the buffer is not looping and if |position| is not on an exact sample
-		frame in the buffer and is past the end of the buffer (or specified
-		duration), extrapolation using the previous values in the buffer is used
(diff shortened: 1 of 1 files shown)


padenot:
cc @karlt.

padenot:
This looks sensible, but @karlt knows more about this.

karlt:
I understand that @rtoy and I see this somewhat differently.
[sub-sample start timing](https://github.com/WebAudio/web-audio-api/issues/2032#issuecomment-708897898) and [upsampling](https://github.com/WebAudio/web-audio-api/issues/2047#issuecomment-525150837) are both situations where extrapolation would be just as useful before the first buffer sample as after the last sample, so it seems undesirable to me to describe extrapolation only after the last sample.

I would suggest just removing "between" from the text in the previous paragraph "UA-supplied algorithm that interpolates -between- sample frames in the neighborhood of position".  If "interpolates" is not considered generic enough to support values outside the provided buffer samples, then it can be replaced with "interpolates or extrapolates".  I would be happy that using "extrapolates" in this way would also include UA-specific sinc function approximations.

rtoy:
Sorry for the delay @karlt 
> I understand that @rtoy and I see this somewhat differently.
> [sub-sample start timing](https://github.com/WebAudio/web-audio-api/issues/2032#issuecomment-708897898) and [upsampling](https://github.com/WebAudio/web-audio-api/issues/2047#issuecomment-525150837) are both situations where extrapolation would be just as useful before the first buffer sample as after the last sample, so it seems undesirable to me to describe extrapolation only after the last sample.
> 
> I would suggest just removing "between" from the text in the previous paragraph "UA-supplied algorithm that interpolates -between- sample frames in the neighborhood of position". If "interpolates" is not considered generic enough to support values outside the provided buffer samples, then it can be replaced with "interpolates or extrapolates". I would be happy that using "extrapolates" in this way would also include UA-specific sinc function approximations.

I'm fine with removing "between" as you suggest.  I'm not sure what you're suggesting for the rest of the new text, though.

I don't think sinc interpolators are forbidden. It might be unexpected if users say `start(t)` and there's non-zero and possibly audible output before time `t`.  For all the DSP work I've done, we always used casual systems and had to account for any filter delays appropriately.

karlt:
> I'm fine with removing "between" as you suggest. I'm not sure what you're suggesting for the rest of the new text, though.

If "between" is removed, I'm suggesting that there is no need to add a new paragraph. i.e. modification to the preceding paragraph is sufficient to remove implication that interpolation (or extrapolation) happens only between samples.

rtoy:
> > I'm fine with removing "between" as you suggest. I'm not sure what you're suggesting for the rest of the new text, though.
> 
> If "between" is removed, I'm suggesting that there is no need to add a new paragraph. i.e. modification to the preceding paragraph is sufficient to remove implication that interpolation (or extrapolation) happens only between samples.

That's fine with me.  I've made the proposed changes.

rtoy:
@karlt says these changes are ok.  I'm going to land this now.

PR #2265, opened 2020-10-14
Fix #2252: Resolve [[pending resume promises]] when resuming

### LINKED ISSUE #2252: When are promises in "[[pending resume promises]]" resolved?
The [`[[pending resume promises]]` slot](https://webaudio.github.io/web-audio-api/#dom-audiocontext-pending-resume-promises-slot) gets filled with promises from early calls to `resume` (before context is allowed to start).

However, as far as I can tell, the spec never says when these promises are resolved.

The steps for [running a control message to resume an AudioContext](https://webaudio.github.io/web-audio-api/#dom-audiocontext-resume) either reject the promises or "clear" the list (why aren't the promises resolved first?). On top of that, I suspect that these promises need to be resolved when the context becomes [allowed to start](https://webaudio.github.io/web-audio-api/#allowed-to-start). In which case, this issue is similar to the comment I added to #2251: the spec does not explain what happens when context becomes allowed to start.

rtoy:
When running a control message to resume an AudioContext and we queue a task on the control thread's event loop, we want to resolve all pending resume promises, in addition to clearing them.

***
<a href="https://pr-preview.s3.amazonaws.com/rtoy/web-audio-api/pull/2265.html" title="Last updated on Oct 14, 2020, 8:28 PM UTC (15ff584)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/2265/47a04f5...rtoy:15ff584.html" title="Last updated on Oct 14, 2020, 8:28 PM UTC (15ff584)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1730,6 +1730,7 @@ Methods</h4>
 				execute these steps:
 
 
+				1. Resolve all promises from {{AudioContext/[[pending resume promises]]}} in order.
 				1. Clear {{AudioContext/[[pending resume promises]]}}. Additionally, remove those
 					promises from {{BaseAudioContext/[[pending promises]]}}.
 


rtoy:
I  think this all we need to say, but let me know if there's more to be added.

padenot (approved):

PR #2270, opened 2020-10-30
Update impl and test report 2020-10-30

rtoy:
Safari 115 has unprefixed WebAudio, so let's update the implementation report and test report to include these new results.

In the test report, I marked the tests for each interface that pass, but did not investigate why the browser fails a test (except for Chrome in some cases).


svgeesus (approved):
This looks good to me. Great to see substantial improvement in Safari. This makes a much stronger implementation report.

rtoy:
I don't plan on doing to many more updates, but I thought an updated Safari result would be helpful to show we actually have basically 3 different impls that are fairly complete.  If Safari adds worklet sometime soon, I'll update the results again then.

PR #2273, opened 2020-11-07
Clarify units used in threshold & knee value calculations

citelao:
I believe the current specification of the [compression curve](https://www.w3.org/TR/webaudio/#compression-curve) incorrectly asks implementers to add [linear gain unit](https://www.w3.org/TR/webaudio/#decibels-to-linear-gain-unit) `threshold` and `knee` values, while implementers should be adding the default, decibel versions.

This change updates the compression curve algorithm to define `knee end threshold` as the sum of the decibel `threshold` and `knee` values (*then* linearized) and uses that for the range calculations.

I believe this error came about because the units used in the compression curve algorithm were not well specified when #1568 updated the algorithm to be clearer.

## Why?

There are a few reasons I think that the sum should be done on the decibel versions:

* [threshold](https://www.w3.org/TR/webaudio/#dom-dynamicscompressornode-threshold) ranges from -100 to 0, while [knee](https://www.w3.org/TR/webaudio/#dom-dynamicscompressornode-knee) ranges from 0 to 40, suggesting that knee is meant to be a direct increase to the threshold
* 0 is meant to be a `hard-knee` compressor (effectively no knee). Adding `decibel threshold + 0` has that effect, whereas adding the linearized threshold to linearized knee of 0 always produces a number greater than 1, which I believe is invalid (db knee of 0 linearizes to 1).
* linearizing 40 decibels produces 100; I'm pretty sure linear gain units can only range from 0-1, so this would also be invalid.

I am not expert here, so I welcome feedback.

Thanks!

***
<a href="https://pr-preview.s3.amazonaws.com/citelao/web-audio-api/pull/2273.html" title="Last updated on Nov 30, 2020, 8:05 PM UTC (5940334)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/2273/d50c68f...citelao:5940334.html" title="Last updated on Nov 30, 2020, 8:05 PM UTC (5940334)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -7522,12 +7522,17 @@ of the same shape.
 		units=] and sampled at the time of processing of this
 		block (as [=k-rate=] parameters).
 		
-	1. Let <var>knee end threshold</var> have the value of
-		{{DynamicsCompressorNode/threshold}} plus
-		{{DynamicsCompressorNode/knee}}, also
+	1. Let <var>knee end threshold</var> have the value, also
 		[=decibels to linear gain unit|converted to linear
-		units=] and sampled at the time of processing of this
-		block (as [=k-rate=] parameters).
+		units=] after calculation, of
+		{{DynamicsCompressorNode/threshold}} plus
+		{{DynamicsCompressorNode/knee}} sampled at the time of
+		processing of this block (as [=k-rate=] parameters).
+		
+		Note: for clarity, the decibel values of
+		{{DynamicsCompressorNode/threshold}} and
+		{{DynamicsCompressorNode/knee}} should be added, not the
+		linear unit values.
 
 	1. Let <var>ratio</var> have the value of the
 		{{DynamicsCompressorNode/ratio}}, sampled at the time
--- a/index.bs
+++ b/index.bs
@@ -7521,18 +7521,14 @@ of the same shape.
 		[=decibels to linear gain unit|converted to linear
 		units=] and sampled at the time of processing of this
 		block (as [=k-rate=] parameters).
+	
+	1. Calculate the sum of {{DynamicsCompressorNode/threshold}} plus
+		{{DynamicsCompressorNode/knee}} also sampled at the time
+		of processing of this block (as [=k-rate=] parameters).
 		
-	1. Let <var>knee end threshold</var> have the value, also
-		[=decibels to linear gain unit|converted to linear
-		units=] after calculation, of
-		{{DynamicsCompressorNode/threshold}} plus
-		{{DynamicsCompressorNode/knee}} sampled at the time of
-		processing of this block (as [=k-rate=] parameters).
-		
-		Note: for clarity, the decibel values of
-		{{DynamicsCompressorNode/threshold}} and
-		{{DynamicsCompressorNode/knee}} should be added, not the
-		linear unit values.
+	1. Let <var>knee end threshold</var> have the value of this
+		sum [=decibels to linear gain unit|converted to linear
+		units=].
 
 	1. Let <var>ratio</var> have the value of the
 		{{DynamicsCompressorNode/ratio}}, sampled at the time
--- a/index.bs
+++ b/index.bs
@@ -7521,6 +7521,13 @@ of the same shape.
 		[=decibels to linear gain unit|converted to linear
 		units=] and sampled at the time of processing of this
 		block (as [=k-rate=] parameters).
+		
+	1. Let <var>knee end threshold</var> have the value of
+		{{DynamicsCompressorNode/threshold}} plus
+		{{DynamicsCompressorNode/knee}}, also
+		[=decibels to linear gain unit|converted to linear
+		units=] and sampled at the time of processing of this
+		block (as [=k-rate=] parameters).
 
 	1. Let <var>ratio</var> have the value of the
 		{{DynamicsCompressorNode/ratio}}, sampled at the time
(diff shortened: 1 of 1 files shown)


citelao:
Who would be the right person to work with to get this change in? Thanks!

rtoy:
We'll review this in the next working group/community group meeting this week.

Thanks for the patch!

citelao:
@rtoy Ah, makes sense. Thank you for reviewing! Happy to help.

rtoy:
> Is this something I should attend?

I think that's up to you.  If you're member of the webaudio slack channel, the info is posted there.  It's Thur, 9 am PST (San Francisco).

rtoy:
Teleconf:  Looks reasonable.  Someone will review this soon.

citelao:
> Teleconf: Looks reasonable. Someone will review this soon.

Thank you for looking into this! Sorry I didn't make the meeting.

index.bs
```diff
@@ -7521,6 +7521,18 @@ of the same shape.
 		[=decibels to linear gain unit|converted to linear
 		units=] and sampled at the time of processing of this
 		block (as [=k-rate=] parameters).
+		
+	1. Let <var>knee end threshold</var> have the value, also
+		[=decibels to linear gain unit|converted to linear
+		units=] after calculation, of
+		{{DynamicsCompressorNode/threshold}} plus
+		{{DynamicsCompressorNode/knee}} sampled at the time of
+		processing of this block (as [=k-rate=] parameters).
```
rtoy, line 7530: I think I'd break this into two steps:

1. Compute the sum of {{DynamicsCompressorNode/threshold}} and {{DynamicsCompressorNode/knee}}, summing the decibel values.
1. Let |knee end threshold} be the sum in the previous step, [=decibels to linear gain|converted to linear units=].

With this change, I think you can remove the Note below.

WDYT?

rtoy (commented):
Thanks for your patch!

Just a small suggestion on phrasing.

rtoy:

index.bs
citelao: I think that's a great clarification! It makes it much simpler to understand the steps. A "Note" is an indicator to me that I'm being unclear :).

I'll update the patch in the next few days! Thanks!

citelao:
@rtoy I'll be making the full edit soon, but 2 more issues popped up.

1. The example function in step 5 is non-continuous 
2. The units in step 5 of [the compression curve](https://www.w3.org/TR/webaudio/#compression-curve) are also unclear

## Example function in step 5 is non-continuous

The example function in step 5 is indeed linear, but it is not continuous with the other parts of the piecewise function and may provide a false impression to implementers (I made this mistake 😄):

> This function is linear, based on the ratio, after the threshold and the soft knee (i.e., 𝑓(𝑥)=1 / 𝑟𝑎𝑡𝑖𝑜⋅𝑥).

Chromium implements this as ([dynamics_compressor_kernel.cc](https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/platform/audio/dynamics_compressor_kernel.cc;l=117?q=dynamics_compressor_kern&ss=chromium)):

> (threshold + knee) + (1 / ratio) * (x - threshold - knee)

While the example function in the spec may simply be to illustrate linear functions, I believe it is confusing and it should be replaced with a more applicable function. Perhaps the above or:

> offset + (1 / ratio) * (x - offset)

(Or we should remind readers that the function must be monotonic & continuous at this point as well).

## Units in step 5 unclear

Also, the units in Step 5 are unclear:

> This function is linear, based on the ratio, after the threshold and the soft knee (i.e., 𝑓(𝑥)=1 / 𝑟𝑎𝑡𝑖𝑜⋅𝑥).

Since implementers have both linear and exponential units to choose from, the compression curve may behave exponentially differently depending on unit choice.

For example, choosing to apply the curve to the decibel value would be off by ~`10^(1/ratio)` versus using linear values.

## Next steps?

My plan is to:

* Fix the units for the threshold & knee value calculations (as originally intended)
* Either:
    * Update the example function or clarify monoticity & linearity **in this PR**
    * Make that change in a separate PR.
* Either:
    * Add the further clarification that the compressor function should be linear *with respect to decibel value* **in this PR**
    * Make that change in a separate PR.

However, I worry that the spec is vague about decibel vs linear values in other places.

Thoughts?    

---

Logic appears to apply the compressor curve to the decibel value:

Screenshot of Logic compressor, threshold -20db, ratio 2:1, graph shows 2:1 ratio for decibels over -20db

As does Chromium ([dynamics_compressor_kernel.cc](https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/platform/audio/dynamics_compressor_kernel.cc;l=117?q=dynamics_compressor_kern&ss=chromium)):

```c++
float y_db = yknee_threshold_db_ + slope_ * (x_db - knee_threshold_db_);
```

rtoy:
Thanks so much for looking into this more!

First, I would recommend looking at the [editor's draft](https://webaudio.github.io/web-audio-api/) instead of the candidate recommendation.

Second, I personally would prefer to fix this issue and file a new one about step 5 needing more work.  Your explanations make sense and should be done.

citelao:
Thanks @rtoy! Updated my PR and filed #2279.

Appreciate the guidance!

rtoy (approved):
This looks fine.  Thanks for fixing this.

I'd like @padenot to take a look too since he wrote most of this.

padenot (approved):
Looks better indeed, thanks for your work!

citelao:
Thank you! Happy to help :)

PR #2277, opened 2020-11-16
Fix compression curve diagram

padenot:
The knee comes after the threshold, and is added to it, so the soft knee
starts at the treshold.

### DIFF
--- a/images/compression-curve.svg
+++ b/images/compression-curve.svg
@@ -13,7 +13,7 @@
    viewBox="0 0 222.99424 225.08009"
    height="63.522602mm"
    width="62.933929mm"
-   inkscape:version="0.91 r13725"
+   inkscape:version="0.92.5 (2060ec1f9f, 2020-04-08)"
    sodipodi:docname="compression-curve.svg">
   <sodipodi:namedview
      pagecolor="#ffffff"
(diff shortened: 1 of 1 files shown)


padenot:
https://github.com/WebAudio/web-audio-api/pull/2277/files?short_path=7b72c9b#diff-7b72c9b801345c6cd8c520be21197c028322286e8ba7ca8bd0f55ae28afc26bb shows the difference.

rtoy (approved):
Looks good.  What did you use to create this image?  Should the source be in the repo?  We've have the source Graffle files for all of the images that used Graffle.

rtoy:
Not sure what's up with bikeshed.  Some intermittent data issue?

padenot:
> Looks good. What did you use to create this image? Should the source be in the repo? We've have the source Graffle files for all of the images that used Graffle.

I used Inkscape, an SVG editor that is free software. The source is the file itself :-).

Unclear what's up with bikeshed.

rtoy:
> > Looks good. What did you use to create this image? Should the source be in the repo? We've have the source Graffle files for all of the images that used Graffle.
> 
> I used Inkscape, an SVG editor that is free software. The source is the file itself :-).
Heh.  Ok.
> 
> Unclear what's up with bikeshed.
Yeah. I don't have this problem when I build locally.  I even updated my copy of bikeshed to the latest version just now.

padenot:
Same, but I also retriggered the travis build, and it still fails.

PR #2286, opened 2020-12-16
Fix #2248: Clean up ABSN start algorithm

### LINKED ISSUE #2248: ABSN.start sends a control message on a suspended context?
**Where Is It**
- https://webaudio.github.io/web-audio-api/#AudioBufferSourceNode-methods
- https://webaudio.github.io/web-audio-api/#dom-audioscheduledsourcenode-start

**Describe the issue**
In the algorithm for [`start()`](https://webaudio.github.io/web-audio-api/#dom-audioscheduledsourcenode-start), step 4.1 says the control message is sent to the rendering thread only if the control thread state is suspended.  And [control thread state](https://webaudio.github.io/web-audio-api/#control-thread-state) says the state corresponds to the states of an [AudioContext state](https://webaudio.github.io/web-audio-api/#dom-baseaudiocontext-state).

Do we really mean that the message is only sent if the context is suspended?  Should this not say running instead?  Or maybe the state isn't the context state? Or maybe it should say something altogether different?

rtoy:
Remove step 5; step 3 already queues a control message to start the buffer source so we don't need a step 5 to send another.  Plus step 5 sends another control message under certaiin conditions.  These conditions aren't useful since the start message should always be sent.  The context will handle them in the right way when it is resumed or already running.

***
<a href="https://pr-preview.s3.amazonaws.com/rtoy/web-audio-api/pull/2286.html" title="Last updated on Mar 8, 2021, 6:34 PM UTC (53c6ac8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/2286/0a52079...rtoy:53c6ac8.html" title="Last updated on Mar 8, 2021, 6:34 PM UTC (53c6ac8)">Diff</a>


rtoy:
My original change is wrong and I've updated it with more text and more changes including the following (as noted in the commit):

Make the algorithms ABSN.start and AudioScheduledSourceNode.start use the same text, more or less.  This means:

- Use the `[[source started]]` flag in ABSN, and add an additional  step in ABSN to set the flag, just like for ASSN.
- Rephrase the last step to mention more clearly we're sending a  message to start the context when possible.
- Add a NOTE to the ABSN step to state what this is meant to do
- Update the last step in ASSN to use the same algorithm as ABSN since  these sources are all supposed to start a context that is currently  suspended by autoplay

index.bs
```diff
@@ -4821,30 +4828,35 @@ Methods</h4>
 		<div algorithm="AudioBufferSourceNode.start()">
 			<span class="synchronous">When this method is called, execute these steps:</span>
 
-			1. <span class="synchronous">If <code>stop</code> has been called on this node, or if an
-				earlier call to <code>start</code> has already occurred, an
+			1. <span class="synchronous">If this {{AudioBufferSourceNode}} internal
+				slot {{AudioScheduledSourceNode/[[source started]]}} is true, an
```
hoch, line 4832: `<code>true</code>`

FYI I believe the `<code>` tag be replaced with a back tick.

index.bs
```diff
@@ -4138,7 +4138,14 @@ Methods</h4>
...
 			5. Send a <a>control message</a> to the associated {{AudioContext}} to
-				<a href="#context-resume">run it in the rendering thread</a>.
+				<a href="#context-resume">start running its rendering thread</a> only when
+				all the following conditions are met:
+				1. The context's <a>control thread state</a> is
+					"{{AudioContextState/suspended}}".
+				1. The context is <a>allowed to start</a>.
+				1. {{[[suspended by user]]}} flag is <code>false</code>.
+
+				NOTE: This allows {{AudioBufferSourceNode/start()}} to start an {{AudioContext}} that would otherwise not be <a>allowed to start</a>.
```
hoch, line 4148: nit: line wrap

index.bs
```diff
@@ -4821,30 +4828,35 @@ Methods</h4>
...
+				all the following conditions are met:
 				1. The context's <a>control thread state</a> is
-					<code>suspended</code>.
-				2. The context is <a>allowed to start</a>.
-				3. {{[[suspended by user]]}} flag is <code>false</code>.
+					{{AudioContextState/suspended}}.
+				1. The context is <a>allowed to start</a>.
+				1. {{[[suspended by user]]}} flag is <code>false</code>.
+
+				NOTE: This allows {{AudioBufferSourceNode/start()}} to start an {{AudioContext}} that would otherwise not be <a>allowed to start</a>.
```
hoch, line 4858: nit: line wrap

hoch (approved):
LGTM with nits

PR #2290, opened 2021-01-19
Fix #2291: Add script to create a zip file for a CR

### LINKED ISSUE #2291: Add script to create a zip file containing files needed for CR.
**Describe the Issue**
Creating a CR by hand requires knowing exactly what is needed.

**Expected Result**
Creating the files for CR should be easier. so a script would be nice to encapsulate all the knowledge needed.

rtoy:
This script compiles the spec and zips up all the files needed to
create a CR ready for publication.  It assumes you've updated index.bs
appropriately for CR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,9 @@ python:
 
 install:
   # Setup bikeshed. See https://tabatkins.github.io/bikeshed/#install-final
-  # - pip3 install bikeshed && bikeshed update
+  - pip3 install bikeshed && bikeshed update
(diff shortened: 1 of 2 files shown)


rtoy:
So we can make progress, I changed how CI gets bikeshed.  Instead of cloning the bikeshed repo, just use pip3 to install it like we used to.

padenot (approved):

padenot:
Ready to merge if this is complete.

rtoy:
I believe this has everything.  If not, we can fix it up as we go along.  Thanks for taking a look.

PR #2301, opened 2021-02-11
Define when and how regular tasks are processed wrt the processing model

### LINKED ISSUE #2008: Integration between event-loop of AudioWorkletGlobalScope and rendering loop
**Describe the issue**

Web Audio describes a control-thread, and a rendering-thread, where the control-thread runs a "traditional" event-loop as described in HTML, and the rendering thread is running a custom "rendering loop", described at https://webaudio.github.io/web-audio-api/#rendering-loop. 

Step 3 of the rendering loop, "Process a render quantum.", ends-up, in the case of a ` AudioWorkletNode`, calling into the `process` method of the corresponding ` AudioWorkletProcessor`. 

It should be noted that the  `AudioWorkletProcessor` runs in a `AudioWorkletGlobalScope`, which is a sub-class of `WorkletGlobalScope`, which should have it's own distinct event-loop as described in https://drafts.css-houdini.org/worklets/#the-event-loop

A further complication is that each `AudioWorkletProcessor`  also has a shipped `MessagePort`, whose `post-message-queue` should be treated as a first-class task-queue for the "event-loop" on which it happens to be, in this case the event-loop of the `AudioWorkletGlobalScope`. 

So the problem is that it's not clear how the rendering loop integrates with the event-loop of the `AudioWorkletGlobalScope`. 

> The rendering thread has another task queue for microtasks for any microtask operation such as resolution of Promises in the AudioWorkletGlobalScope.

This seems to imply a kind of integration where the rendering loop would run a microtask checkpoint for the `AudioWorkletGlobalScope`, as is found in Step 4 of the rendering loop. 

However that doesn't cover messages received on the port, since those aren't microtasks. 

If we take this example:

https://github.com/GoogleChromeLabs/web-audio-samples/blob/master/audio-worklet/design-pattern/shared-buffer/shared-buffer-worklet-processor.js

We see an interplay between the `onmessage` handler on the port, and the `process` method, where `process` essentially does nothing until an "initialization" message has been received on the port. 

In practice, how does a UA interleave running tasks on the `AudioWorkletGlobalScope`, for example when a new message is received on the port, with running the rendering loop which itself calls into `process` of the processor running in the context of `AudioWorkletGlobalScope`?

And what about tasks enqueued on the event-loop of `AudioWorkletGlobalScope` as a result of the call to https://drafts.css-houdini.org/worklets/#dom-worklet-addmodule ?

I can imagine two ways to do it in practice with the current spec:  

1. The task-queues from the event-loop of `AudioWorkletGlobalScope` could be plugged into the "control-message-queue" of the rendering thread? 
2. run the event-loop of `AudioWorkletGlobalScope` in parallel to the render-loop, but somehow "stop" it when the render-loop wants to call `process`?

It appears pretty clear that the goal is not running the event-loop of the  `AudioWorkletGlobalScope` fully in parallel to the render loop, since that would create potential race-conditions between the `onmessage` handler of a processor and it's `process` method. Yet I cannot find anything in the spec that integrates both event-loops so as to interleave both sequentially in some way. 

A solution could be rewording https://webaudio.github.io/web-audio-api/#rendering-loop, where step 4 would process all task-queues of the event-loop of the `AudioWorkletGlobalScope`, with microtask checkpoints interleaved?

**Where Is It**
https://webaudio.github.io/web-audio-api/#rendering-thread

**Additional Information**
Could be relevant for worklets in general, see https://github.com/whatwg/html/issues/4213

padenot:
A first cut for this, after reading lots of WHATWG text. Step 3.1 is necessary to avoid DOSing the render thread. I don't know if we want to be more precise for dispatching tasks, but considering the existence of an `implied event loop`, we're probably good.

This fixes #2008.

***
<a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/pull/2301.html" title="Last updated on Feb 19, 2021, 5:43 PM UTC (61d56a4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/2301/12411f1...61d56a4.html" title="Last updated on Feb 19, 2021, 5:43 PM UTC (61d56a4)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -11119,7 +11119,7 @@ in the algorithm of <a href="#rendering-a-graph">rendering a graph</a>.
 </div>
 
 The audio callback is also queued as a task in the <a
-	href="#control-message-queue">control message queue</a>. The UA MUST perform
+href="#control-message-queue">control message queue</a>. The UA MUST perform
 the following algorithms to process render quanta to fulfill such task by
 filling up the requested buffer size. Along with the <a>control message
 queue</a>, each {{AudioContext}} has a regular <a
--- a/index.bs
+++ b/index.bs
@@ -11118,17 +11118,20 @@ in the algorithm of <a href="#rendering-a-graph">rendering a graph</a>.
 	finished.
 </div>
 
-The audio callback is also queued as a task in the
-<a href="#control-message-queue">control message queue</a>. The UA MUST perform
+The audio callback is also queued as a task in the <a
+	href="#control-message-queue">control message queue</a>. The UA MUST perform
 the following algorithms to process render quanta to fulfill such task by
-filling up the requested buffer size. Along with the primary message queue, each
-{{AudioContext}} has a regular task queue for tasks that are posted to
-the rendering thread from the control thread. An additional microtask checkpoint
-is performed after processing a render quantum to run any microtasks that might
-have been queued during the execution of the `process` methods of
-{{AudioWorkletProcessor}}.
+filling up the requested buffer size. Along with the <a>control message
+queue</a>, each {{AudioContext}} has a regular <a
+href="https://html.spec.whatwg.org/multipage/webappapis.html#task-queue">task
+queue</a>, called its <dfn for="BaseAudioContext">associated task queue</dfn>
+for tasks that are posted to the rendering thread from the control thread. An
+additional microtask checkpoint is performed after processing a render quantum
+to run any microtasks that might have been queued during the execution of the
+`process` methods of {{AudioWorkletProcessor}}.
 
-All tasks posted from an {{AudioWorkletNode}} are posted to this task queue.
+All tasks posted from an {{AudioWorkletNode}} are posted to the [=associated
+task queue=] of its associated {{BaseAudioContext}}.
 
 <div id="rendering-initialization" algorithm="initialize rendering loop">
 	The following step MUST be performed once before the rendering loop starts.
@@ -11139,7 +11142,7 @@ All tasks posted from an {{AudioWorkletNode}} are posted to this task queue.
 </div>
 
 <div id="rendering-a-graph" algorithm="rendering a graph">
-  The following sterp MUST be performed when rendering a single render quantum.
+  The following steps MUST be performed when rendering a render quantum.
 
 	1. Let <var>render result</var> be <code>false</code>.
 
@@ -11158,10 +11161,9 @@ All tasks posted from an {{AudioWorkletNode}} are posted to this task queue.
 			2. Remove the [=oldest message=] of <var>Q<sub>rendering</sub></var>.
 
 
-	3. Process the regular event queue.
+	3. Process the {{BaseAudioContext}}'s [=associated task queue=].
 
-		1. Let <var>task count</var> be the number of tasks in the task queue of
-			the task queue of the {{AudioContext}} being rendered.
+		1. Let <var>task count</var> be the number of tasks in the task queue of the {{AudioContext}} being rendered.
 
 		2. Spin the event loop until <var>task count</var> tasks have been executed.
 
--- a/index.bs
+++ b/index.bs
@@ -10085,6 +10085,9 @@ Every {{AudioWorkletProcessor}} has an associated <dfn>active source</dfn> flag,
 the node to be retained in memory and perform audio processing in
 the absence of any connected inputs.
 
+All tasks posted from an {{AudioWorkletNode}} are posted to the task queue of
+its associated {{BaseAudioContext}}.
+
 <xmp class="idl">
 [Exposed=Window]
 interface AudioParamMap {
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -11131,6 +11139,8 @@ operation such as resolution of {{Promise}}s in the {{AudioWorkletGlobalScope}}.
 </div>
 
 <div id="rendering-a-graph" algorithm="rendering a graph">
+  The following sterp MUST be performed when rendering a single render quantum.
```
hoch, line 11142: s/sterp/steps

Also I would remove `single`.

index.bs
```diff
@@ -11147,7 +11157,15 @@ operation such as resolution of {{Promise}}s in the {{AudioWorkletGlobalScope}}.
 
 			2. Remove the [=oldest message=] of <var>Q<sub>rendering</sub></var>.
 
-	3. Process a render quantum.
+
+	3. Process the regular event queue.
```
hoch, line 11161: `regular event queue` - I think it's better to define 3 different types of queues somewhere. Also it seems like we are using different terms interchangeably. (e.g. task queue, message queue, and event queue)

Let's make a section and stick to it.

index.bs
```diff
@@ -11147,7 +11157,15 @@ operation such as resolution of {{Promise}}s in the {{AudioWorkletGlobalScope}}.
...
+
+		1. Let <var>task count</var> be the number of tasks in the task queue of
+			the task queue of the {{AudioContext}} being rendered.
```
hoch, line 11164: `in the task queue of the task queue`

Typo?

index.bs
```diff
@@ -11147,7 +11157,15 @@ operation such as resolution of {{Promise}}s in the {{AudioWorkletGlobalScope}}.
...
+
+		2. Spin the event loop until <var>task count</var> tasks have been executed.
```
hoch, line 11166: Do we really need `task count`? Can we just say:

"Run the event loop until all queued tasks in _Q_ are executed."

(_Q_ can be defined somewhere)

hoch (commented):
Thank you so much for working on this! 👍

A general comment: We have 1) the main thread task queue, 2) context's control message queue, and 3) render thread's message queue. Probably it's better to clarify them in this context.

L.10088 ~ L.10090 seems like a great place to add a section for this clarification.

index.bs
```diff
@@ -11118,9 +11121,14 @@ in the algorithm of <a href="#rendering-a-graph">rendering a graph</a>.
...
-<a>rendering thread</a> has another task queue for microtasks for any microtask
-operation such as resolution of {{Promise}}s in the {{AudioWorkletGlobalScope}}.
+filling up the requested buffer size. Along with the primary message queue, each
+{{AudioContext}} has a regular task queue for tasks that are posted to
+the rendering thread from the control thread. An additional microtask checkpoint
+is performed after processing a render quantum to run any microtasks that might
+have been queued during the execution of the `process` methods of
+{{AudioWorkletProcessor}}.
+
+All tasks posted from an {{AudioWorkletNode}} are posted to this task queue.
```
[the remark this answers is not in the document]
padenot: No, those are different.

index.bs
padenot: I haven't found a way to define this more cleanly, this is unambiguous. In particular, I haven't found a way to define `Q`. I'm happy to hear any idea though.

index.bs
```diff
@@ -10085,6 +10085,9 @@ Every {{AudioWorkletProcessor}} has an associated <dfn>active source</dfn> flag,
 the node to be retained in memory and perform audio processing in
 the absence of any connected inputs.
 
+All tasks posted from an {{AudioWorkletNode}} are posted to the task queue of
```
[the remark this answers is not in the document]
padenot: No, it's the task queue of the AudioContext.

index.bs
```diff
@@ -11118,9 +11121,14 @@ in the algorithm of <a href="#rendering-a-graph">rendering a graph</a>.
 The audio callback is also queued as a task in the
 <a href="#control-message-queue">control message queue</a>. The UA MUST perform
 the following algorithms to process render quanta to fulfill such task by
-filling up the requested buffer size. Along with the primary message queue, the
-<a>rendering thread</a> has another task queue for microtasks for any microtask
-operation such as resolution of {{Promise}}s in the {{AudioWorkletGlobalScope}}.
+filling up the requested buffer size. Along with the primary message queue, each
+{{AudioContext}} has a regular task queue for tasks that are posted to
```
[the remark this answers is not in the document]
padenot: The message queue is a queue of messages, and a task queue is a queue of tasks.

index.bs
padenot: Sure. The terms are not used interchangeably however.

padenot (commented):
Here is a new iteration. I've tried to use linking to make it clear. I've also fixed the typos and other issues.

index.bs
```diff
@@ -11115,12 +11118,20 @@ in the algorithm of <a href="#rendering-a-graph">rendering a graph</a>.
 	finished.
 </div>
 
-The audio callback is also queued as a task in the
-<a href="#control-message-queue">control message queue</a>. The UA MUST perform
+The audio callback is also queued as a task in the <a
+	href="#control-message-queue">control message queue</a>. The UA MUST perform
```
hoch, line 11122: nit: do we need to indent here?

hoch (approved):
Thanks! The new iteration looks good to me.

index.bs
padenot: No idea. I'll check locally by running `bikeshed` and merge as is if needed, or fixed if not. Thanks for the quick re-review!

hoch: 👍

padenot: (`bikeshed` accepts it).

PR #2308, opened 2021-03-11
Fix #2251: Refine phrasing for "allowed to start"

### LINKED ISSUE #2251: Be explicit that initial transition to running needs to take [[suspended by user]] flag into account?
I just [reported an issue against Webkit](https://bugs.webkit.org/show_bug.cgi?id=216972) which, unless I missed something, does not take the `[[suspended by user]]` flag into account when it runs the initial transition to `running` (delayed in Webkit). Other browser engines seem to be handling this case correctly, meaning that the following code will successfully end up in a `suspended` state across browsers (except Safari):

```js
button.addEventListener('click', async () => {
  const AudioContext = window.AudioContext || window.webkitAudioContext;
  const audioContext = new AudioContext();
  await audioContext.suspend();
  const node = audioContext.createOscillator();
  // State is `suspended`
});
```

That said, the spec does not explicitly state that the initial transition to `running` should take the `[[suspended by user]]` flag into account:
https://webaudio.github.io/web-audio-api/#allowed-to-start

If you all agree that's the right behavior, perhaps it should?

I note that the spec is neither fully clear about the fact that this "initial transition" will happen, nor is it about what that transition will do. For instance, the text could perhaps state that when the transition happens, and provided that the `[[suspended by user]]` flag is not set, the user agent must "send a control message to start processing".

rtoy:
The phrasing using "delay" is confusing.  Replace it with "disallow" which reflects more precisely what happens.  When a context is not allowed to start, it isn't really "delayed"; it can only start under specific actions with a user-gesture.

***
<a href="https://pr-preview.s3.amazonaws.com/rtoy/web-audio-api/pull/2308.html" title="Last updated on Mar 11, 2021, 6:55 PM UTC (8d7af81)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/2308/d360708...rtoy:8d7af81.html" title="Last updated on Mar 11, 2021, 6:55 PM UTC (8d7af81)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1373,7 +1373,7 @@ interface AudioContext : BaseAudioContext {
 An {{AudioContext}} is said to be <dfn>allowed to start</dfn> if the user agent
 allows the context state to transition from "{{AudioContextState/suspended}}" to
 "{{AudioContextState/running}}". A user agent may disallow this initial transition,
-to allow it only when the {{AudioContext}}'s [=relevant global object=] has
+and to allow it only when the {{AudioContext}}'s [=relevant global object=] has
 [=sticky activation=].
 
 {{AudioContext}} has an internal slot:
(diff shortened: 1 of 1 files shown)


padenot:
Why not indeed.

hoch (approved):

Issue #2119, opened 2020-01-10
Lifetime of pending processor construction data and exceptions in instantiation of AudioWorkletProcessor when AudioWorkletProcessor() constructor is not invoked


karlt:
[The instantiation of AudioWorkletProcessor algorithm](https://webaudio.github.io/web-audio-api/#AudioWorkletProcessor-instantiation) sets the node reference and transferred port of the AudioWorkletGlobalScope's pending processor construction data.

Only the [AudioWorkletProcessor() constructor](https://webaudio.github.io/web-audio-api/#dom-audioworkletprocessor-audioworkletprocessor) arranges for this data to be cleared, but "Construct a callback function" in the [instantiation of AudioWorkletProcessor](https://webaudio.github.io/web-audio-api/#AudioWorkletProcessor-instantiation) does not necessarily invoke the [AudioWorkletProcessor() constructor](https://webaudio.github.io/web-audio-api/#dom-audioworkletprocessor-audioworkletprocessor).

I assume that steps to "Empty the pending processor construction data slot" and "queue a task to the control thread to fire an event named processorerror" should be moved from the AudioWorkletProcessor() constructor to the instantiation of AudioWorkletProcessor algorithm?

Some more details are in https://github.com/WebAudio/web-audio-api/issues/2096, but perhaps all the details there occluded the core of the issue.

guest271314:
If gather this issue correctly this means there is a significant difference in processing between

```
constructor() {
  this.port.onmessage = e => {
    // messages, data transferred immediately upon AudioWorkletNode()
  }
}
```

and 

```
constructor() {
  this.port.onmessage = this.handMessage.bind(this);
}
handleMessage(e) {
  // do stuff
}
```

as to when `process()` is contructed begins processing inputs, outputs?

PR #2049, opened 2019-08-27
Add "pending construction processor data" and relevant logics

hoch:
Fixes #2027 based on https://github.com/WebAudio/web-audio-api/issues/2027#issuecomment-524406667.

cc @bzbarsky

***
<a href="https://pr-preview.s3.amazonaws.com/hoch/web-audio-api/pull/2049.html" title="Last updated on Aug 28, 2019, 7:25 PM UTC (3d3c998)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/2049/d3d9451...hoch:3d3c998.html" title="Last updated on Aug 28, 2019, 7:25 PM UTC (3d3c998)">Diff</a>

PR #2318, opened 2021-04-13
Restore empty of pending processor construction data after successful…

karlt:
… initialization of AudioWorkletProcessor#port

https://github.com/WebAudio/web-audio-api/pull/2306/files moved two "Empty the [=pending processor construction data=] slot" steps from the AudioWorkletProcessor constructor into one step in "The instantiation of AudioWorkletProcessor" algorithm.

However, we need to keep one of the steps in the AudioWorkletProcessor constructor in order to keep the behavior of allowing only one AudioWorkletProcessor with the construction data, as intended in https://github.com/WebAudio/web-audio-api/pull/2049#pullrequestreview-280367894 and tested in https://github.com/web-platform-tests/wpt/pull/21454

Followup for https://github.com/WebAudio/web-audio-api/issues/2119

***

### :boom: Error: 500 Internal Server Error :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Apr 13, 2021, 8:22 AM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [HTML Diff Service](http://services.w3.org/htmldiff) - The HTML Diff Service is used to create HTML diffs of the spec changes suggested in a pull request.

:link: [Related URL](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fpr-preview.s3.amazonaws.com%2FWebAudio%2Fweb-audio-api%2Fpull%2F2318%2Fe29a1cd.html&doc2=https%3A%2F%2Fpr-preview.s3.amazonaws.com%2Fkarlt%2Fweb-audio-api%2Fpull%2F2318.html)

```
<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>500 Internal Server Error</title>
</head><body>
<h1>Internal Server Error</h1>
<p>The server encountered an internal error or
misconfiguration and was unable to complete
your request.</p>
<p>Please contact the server administrator at 
 sysreq@w3.org to inform them of the time this error occurred,
 and the actions you performed just before this error.</p>
<p>More information about this error may be available
in the server error log.</p>
</body></html>

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20WebAudio/web-audio-api%232318.)._
</details>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -10509,6 +10509,9 @@ Constructors</h5>
 			1. Set <var>processor</var>’s
 				{{AudioWorkletProcessor/port}}
 				to <var>deserializedPort</var>.
+
+			1. Empty the [=pending processor construction data=]
+				slot.
 		</div>
 </dl>
 


karlt:
Non-substantive as editorial in restoring previous text.

rtoy:
Adding @hoch to review since he made the original change.

hoch:
@karlt 

Please help me understand. My understanding is that the current algorithm always performs 1.32.2.3 step 9 performed no matter what, even when the step 8 throws an exception in the constructor callback. (At least that was my intention)

> allowing only one AudioWorkletProcessor with the construction data

Do you believe the current text allows multiple/concurrent constructor invocations?

karlt:
> My understanding is that the current algorithm always performs 1.32.2.3 step 9 performed no matter what, even when the step 8 throws an exception in the constructor callback. (At least that was my intention)

Yes, and that's all good and necessary, thanks, because the `AudioWorkletProcessor` constructor may or may not be invoked, but consider what may happen before step 9 is reached.

> Do you believe the current text allows multiple/concurrent constructor invocations?

Yes, there is an example in the test where `processorCtor` invoked in step 8 of the ["instantiation"](https://webaudio.github.io/web-audio-api/#AudioWorkletProcessor-instantiation) algorithm [attempts to construct](https://github.com/web-platform-tests/wpt/blob/master/webaudio/the-audio-api/the-audioworklet-interface/processors/construction-port-new-after-new.js#L3-L6) two `AudioWorkletProcessor` objects before returning to the instantiation algorithm.

hoch:
> Yes, there is an example in the test where processorCtor invoked in step 8 of the "instantiation" algorithm attempts to construct two AudioWorkletProcessor objects before returning to the instantiation algorithm.

Ah. That makes sense. Can you share the link to the test? I would like to take a look.

That said, this PR looks good to me.

hoch (approved):

karlt:
> Can you share the link to the test?

https://github.com/web-platform-tests/wpt/blob/master/webaudio/the-audio-api/the-audioworklet-interface/processors/construction-port-new-after-new.js#L3-L6

Thanks for your work.

hoch:
Yup, that test case confirms my guess. Thanks for catching this!

PR #2352, opened 2021-06-04
RangeError is thrown only for negative cancelTime

rtoy:
Fixes WebAudio/web-audio-api-v2#127 by removing the text that says a
`RangeError` is also thrown for non-finite values.  That's not
possible because the type of `cancelTime` is `double`, not
`unrestricted double`.  Non-finite values are handled at a higher
level before these methods even see the number.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -97,6 +97,8 @@ window.addEventListener("DOMContentLoaded", function () {
 	});
 });
 </script>
+<!-- Handles candidate corrections and additions -->
+<script src="fixup.js"></script>
 <style>
 @media (prefers-color-scheme: light) {
 :root {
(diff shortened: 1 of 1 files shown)


rtoy:
@svgeesus My guess on how a candidate correction would work.  I just copied fixup.js from https://github.com/w3c/tr-design/issues that is mentioned in http://fantasai.inkedblade.net/style/design/w3c-restyle/2020/readme.

I have no idea if that's the right thing to do or not.  Also, for the id, I used c127, where 127 is the issue number of the thing we're fixing.  

And finally, the toggle buttons change both the text in cancelAndHoldAtTime and cancelScheduledValues.  Perhaps that's not the best, even though it is one issue.

padenot:
The Web Audio API looks good, in any case.

svgeesus (commented):
Don't the changed lines need to have ins and del round them?

rtoy:
There are no insertions so no `<ins>`.  But there are deletions marked by `<del>`.  You have to scroll way over to the right to see it where we delete the phrase "or is not a finite number".

I see now that we don't need to include fixup.js in our repo.  We can just load it up from w3c.org as you did in the audio eq cookbook note.

rtoy:
Need #2363 to be merged first now since that PR has all the bits needed, including adding a change log entry

hoch:
TODO: needs rebase.

PR #2358, opened 2021-06-25
Use new Web IDL buffer primitives

domenic:
See https://github.com/heycam/webidl/pull/987. In particular, "get a reference to the bytes" is going away, replaced with "write".

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1145,17 +1145,14 @@ Methods</h4>
 
 			2. Let <var>promise</var> be a new Promise.
 
-			3. If the operation <a href="https://tc39.github.io/ecma262/#sec-isdetachedbuffer"><code>IsDetachedBuffer</code></a>
-				(described in [[!ECMASCRIPT]]) on {{BaseAudioContext/decodeAudioData(audioData, successCallback, errorCallback)/audioData!!argument}} is
-				<code>false</code>, execute the following steps:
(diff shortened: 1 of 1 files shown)


rtoy:
In addition to the failures shown by Travis CI, we can't land this as is since the spec if a Living Standard.  I think these changes need to be marked up as shown in https://fantasai.inkedblade.net/style/design/w3c-restyle/2020/readme#amendment.

I'd be more than happy to do this for you, if you like.  We need some experience in doing these things, and this change seems another relatively simple update. (Well except for the bikeshed failures; I'll have to look into that a bit.)

domenic:
Yes, I'd appreciate if someone else could take this over. It's a real shame that we're still producing dead standards though :(. I hope the editor's draft at least can be kept up to date so that implementers have a source of truth to look.

The Bikeshed failures are likely due to Web IDL not yet deploying its new version, due to Travis CI issues. They should go away when we get that figured out.

(Note that once that is figured out, all dead versions of the Web Audio standard will have broken links... I guess that's pretty typical for dead standards though.)

rtoy:
Sure, I can take over PR.

I believe the editors draft will have all of the annotations for the changes, so it should be up-to-date with all the changes.  @svgeesus will have to help us to figure out how to move the editor's draft to recommendation at some later date.  Not sure if that's been figured out yet.

padenot:
We can close this, this has been handled in #2363.

PR #2363, opened 2021-06-28
Fix #2361: Use new Web IDL buffer primitives

### LINKED ISSUE #2361: Use new Web IDL buffer primitives
**Describe the issue**

See heycam/webidl#987. In particular, "get a reference to the bytes" is going away, replaced with "write".

Also see #2358 for the changes that are needed.

rtoy:
Updates spec to use new Web IDL buffer primitives.  The changes are
marked as candidate corrections, but we use several correction
amendments to make it easy to see what the new and old text will be.
Using just one makes it super-hard to figure out what controls the
text, and the text sometimes doesn't always clearly show what's
deleted and what's inserted because the links are colored correctly,
and struck-through text doesn't strike through links.  (A bikeshed
issue?)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1145,17 +1145,14 @@ Methods</h4>
 
 			2. Let <var>promise</var> be a new Promise.
 
-			3. If the operation <a href="https://tc39.github.io/ecma262/#sec-isdetachedbuffer"><code>IsDetachedBuffer</code></a>
-				(described in [[!ECMASCRIPT]]) on {{BaseAudioContext/decodeAudioData(audioData, successCallback, errorCallback)/audioData!!argument}} is
-				<code>false</code>, execute the following steps:
+			3. If {{BaseAudioContext/decodeAudioData(audioData, successCallback, errorCallback)/audioData!!argument}}
+				is [=BufferSource/detached=], execute the following steps:
 
 				1. Append <var>promise</var> to {{BaseAudioContext/[[pending promises]]}}.
 
-				2. <a href="https://tc39.github.io/ecma262/#sec-detacharraybuffer">
-					Detach</a> the {{BaseAudioContext/decodeAudioData(audioData,
+				2. [=ArrayBuffer/Detach=] the {{BaseAudioContext/decodeAudioData(audioData,
 					successCallback, errorCallback)/audioData!!argument}} {{ArrayBuffer}}.
-					This operation is described in [[!ECMASCRIPT]]. If this operations
-					throws, jump to the step 3.
+					If this operations throws, jump to the step 3.
 
 				3. Queue a decoding operation to be performed on another thread.
 
@@ -2513,7 +2510,7 @@ Methods</h4>
 	: <dfn>getChannelData(channel)</dfn>
 	::
 		According to the rules described in <a href="#acquire-the-content">acquire the content</a>
-		either [=get a reference to the buffer source|get a reference to=]
+		either [=ArrayBufferView/write=] into
 		or [=get a copy of the buffer source|get a copy of=]
 		the bytes stored in {{[[internal data]]}} in a new
 		{{Float32Array}}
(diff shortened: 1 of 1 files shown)


rtoy:
Biekshed has been updated to include the WebIDL references that were previously failing in https://travis-ci.com/github/WebAudio/web-audio-api/builds/230909490.

However, the link for "get a reference to the buffer source" is now no longer part of WebIDL,  I think that means we should just remove the link for the `<del>` part since it doesn't work.

index.bs
```diff
@@ -2512,8 +2519,14 @@ Methods</h4>
...
 	::
+		<div class="correction" id="c2361-2">
+		  <span class="marker">Candidate Correction
+		  <a href="https://github.com/WebAudio/web-audio-api/issues/2361">Issue 2361-2.</a>
+		  </span>
+		  Use new Web IDL buffer primitives
+		</div>
 		According to the rules described in <a href="#acquire-the-content">acquire the content</a>
-		either [=get a reference to the buffer source|get a reference to=]
+		either <del cite=#c2361-2>get a reference to</del><ins cite=#c2361-2>[=ArrayBufferView/write=]</ins> into
```
rtoy, line 2529: The phrasing for the new part is a bit confusing to me.  The new part says

> According to the rules described in acquire the content either write into...

But we're not writing anything at this point.  `getChannelData` allows you to get a copy or write at some later time; we're not writing any data now.

Is this the phrasing we really want to use?

@domenic WDYT?

index.bs
```diff
@@ -12680,6 +12735,11 @@ class Vec3 {
 
 <h2 id="changes">
 Change Log</h2>
+<h3 id="candidate-changes-1">
+  Candidate Changes since Recommendation of 17 Jun 2021
+</h3>
+  * <a href="https://github.com/WebAudio/web-audio-api/pull/2361">PR 2361</a>: Use new Web IDL buffer primitives
```
rtoy, line 12741: This correction used 7 correction boxes to make it easy to see that something was changed.  Perhaps it would make sense to link to all 7 boxes here in the change log so there's an easy way to find all 7?

index.bs
domenic: Yeah, I'm not sure I understood this part of the spec in its original formulation either; it didn't provide an algorithmic series of steps that explain what's going on. I'd be interested in seeing what the implementations do as an algorithm, so that we could make this spec precise instead.

domenic: To be clear, the existing spec didn't really make sense either, as it was trying to apply the "get a reference to"/"get a copy to" algorithm to [[internal data]], which is not a BufferSource.

rtoy: Fair enough.  I'll try to rewrite it a little so it seems less awkward.

index.bs
```diff
@@ -1136,6 +1138,13 @@ Methods</h4>
 		is via its promise return value, the callback parameters are
 		provided for legacy reasons.
 
+		<div class="correction" id="c2361-1">
+		  <span class="marker">Candidate Correction
+		  <a href="https://github.com/WebAudio/web-audio-api/issues/2361">Issue 2361-1.</a>
+		  </span>
+		  Use new Web IDL buffer primitives
```
rtoy, line 1145: Since there are 7 related changes, it might be useful to add buttons to move to the next (and previous) corrections.  This clutters things up a bit, but makes it easier to visit all the related changes.  Or we can just expect the user to look at the changelog to find all the related corrections, where we can generate the list automatically.

rtoy:
See #2366 for an example of adding previous and next buttons in each amendment box so it's easy to navigate each of the related changes.  This won't show up in the preview.  You'll have to build it yourself to see.

rtoy:
Previous and Next buttons added.  The JS code is now inline, so the preview shows the buttons and also the generated list of amendments for the changelog.

I'd appreciate help in getting the JS code nicer.  This was just something I hacked together that worked for this case.

index.bs
```diff
@@ -97,6 +97,81 @@ window.addEventListener("DOMContentLoaded", function () {
 	});
 });
 </script>
+<script>
```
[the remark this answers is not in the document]
rtoy: amendment.js should be removed since I moved all of that into index.bs so that the preview works.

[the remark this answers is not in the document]
hoch: Understood.

index.bs
rtoy: Without the buttons, finding related changes is harder.  We do have the ChangeLog which has a list of the changes, but having to go back and forth to the change log makes looking at the changes harder.

See also https://github.com/WebAudio/web-audio-api/pull/2366#issuecomment-878378045


Not sure what you mean by "new practice" and what part of the change you're referring to.

rtoy:
Also, take a look at https://pr-preview.s3.amazonaws.com/rtoy/web-audio-api/pull/2367.html#c2361-1 where the amendment box is scoped to include all the changes for the change.  Works nicely for paragraphs, but for algorithms, I think the box needs to be around all the steps in the algorithm, so the box scope can be pretty big, as shown in 2361-1.

Alternatively, we could maybe just add amendment boxes in each step, since we're keeping track of related changes.

rtoy:
See https://pr-preview.s3.amazonaws.com/rtoy/web-audio-api/pull/2368.html#c2361-1 for an example where we restrict the scope of a change the the smallest reasonable block.  This is kind of nice when the changes occur in an algorithm block so we don't include the entire algorithm in the scope of the change block.

I find having the scope of the changed marked out being pretty nice, especially for algorithms since it can be hard to know where the change would end.

rtoy:
Oh, according to https://fantasai.inkedblade.net/style/design/w3c-restyle/2020/readme#proposed-changes, the changes should be marked as proposed corrections/additions.

rtoy:
Merged in the proposals from #2367 and #2368 into this.  Now the boxes include the changed text area and each box covers a smaller change, such as a paragraph or one step in an algorithm.

Also updated the buttons to say "Previous Change" and "Next Change" to make it easier to understand the the previous/next buttons really did.

padenot (approved):

PR #2377, opened 2021-09-08
Migrate Travis CI to GitHub actions

hoch:
Sample link: https://hoch.github.io/web-audio-api/

The caveat is I had to remove index.html from `.gitignore` to make this work. Also we're not relying on w3c/spec-prod and that it will take more effort to maintain. That's why I still prefer fixing the spec errors and using w3c/spec-prod if possible.

What do you think?


hoch:
As we discussed on 9/9, the WG agreed on this approach.

@padenot Can you take a look and let me know what you think?

hoch:
@padenot friendly ping

padenot (approved):
Thanks!

PR #2379, opened 2021-09-18
fix: link to web audio api spec

guiseek:


hoch (approved):
Thank you!

padenot:
As discussed in the call last week, we can just merge this, it's in the README, not the spec text, no need for additional markup.

Issue #2450, opened 2019-07-19
Allow user-selectable render quantum size


rtoy:
**Describe the feature**
Allow an `AudioContext` and `OfflineAudioContext` to have a user-selectable render quantum size instead of the current fixed size of 128.

This allows better integration with AudioDeviceClient which can render in many different sizes.

**Is there a prototype?**
No.  Can't prototype this.

**Describe the feature in more detail**
Basically add a new dictionary member to specify a render size for the constructors.  The default is 128, of course.

juj:
How likely is it that this change will land?

Also, if this improvement lands, will the array length in AudioWorklets remain at fixed 128 samples, or will it reflect the quantum size of the context? Reading the PR suggests that it would, though I see the BitCrusher example was not changed: https://github.com/WebAudio/web-audio-api/pull/2469/files#r1117951767

Documentation like https://developer.chrome.com/blog/audio-worklet/#custom-audioparam would then benefit from updating.

rtoy:
FWIW, I had a bunch of CLs for Chrome that implemented this.  IIRC everything was working, except I had not yet handled anything having to do with FFTs.  Complicated stuff, but I think I could make it work since the FFT library supported non-powers of 2 FFTs.  I was going to restrict the set of buffer sizes to be lengths that were supported by the FFT.  Fortunately, this included buffer sizes like 160, 192, 240, etc. that are common on Android devices.

ElizabethHudnott:
Is this thread only about choosing a single render quantum size globally for the whole AudioContext? Or would it also be possible to have a subgraph that operated at some submultiple of the main quantum size, for example so that you could implement feedback FM outside of an audio worklet using a quantum size of just a couple of samples, as is possible in MaxMSP?

padenot:
It's for the entire AudioContext. Feedback FM or other DSP algorithms that use very short feedback loops are better implemented using `AudioWorkletProcessor`.

The goal of this is to align the render quantum to what the OS uses to maximize performance and also to potentially lower latency.

haywirez:
Has there been a discussion about considering adding support for changing this setting after the context is set up and running? It's a lot of added complexity, but I imagine there's a close relation to the [Render Capacity API](https://github.com/w3ctag/design-reviews/issues/843). It could be useful to change this without having to re-initialize a web app's entire audio graph, maybe while the context is suspended.

hoch:
2023 TPAC Audio WG Discussion:
[...]

PR #2462, opened 2021-12-01
Add AudioRenderCapacity interface and related classes

### LINKED ISSUE #2444: API proposal to preemptively determine audio thread overload (Render Capacity)
The Web Audio API defines audio production apps as a supported use-case, such as wave editors, digital audio workstations, and the like. And it's indeed quite adequate at the task, we at [AudioNodes](https://audionodes.com) had great success implementing it for a remarkably wide variety of features (even despite its painful limitation of not working in WebWorker threads).

One common trait of these applications is the capability of accepting virtually unlimited user content, which will essentially always result in hitting the limit of audio processing capabilities on a given machine at one point -- i.e. there is so many audio content that the **audio thread simply cannot keep up with processing** everything, it becomes overloaded, no matter how well it is optimized.

I believe this is widely and well understood in the audio production industry, and usually the solution to help the user avoid overloading is displaying a warning indicator of some kind, letting the user know the audio processing thread is about to be overloaded, and audio glitching will occur (so the user can know they should go easy on adding more contents).

_Note: in native apps (mostly C++ based), this is most commonly implemented as a CPU load meter (for the audio thread's core), which you can keep your eye on to know how far you are from the limit, roughly._

**Currently, the Web Audio API does not expose a comparable API to facilitate monitoring audio processing load, or overload.**

It's possible to _figure it out_, mainly in special cases with above-web-standard privileges (such as an Electron app). However, this is quite difficult to get right (even from native C++ side) without implementations taking a spec-defined standard into consideration.

I'd wish to propose a small set of light, straightforward, low-privacy implication API additions to enable this:

```js
audioContext.isOverloaded(); // 'true' or 'false'
```

```js
audioContext.addEventListener("overloadchange", function (e) {
    e.isOverloaded; // 'true' or 'false'
});
```

For obvious reasons, this is an extension to the `AudioContext` interface, not `BaseAudioContext`, as overload detection is not applicable for OfflineAudioContext processing. Having an event dedicated for the same purpose avoids the need of a polling check.

It is up to implementations to decide how _exactly_ it is determined whether the audio thread is considered overloaded or not, optionally taking the AudioContext `latencyHint` setting into consideration.

This would enable Web Audio API-based apps let the user know about high audio thread load, and display possible options or hints at steps for the user to take to avoid audio glitching.

**Privacy implications**

Exposing this information should have little to no privacy implications, as (1) it is rarely clear why exactly the audio thread is overloaded (it could be due to low device capabilities, or high CPU use by other processes), and (2) it does not provide a more accurate way to determine device capabilities than what is already possible with a simple scripted benchmark.

hoch:
Fixes #2444.

The PR is ready for rebase/merge.

***
<a href="https://pr-preview.s3.amazonaws.com/hoch/web-audio-api/pull/2462.html" title="Last updated on Dec 1, 2021, 10:55 PM UTC (38f1a04)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/2462/ae8388c...hoch:38f1a04.html" title="Last updated on Dec 1, 2021, 10:55 PM UTC (38f1a04)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/hoch/web-audio-api/pull/2462.html" title="Last updated on Apr 19, 2022, 5:32 PM UTC (ecf80f3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/2462/9ff540e...hoch:ecf80f3.html" title="Last updated on Apr 19, 2022, 5:32 PM UTC (ecf80f3)">Diff</a>


samuelweiler:
Please add an analysis of how much fingerprinting surface this adds to the spec's privacy considerations section.  If you think there's none, justify that claim.

hoch:
Thanks for chiming in, @samuelweiler.

- The real-time `AudioContext` is gated by the autoplay policy. Without an explicit user activation, the audio renderer does not run at all and thus this API will return 0 for all properties. (https://webaudio.github.io/web-audio-api/#allowed-to-start)
- `OfflineAudioContext` (which is quite popular for fingerprinting) does NOT have this API.
- This API metric only represents the computational load of the audio thread, and its site-specific characteristics can be widely different. (e.g. a web-based DAW and a Google doodle)
- We prefer the event-based pattern over the polling pattern because it helps avoid drive-by/automatic fingerprinting.
100 buckets (i.e. 0.00 ~ 0.99) for `averageCapacity`, `maxCapacity` and exposing the number of buffer underruns are also sensible from the privacy perspective, and it satisfies the developer’s’ needs.

index.bs
```diff
@@ -2100,6 +2106,113 @@ Dictionary {{AudioTimestamp}} Members</h5>
...
+<dl dfn-type=attribute dfn-for="AudioRenderPerformanceEvent">
+	: <dfn>timestamp</dfn>
+	::
+		{{BaseAudioContext/currentTime}} of {{AudioContext}} when the
+		event was fired.
+	: <dfn>averageCapacity</dfn>
+	::
+		An average of the collected raw render capacity values over the
+		specified interval. The precision is limited to 1/100th.
+	: <dfn>maxCapacity</dfn>
```
padenot, line 2200: Similarly, `maxLoad`. We might be able to find a better name because `maxLoad` could be the maximum acceptable load or something. I'm not sure.

index.bs
```diff
@@ -2100,6 +2106,113 @@ Dictionary {{AudioTimestamp}} Members</h5>
...
+	::
+		A maximum value from the collected raw render capacity values
+		over the specified interval. The precision is limited to
+		1/100th.
+	: <dfn>underrunRatio</dfn>
+	::
+		A ratio between the number of buffer underruns and the number
+		of render quanta. Where \(u\) is the number of underruns per
```
padenot, line 2208: It's probably obvious to us, but it's probably best to spec this in terms of "collection period", which are period of `udpateInterval` seconds, that are contiguous, not overlapping (since we have discrete blocks under the hood).  Then this underrun ratio is a property of a particular collection period, and it's clear when to "reset" the data collection.

When the system-level block size is not a divisor of the `updateInterval` * `AudioContext.sampleRate`, then we should spec it that it's `floor`ed (because we don't have the information for the very last block: it hasn't been rendered yet. It's more complicated if there is resampling or buffering somewhere in the browser, I guess.

index.bs
```diff
@@ -1663,6 +1664,11 @@ Attributes</h4>
 		audio output device remains the same. If the audio output
 		device is changed the {{outputLatency}}
 		attribute value will be updated accordingly.
+		
+	: <dfn>renderPerformance</dfn>
+	::
+		Returns an {{AudioRenderPerformance}} instance associated with
```
padenot, line 1670: `s/an/the/` ?

index.bs
```diff
@@ -2100,6 +2106,113 @@ Dictionary {{AudioTimestamp}} Members</h5>
...
+</pre>
+
+<h5 id="AudioRenderPerformanceEvent-attributes">
+Attributes</h5>
+
+<dl dfn-type=attribute dfn-for="AudioRenderPerformanceEvent">
+	: <dfn>timestamp</dfn>
+	::
+		{{BaseAudioContext/currentTime}} of {{AudioContext}} when the
+		event was fired.
```
padenot, line 2195: I think we want to spec this as not being the time the event was fired, but the start (or end) of the collection period.

To fire and even from the render thread, the following prose needs to be used:

> Queue a task to fire a simple event name "update" at the `AudioRenderPerformance`

This means that the render thread dispatches something to the control thread, and then the event is fired. There can be a time period in between the dispatch from the render thread to the control thread and the event handler being executed.

padenot (commented):
How does this work when the buffer size of the device is not aligned on the render quantum size? In its current shape, this proposal only measure the time it took to render a block, and the duration of this block.

It seems more useful to measure the time it took to render a system-level audio callback divided by the buffer size of this system-level audio callback, divided by the sample-rate of this system-level audio stream. Or maybe I'm reading too much when it's written "render quanta", and it's not the same as our "render quanta" in the spec, in which case we probably want to use another term. That may be what "raw capacity value" means.

An example to make it clear what I mean:

In the (common) case of a buffer size that is 192 frames at 48000 Hz (precisely 4 ms), if we have a graph that has a render quantum of 128 frames / 2.66ms (the default in all implementations for now) and renders in an average of 2ms, it's `ceil(2 / 2.66 * 100) =  76`.

In reality, every other callback, [256 frames of audio need to be rendered](https://github.com/WebAudio/web-audio-api/issues/2450#issuecomment-572469654), so about 5.32ms (excluding non-linearity effects, I think the point still stands), which underruns. So we're in a situation where the load is lower than 100, but underruns are clearly happening.

index.bs

index.bs
```diff
@@ -2100,6 +2106,113 @@ Dictionary {{AudioTimestamp}} Members</h5>
...
+
+<dl dfn-type=dict-member dfn-for="AudioRenderPerformanceOptions">
+	: <dfn>updateInterval</dfn>
+	::
+		An interval (in second) that capacity/underrun values are
+		updated. A render capacity value is calculated per render
+		quantum, and each value will be collected over the specified
+		interval period. For example, if the renderer runs at a 48Khz
+		sample rate with 128 frames of render quantum over 1 second
+		interval, 375 capacity values will be collected.
```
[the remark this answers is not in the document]
hoch: Good point. Done.

index.bs
```diff
@@ -2100,6 +2106,113 @@ Dictionary {{AudioTimestamp}} Members</h5>
...
+
+<h5 id="AudioRenderPerformanceEvent-attributes">
+Attributes</h5>
+
+<dl dfn-type=attribute dfn-for="AudioRenderPerformanceEvent">
+	: <dfn>timestamp</dfn>
+	::
+		{{BaseAudioContext/currentTime}} of {{AudioContext}} when the
+		event was fired.
+	: <dfn>averageCapacity</dfn>
```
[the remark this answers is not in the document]
hoch: I like the suggestion.

How about keeping the name "capacity" in the interface name, and change the properties to use "Load"?

index.bs
hoch: Yes. The term "load" is appropriate for the properties.

index.bs
```diff
@@ -2100,6 +2106,113 @@ Dictionary {{AudioTimestamp}} Members</h5>
...
+	::
+		An average of the collected raw render capacity values over the
+		specified interval. The precision is limited to 1/100th.
+	: <dfn>maxCapacity</dfn>
+	::
+		A maximum value from the collected raw render capacity values
+		over the specified interval. The precision is limited to
+		1/100th.
```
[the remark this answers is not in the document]
hoch: Or can this be calculated/approximated in the user land? That'd be my preference.

index.bs
hoch: I might need a bit more clarifications here - how do you want to rephrase it? I liked the term "collection period", but do you want to use the term to explain this `underrun ratio` field? Or do we want to revise all the language above with this new term?

index.bs
hoch: I'd like to work on that change as a follow-up PR. This PR focuses on the new interface, and the next one will revise the relevant portion of the rendering loop algorithm. What do you think?

hoch:
My response to https://github.com/WebAudio/web-audio-api/pull/2462#pullrequestreview-861034880:

> How does this work when the buffer size of the device is not aligned on the render quantum size? In its current shape, this proposal only measure the time it took to render a block, and the duration of this block.
>
> It seems more useful to measure the time it took to render a system-level audio callback divided by the buffer size of this system-level audio callback, divided by the sample-rate of this system-level audio stream. Or maybe I'm reading too much when it's written "render quanta", and it's not the same as our "render quanta" in the spec, in which case we probably want to use another term. That may be what "raw capacity value" means.

I also noticed this problem - my approach was to avoid mentioning the underlying callback buffer size and how it is connected to the rendering algorithm. The reason is that the concept is not formally specified. (We only have 2 instances of "system audio callback", and we don't explain or define how it works)

Instead, the current language only cares about the render quantum and the user-specified time interval. We can perhaps clarify by introducing a new section that explains how the Web Audio renderer works with the underlying audio system, but 1) that alone can be a significant spec work and 2) there might be some UA-specific details in this. Not sure if we want this.

An example to make it clear what I mean:

> In the (common) case of a buffer size that is 192 frames at 48000 Hz (precisely 4 ms), if we have a graph that has a render quantum of 128 frames / 2.66ms (the default in all implementations for now) and renders in an average of 2ms, it's ceil(2 / 2.66 * 100) = 76.

Yes, but do 192 frames matter? The render quantum and a load value is enough information.

> In reality, every other callback, 256 frames of audio need to be rendered, so about 5.32ms (excluding non-linearity effects, I think the point still stands), which underruns. So we're in a situation where the load is lower than 100, but underruns are clearly happening.

I am not following this point - why 256 frames are needed every other callback?

In general, I think we have an agreement on design - but perhaps we are not sure how to solve the uncertainty around the underlying callback, its interval, and how the "batch" rendering mechanism.

index.bs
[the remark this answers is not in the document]
padenot: How would you do this with this current proposed API, since there are no raw values (which means effectively there is a destructive low pass on the data)? Authors aim at a very small stddev in real-time system, because it allows maximizing the load at which it's safe to run the system, so it's a useful information.

index.bs
padenot: Sure, once we figure out what to measure, we can fix when the measurement happens.

index.bs
padenot: I think a collection period can be the number of frames of a system-level audio callback, divided by the sample-rate the system-level audio stream runs at. An underrun is when the time it took to render the audio for a particular callback exceeded this collection period (e.g. it took 5ms to render 4ms of audio). The underrun ratio is then the number of underrun over the number of collection periods.

padenot:
>I am not following this point - why 256 frames are needed every other callback?

I've tried to make it clear in the link, via a table: since the 128 is not a divisor of 192, when a callback arrives, with a time budget of 4ms (at 48kHz), 256 frames need to be rendered (to have 192 frames to give to the system callback), which means 64 frames are left in a buffer in the browser. Then the next callback comes, 192 frames are needed again, with the same time budget of 4ms. There are 64 frames left in the buffer, so the web browser only has to render 128 frames (128 + 64 = 192) to service the system audio callback. It repeats like this for the lifetime of the `AudioContext`.

So every callback, with a 4ms budget, we have either 256 or 128 frames to render. This means that an effective load histogram _should_ be bimodal, but it isn't with the current proposed spec text. This is important because it signals to authors that there is a problem and that the maximum load they can have on this configuration is half the maximum theoretical load.

I think it's perfectly workable with the current API shape, though. But in general, I'm afraid that not taking into account the realities of the underlying system will yield an API that is less useful than it could be.

If we shift the API to collect data on the system-level audio callbacks instead of render quanta, it immediately becomes a lot more useful for the user. There is in fact no need for render quanta (in the Web Audio API sense of a block of, for now, 128 frames) to be mentioned.

index.bs
hoch: How about "peakLoad"?

hoch:
Conclusion from WG teleconf on 1/27/2022:

* Agreed that we need a new section describes how the underlying platform (or system) audio callback drives the WebAudio renderer. We can take advantage of this new section to describe the render capacity interface and also the user-selectable render quantum feature.
* Adding the new section above does not block this PR. Revise the current PR based on the new concept now and we can add the new section afterwards.
* Revise the calculation process based on the new section. Some new terms need to be defined:
  * Collection period: the number of frames of a system-level audio callback, divided by the system sample-rate.
  * Underrun ratio: the number of underruns over the number of collection periods.

index.bs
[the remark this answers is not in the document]
hoch: I rephrased the text based on our discussion; now it explains in terms of the system-level audio callback instead of render quanta.

index.bs
hoch: There still is a concern about the system-level block size, but we'll get to it once we can agree on the general direction of the revision.

index.bs
hoch: One problem is that our rendering loop algorithm doesn't recognize the system-level audio callback. Firing event from the rendering loop can be addressed after we revise the render loop algorithm.

hoch:
@padenot @svgeesus Do we need a TAG review for this?

svgeesus:
> Do we need a TAG review for this?

Need, no. 

However, especially if there is an explainer for this, requesting review would be helpful at this early stage. Or we could wait until the Recommendation is republished with the next batch of new features.

index.bs
hoch: I updated the text, so please take a look again.

index.bs
hoch: This is addressed. Please take a look at the new revision.

padenot (approved):
Looks good, thanks!

index.bs
```diff
@@ -1510,23 +1511,36 @@ enum AudioContextLatencyCategory {
...
+<div class="correction proposed" id="c2444-1">
+	<span class="marker">Proposed Addition
+		<a href="https://github.com/WebAudio/web-audio-api/issues/2444">Issue
+			2444</a>.
+	</span>
+	Add AudioRenderCapacity Interface
+	<div class="amendment-buttons">
+		Buttons here
+	</div>
+	<ins cite=#c2444>
```
hoch, line 1523: @padenot We have a technical problem here - it seems like having `ins` and `del` doesn't work within the scope of `xmp`. This was the only way to make it pass the build check. Please advise if you have a better idea to handle this.

hoch:
@samuelweiler Please take a look at https://github.com/WebAudio/web-audio-api/pull/2462#issuecomment-1011333801 before we merge this change.

index.bs
padenot: This is https://github.com/tabatkins/bikeshed/issues/2109, filed by Raymond a while back. Suggested workaround is https://github.com/tabatkins/bikeshed/issues/2109#issuecomment-931612851.

hoch: Hmm. I actually tried that method in previous commit, but then bikeshed complained about duplicate IDL definition. I'll try again.

hoch: Huh, I used `data-no-idl` to disable IDL parsing on the previous IDL definition. It worked. :)

tidoust: > Huh, I used `data-no-idl` to disable IDL parsing on the previous IDL definition. It worked. :)

... but Bikeshed still adds the block to the final IDL index, so duplication still exists for extraction tools that process the index. See #2486 for a possible fix.

PR #2469, opened 2022-02-08
Allow specifying a different render-quantum-size for an AudioContext or OfflineAudioContext

padenot:
This fixes #13.

Without the heavy markup for now, this will come later.

This largely implements what was discussed.

***
<a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/pull/2469.html" title="Last updated on Sep 21, 2023, 3:11 PM UTC (c3d6d5d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/2469/940d58d...c3d6d5d.html" title="Last updated on Sep 21, 2023, 3:11 PM UTC (c3d6d5d)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -766,10 +766,10 @@ initially empty ordered list of promises.
 Each {{BaseAudioContext}} has a unique
 <a href="https://html.spec.whatwg.org/multipage/media.html#media-element-event-task-source">
 media element event task source</a>.
-Additionally, a {{BaseAudioContext}} has two private slots <dfn attribute
+Additionally, a {{BaseAudioContext}} has two several private slots <dfn attribute
 for="BaseAudioContext">[[rendering thread state]]</dfn> and <dfn attribute
 for="BaseAudioContext">[[control thread state]]</dfn> that take values from
-{{AudioContextState}}, and that are both 
+{{AudioContextState}}, and that are both
 <div class="correction proposed" id="c2373-1">
   <span class="marker">Proposed Correction
     <a href="https://github.com/WebAudio/web-audio-api/issues/2373">Issue 2373</a>-1.
@@ -781,7 +781,8 @@ for="BaseAudioContext">[[control thread state]]</dfn> that take values from
   <del>initialy</del>
   <ins>initially</ins>
 </div>
-set to <code>"suspended"</code>.
+set to <code>"suspended"</code>, and a private slot <dfn attribute for="BaseAudioContext">[[render quantum size]]</dfn> that is an
+unsigned integer.
 
 <pre class="idl">
 enum AudioContextState {
@@ -818,6 +819,42 @@ enum AudioContextState {
 </table>
 </div>
 
+<pre class="idl">
+ enum AudioContextRenderSizeCategory {
+    "default",
+    "hardware"
+};
+</pre>
+
+<div class="enum-description">
+<table class="simple" dfn-for="AudioContextRenderSizeCategory" dfn-type="enum-value">
+    <thead>
+        <tr>
+            <th scope="col" colspan="2">
+                Enumeration description
+            </th>
+        </tr>
+    <tbody>
+        <tr>
+            <td>
+                "<dfn>default</dfn>"
+            <td>
+                The AudioContext's <a>render quantum size</a> is the default value of 128
+                frames.
+        <tr>
+            <td>
+                "<dfn>hardware</dfn>"
+            </td>
+            <td>
+                The User-Agent picks a <a>render quantum size</a> that is best for the
+                current configuration.
+
+                Note: This exposes information about the host and can be used for fingerprinting.
+            </td>
+        </tr>
+</table>
+</div>
+
 <xmp class="idl">
 callback DecodeErrorCallback = undefined (DOMException error);
 
@@ -830,6 +867,7 @@ interface BaseAudioContext : EventTarget {
     readonly attribute double currentTime;
     readonly attribute AudioListener listener;
     readonly attribute AudioContextState state;
+    readonly attribute unsigned long renderQuantumSize;
     [SameObject, SecureContext]
     readonly attribute AudioWorklet audioWorklet;
     attribute EventHandler onstatechange;
@@ -958,6 +996,12 @@ Attributes</h4>
     ::
         Describes the current state of the {{BaseAudioContext}}.  Getting this
         attribute returns the contents of the {{[[control thread state]]}} slot.
+
+
+    : <dfn>renderQuantumSize</dfn>
+    ::
+        Getting this attribute returns the value of {{BaseAudioContext/[[render
+        quantum size]]}} slot.
 </dl>
 
 <h4 id="BaseAudioContent-methods">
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2160,7 +2225,19 @@ Constructors</h4>
 			2. Set the {{[[rendering thread state]]}} for
 				<var>c</var> to <code>"suspended"</code>.
 
-			3. Construct an {{AudioDestinationNode}} with its
+			3. Determine the {{[[render quantum size]]}} for this
+				 {{OfflineAudioContext}}, based on the value of
+				 the {{OfflineAudioContextOptions/renderSizeHint}}:
+
+				 1. If it has the default value of <code>"default"</code> or
```
hoch, line 2232: Is there a case where 'default' and 'hardware' do not match?

index.bs
```diff
@@ -10171,6 +10257,11 @@ Attributes</h5>
 	: <dfn>sampleRate</dfn>
 	::
 		The sample rate of the associated {{BaseAudioContext}}.
+
+	: <dfn>renderQuantumSize</dfn>
+	::
+		The value of the private slot [[render quantum size]] of the associated
```
hoch, line 10263: "Returns the value of the private slot ..."

hoch (commented):
Looks great! I have a few comments.

hoch:
The direction of this work looks great, but we (@hoch and @padenot representing Chrome and FireFox respectively) agreed that we don't have capacity to finish the actual implementation of this feature in 2022. We believe bringing the spec change when we don't have any implementations support it will confuse developers, so we decided to put this PR on hold until we have a plan for the actual engineering work.

index.bs
```diff
@@ -10147,6 +10232,7 @@ interface AudioWorkletGlobalScope : WorkletGlobalScope {
 	readonly attribute unsigned long long currentFrame;
 	readonly attribute double currentTime;
 	readonly attribute float sampleRate;
+	readonly attribute unsigned long renderQuantumSize;
```
juj, line 10235: This seems to affect the AudioWorklet callback quantum size?

In the spec there is the BitCrusher example, which has a comment

```
// The bitDepth parameter array has 128 sample values.
```
that probably could also updated in this PR, if that will change?

padenot: It will indeed change the `AudioWorkletProcessor` callback buffer sizes. This allows having the best performance and lowest audio latency.

But by default, an `AudioContext` will always have 128-frames render quantum size, because we've found so much code hard-coding 128. If the `AudioContext` is however constructed with the right attributes, then the buffers might be different in size. We're simply doing `new AudioContext()` in the example, so we're guaranteed to have 128 frames here. But we can certainly rework this to not say `128 sample values`, that's probably better.

Also, in the same vein, we could add a non-normative note saying that hard-coding values when you have `buffer.length` is bad practice.

index.bs
```diff
@@ -761,7 +761,7 @@ initially empty ordered list of promises.
 Each {{BaseAudioContext}} has a unique
 <a href="https://html.spec.whatwg.org/multipage/media.html#media-element-event-task-source">
 media element event task source</a>.
-Additionally, a {{BaseAudioContext}} has two private slots <dfn attribute
+Additionally, a {{BaseAudioContext}} has several private slots <dfn attribute
```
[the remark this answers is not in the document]
hoch: Bringing this to your attention again - @padenot

index.bs
```diff
@@ -813,6 +814,41 @@ enum AudioContextState {
...
+				"<dfn>default</dfn>"
+			<td>
+				The AudioContext's <a>render quantum size</a> is the default value of 128
+				frames.
+		<tr>
+			<td>
+				"<dfn>hardware</dfn>"
+			</td>
+			<td>
+				The User-Agent picks a <a>render quantum size</a> that is best for the
```
hoch, line 844: I believe this value can be implementation specific. (possible privacy discussion)

index.bs
```diff
@@ -952,6 +989,11 @@ Attributes</h4>
 	::
 		Describes the current state of the {{BaseAudioContext}}.  Getting this
 		attribute returns the contents of the {{[[control thread state]]}} slot.
+
+	: <dfn>renderQuantumSize</dfn>
+	::
+		Getting this attribute returns the value of {{BaseAudioContext/[[render
```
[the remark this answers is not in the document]
hoch: ping @padenot

index.bs
```diff
@@ -1563,21 +1605,34 @@ Constructors</h4>
 			<span class="synchronous">When creating an {{AudioContext}},
 			execute these steps:</span>
 
-			1. Set a {{[[control thread state]]}} to <code>suspended</code> on the {{AudioContext}}.
+			1. Set a private slot called {{[[control thread state]]}} to <code>suspended</code> on the {{AudioContext}}.
```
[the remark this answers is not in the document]
hoch: ping @padenot

index.bs
```diff
@@ -1563,21 +1605,34 @@ Constructors</h4>
...
+
+			2. Set a private slot called {{[[rendering thread state]]}} to <code>suspended</code> on the {{AudioContext}}.
```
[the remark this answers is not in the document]
hoch: ping @padenot

index.bs
```diff
@@ -1563,21 +1605,34 @@ Constructors</h4>
...
 
-			2. Set a {{[[rendering thread state]]}} to <code>suspended</code> on the {{AudioContext}}.
+			4. Set a private slot called {{[[render quantum size]]}} to
```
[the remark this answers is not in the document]
hoch: ping @padenot

index.bs
```diff
@@ -2160,7 +2225,19 @@ Constructors</h4>
...
-			3. Construct an {{AudioDestinationNode}} with its
+			3. Determine the {{[[render quantum size]]}} for this
+				 {{OfflineAudioContext}}, based on the value of
+				 the {{OfflineAudioContextOptions/renderSizeHint}}:
+
+				 1. If it has the default value of <code>"default"</code> or
+					<code>"hardware</code>", set the {{[[render quantum size]]}} private
+					slot to 128.
+
+				 2. Else, if an integer has been passed, the User-Agent can decide to
```
hoch, line 2236: Don't we need the third step where the UA decides its own value because (2) failed?

index.bs
hoch, line 2232: I am a bit confused about the step (1). Does this mean that if `[render quantum size]` is equal to `default` or `hardware` then we use 128?

index.bs
hoch: @juj Does the response above answer your question?

hoch (approved):
LGTM with a few nits.

index.bs
juj: Yes it did, thanks.

index.bs
padenot: It's usually device specific -- but it's indeed something that can be used for fingerprinting. I'll add something.

index.bs
padenot: This is for `OfflineAudioContext`, not `AudioContext`.

index.bs
```diff
@@ -12182,9 +12450,25 @@ asynchronous sections can't be reordered.
...
+  <div class="amendment-buttons"></div>
+  <del>
 Rendering an audio graph is done in blocks of 128 samples-frames. A
-block of 128 samples-frames is called a <dfn>render quantum</dfn>, and
-the <dfn>render quantum size</dfn> is 128.
+block of 128 samples-frames is called a <a>render quantum</a>, and
+the <a>render quantum size</a> is 128.
+  </del>
+  <ins>
+Rendering an audio graph is done in blocks of samples-frames, the size of this
```
hoch, line 12465: To make it read better:

Audio graph rendering is done in blocks of sample-frames, with the size of each block remaining constant for the lifetime of a BaseAudioContext. The number of sample-frames in a block is called `render quantum size`, and the block itself is called a `render quantum`. Its default value is 128, and it can be configured by setting AudioContextOptions/renderSizeHint.

index.bs
```diff
@@ -2848,6 +3049,16 @@ Constructors</h4>
                     1. Set the {{[[rendering thread state]]}} for
                         |c| to <code>"suspended"</code>.
 
+                    1. Determine the {{[[render quantum size]]}} for this {{OfflineAudioContext}}, based on the value of the {{OfflineAudioContextOptions/renderSizeHint}}:
```
hoch, line 3052: nit: line wrapping for consistency

hoch (approved):
Still LGTM, but with two nits.

PR #2476, opened 2022-02-23
Explain that system-level audio callbacks exist.

### LINKED ISSUE #2475: Explain that system-level audio callbacks exist

padenot:
This fixes #2475.

@hoch, this helps #2462 somewhat. There is a bit of duplication but not much.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -176,6 +176,7 @@ window.addEventListener('load', (event) => {
   ListAmendments("c2373", "Proposed Correction", "change-list-2373");
   ListAmendments("c2321", "Proposed Correction", "change-list-2321");
   ListAmendments("c2375", "Proposed Correction", "change-list-2375");
+  ListAmendments("c2475", "Proposed Addition", "change-list-2475");
 });
 </script>
 <style>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -11391,15 +11392,56 @@ The algorithm for rendering a block of audio from a {{BaseAudioContext}}
...
+<div class="correction proposed" id="c2475-1">
+	<span class="marker">Proposed Addition
+		<a href="https://github.com/WebAudio/web-audio-api/issues/2475">Issue
+			2375</a>.
+	</span>
+	Add definition for various concepts related to system-level audio callbacks
+	<div class="amendment-buttons">
+		Buttons here
+	</div>
+	<ins cite=#c2375>
```
hoch, line 11404: Is 2375 correct? Perhaps this should be 2475?

2375 seems like a closed issue on MediaStreamTrack.

index.bs
```diff
@@ -11391,15 +11392,56 @@ The algorithm for rendering a block of audio from a {{BaseAudioContext}}
...
+			divisor of the <a>system-level audio callback buffer size</a>. This causes
+			increased audio latencies and reduced possible maximum load without
+			<a>audio buffer underrun</a>.
+		</div>
+
+		An {{OfflineAudioContext}} is not required to have a system-level audio
+		callback, but behaves as if it did with the callback happening as soon as
+		the previous callback is finished.
+	</ins>
+	<del cite=#c2375>
```
hoch, line 11433: Ditto on 2375.

hoch (commented):
Thanks for working on this, @padenot!

index.bs
```diff
@@ -11391,15 +11392,56 @@ The algorithm for rendering a block of audio from a {{BaseAudioContext}}
...
+	<ins cite=#c2375>
+		The {{AudioContext}} <a>rendering thread</a> is running off a
+		<dfn>system-level audio callback</dfn>, that executes in an isochronous
+		fashion. This <a>system-level audio callback</a> is called repeatidly, and
+		each call has a <dfn>system-level audio callback buffer size</dfn>
+		(potentially variable), which is the amount of audio that needs to be
+		computed on time before the <a>system-level audio callback</a> returns.
+
+		A <dfn>load value</dfn> is computed for each <a>system-level audio
+		callback</a>, by dividing the duration taken by the implementation to
```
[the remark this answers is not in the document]
padenot: I've rephrased this.

padenot:
All the comments addressed.

index.bs
```diff
@@ -11391,15 +11392,56 @@ The algorithm for rendering a block of audio from a {{BaseAudioContext}}
...
+	</span>
+	Add definition for various concepts related to system-level audio callbacks
+	<div class="amendment-buttons">
+		Buttons here
+	</div>
+	<ins cite=#c2475>
+		The {{AudioContext}} <a>rendering thread</a> is driven by a
+		<dfn>system-level audio callback</dfn>, that is periodically invoked in an
+		isochronous fashion. This <a>system-level audio callback</a> is called
+		repeatedly, and each call has a <dfn>system-level audio callback buffer
```
hoch, line 11408: > This <a>system-level audio callback</a> is called repeatedly, 

Seems redundant. The previous sentence says "periodically" and "an isochronous fashion". I also heard people saying "isochronous" is not a common term a dozen time. It's perfectly fine to keep it, but we also can try to make it more accessible to developers.

index.bs
```diff
@@ -11391,15 +11392,56 @@ The algorithm for rendering a block of audio from a {{BaseAudioContext}}
...
+		size</dfn> (that is potentially variable between calls), which is the
```
hoch, line 11409: | (that is potentially variable between calls)

Perhaps we can drop this and merge it to the next sentence?

hoch, line 11409: | (that is potentially variable between calls)

Perhaps we can drop this and merge it to the next sentence?

index.bs
```diff
@@ -11391,15 +11392,56 @@ The algorithm for rendering a block of audio from a {{BaseAudioContext}}
...
+		number of sample-frames that needs to be computed on time before the
```
hoch, line 11410: "the number of sample frames" -> "a varying number of sample-frames"

hoch, line 11410: "the number of sample frames" -> "a varying number of sample-frames"

index.bs
```diff
@@ -11391,15 +11392,56 @@ The algorithm for rendering a block of audio from a {{BaseAudioContext}}
...
+		<a>system-level audio callback</a> returns.
```
hoch, line 11411: Also - "before the next system-level audio callback arrives."

IMO the callback can't return if the work isn't finished yet.

hoch, line 11411: Also - "before the next system-level audio callback arrives."

The callback can't return if the work isn't finished yet.

index.bs
```diff
@@ -11391,15 +11392,56 @@ The algorithm for rendering a block of audio from a {{BaseAudioContext}}
...
+
+		A <dfn>load value</dfn> is computed for each <a>system-level audio
+			callback</a>, by dividing its execution duration by the <a>system-level
+			audio callback buffer size</a> divided by {{BaseAudioContext/sampleRate}}.
+
+		The <a>load value</a> usually is below 1.0, meaning that it took less time
```
hoch, line 11417: Perhaps:

Ideally the load value is below 1.0, meaning that the renderer was able to finish the callback before the given callback interval. When the load value goes beyond 1.0, it means the system could not render audio fast enough resulting in a "buffer underrun".

hoch, line 11417: Perhaps:

Ideally the load value is below 1.0, meaning that the renderer was able to finish the callback before the given callback interval. When the load value goes beyond 1.0, it means the system could not render audio fast enough resulting in a "buffer underrun".

index.bs
```diff
@@ -11391,15 +11392,56 @@ The algorithm for rendering a block of audio from a {{BaseAudioContext}}
...
+	system could not render audio fast enough for real-time.
+
+		<div class=note>
+			The <a>render quantum size</a> for an audio graph is not necessarily a
+			divisor of the <a>system-level audio callback buffer size</a>. This causes
+			increased audio latencies and reduced possible maximum load without
+			<a>audio buffer underrun</a>.
+		</div>
+
+		An {{OfflineAudioContext}} is not required to have a system-level audio
```
hoch, line 11429: Perhaps:

Note that the concept of system-audio callback and the load value does not apply to the OfflineAudioContext.

hoch, line 11429: Perhaps:

Note that the concept of system-audio callback and the load value does not apply to the OfflineAudioContext.

hoch (commented):
Sorry that I didn't catch these in the previous round!

index.bs
padenot: I rephrased this.

index.bs
padenot: Rephrased as well.

index.bs
padenot: This is problematic, because intervals aren't necessarily regular (e.g. on Linux/PulseAudio, or Windows sometimes). This really is a matter of rendering it in less time that it takes to output.

hoch (approved):
LGTM

The markup also looks good. Feel free to merge!

PR #2486, opened 2022-04-26
Add "extract" class to old AudioContext IDL definition

tidoust:
The `data-no-idl` attribute that was added in #2462 tells Bikeshed not to process the current `AudioContext` IDL block. This is enough to work around duplicate definition errors that would otherwise occur.

However, even though it is not processed, Bikeshed still adds the IDL block to the [final IDL index](https://webaudio.github.io/web-audio-api/#idl-index), which thus contains both the current and future `AudioContext` interface definitions.

The `extract` class tells Bikeshed not to include the current IDL block in the final IDL index.

Note: I'm taking for granted that the working group would like the "future" IDL block to be the one that becomes authoritative, and in particular the one that shows up in Web Platform Tests. If the group rather wanted the "current" IDL block to remain the authoritative one for now, the `extract` class should rather be added to the future IDL block.

***
<a href="https://pr-preview.s3.amazonaws.com/tidoust/web-audio-api/pull/2486.html" title="Last updated on Apr 26, 2022, 8:27 AM UTC (0cccc72)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/2486/6efc9f0...tidoust:0cccc72.html" title="Last updated on Apr 26, 2022, 8:27 AM UTC (0cccc72)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1521,7 +1521,7 @@ enum AudioContextLatencyCategory {
 		Buttons here
 	</div>
 	<del cite=#c2444>
-		<xmp class="idl" data-no-idl>
+		<xmp class="idl extract" data-no-idl>
 		[Exposed=Window]
 		interface AudioContext : BaseAudioContext {
 			constructor (optional AudioContextOptions contextOptions = {});


hoch:
Thanks for the suggestion, François!

> The extract class tells Bikeshed not to include the current IDL block in the final IDL index.

Then can we remove `data-no-idl` in favor of `extract`? Or do we need to keep them both?

tidoust:
> > The extract class tells Bikeshed not to include the current IDL block in the final IDL index.
> 
> Then can we remove `data-no-idl` in favor of `extract`?

That would be too simple, wouldn't it? ;)

>  Or do we need to keep them both?

As long as authoring tools don't add support for such `<del>` and `<ins>` patterns, I think you need both classes. As far as I can tell:
- `data-no-idl` tells Bikeshed: "This block is normative IDL but don't attempt to parse it as IDL and to create links for it. It should still be included in the IDL index".
- `extract` tells Bikeshed: "This block is informative IDL, you can parse it as IDL and create links for it, but ignore it in the IDL index.

You don't want Bikeshed to create links for the block, otherwise it will report duplicate issues with the future version, so you need `data-no-idl`. And you don't want Bikeshed to add the current block to the final IDL index, so you also need `extract`.

hoch:
Thank you for clarification. TIL! 😃

Issue #2400, opened 2014-11-21
Access to a different output device: AudioContext.setSinkId()


cwilso:
Should be able to specify different audio devices, using media device selectors.

hoch:
So the consensus so far:
- This API only applies to AudioContext (real-time).
- The channel count heuristics will be handled by the destination node.
- How to handle an invalid device ID:
  - AudioContext(): use the default audio device.
  - AudioContext.setSinkId(): reject the promise.
  - 'null': AudioContext will use the "muted" mode, driven by a synthetic clock on a non-realtime priority worker thread.

hoch:
To get the API design discussion started, here's the initial WebIDL:

```webidl
enum AudioContextSinkCategory {
  "none"
}

partial interface AudioContext : BaseAudioContext {
  [SecureContext] Promise<undefined> setSinkId((DOMString or AudioContextSinkCategory) sinkId);
  readonly attribute DOMString sinkId;
  attribute EventHandler onsinkchange;
}

dictionary AudioContextOptions {
  (AudioContextLatencyCategory or double) latencyHint = "interactive";
  float sampleRate;
  (DOMString or AudioContextSinkCategory) sinkId;
}
```

Questions:
- What if the sample rate of the current device is different from the newly selected device?

hoch:
Teleconf 5/5:
- The overall API shape looks great. @hoch will start a draft/PR based on this proposal.
- `onsinkchange` won't be fired immediately after the construction. `onstatechange` after construction can already signal that the device is running.
- `sinkId` will be `undefined` when this field is not set by the constructor options or `setSinkId()`.

hoch:
I believe we can reuse the text from `AudioElement.setSinkId()`, but the context construction should be different. Here's my proposal:

---

When the AudioContext constructor is executed, the user agent MUST perform the following additional steps:

**AudioContext(contextOptions)**

1. Let context have a `[[SinkId]]` internal slot, initialize to “”.
2. If contextOption is given apply the options:
    1. If contextOptions.sinkId is specified, run the following substeps:
       1. if document is not allowed to use the feature identified by `“speaker-selection”`, abort these substeps.
       2. if `contextOptions.sinkId` is not the empty string and does not match any audio output device identified by the result that would be provided by `enumerateDevices()`, set `[[SinkId]]` to `contextOptions.sinkId` and abort these substeps.
[...]

PR #2498, opened 2022-07-20
AudioContext.setSinkId()

hoch:
[work-in-progress] Uploading this PR to have further discussion.

TODO:
- [ ] Add `onsinkchange` event handler
- [ ] Confirm defaulting behavior (with no valid output device)
- [ ] Confirm the process of "acquiring system resource"

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1511,6 +1511,28 @@ enum AudioContextLatencyCategory {
 </table>
 </div>
 
+<pre class="idl">
+enum AudioContextSinkCategory {
+		"none"
+};
+</pre>
+	
+<div class="enum-description">
+<table class="simple" dfn-type=enum-value dfn-for="AudioContextSinkCategory">
+	<thead>
+		<tr>
+			<th scope="col" colspan="2">
+				Enumeration description
+	<tbody>
+		<tr>
+			<td>
+				"<dfn>none</dfn>"
+			<td>
+				The audio graph will be rendered by a worker thread without using an
+				audio output device.
+</table>
+</div>
+
 <div class="correction proposed" id="c2444-1">
 	<span class="marker">Proposed Addition
 		<a href="https://github.com/WebAudio/web-audio-api/issues/2444">Issue
@@ -1546,6 +1568,7 @@ enum AudioContextLatencyCategory {
 			constructor (optional AudioContextOptions contextOptions = {});
 			readonly attribute double baseLatency;
 			readonly attribute double outputLatency;
+			readonly attribute DOMString sinkId;
 			[SecureContext] readonly attribute AudioRenderCapacity renderCapacity;
 			AudioTimestamp getOutputTimestamp ();
 			Promise<undefined> resume ();
@@ -1567,13 +1590,18 @@ allows the context state to transition from "{{AudioContextState/suspended}}" to
 and to allow it only when the {{AudioContext}}'s [=relevant global object=] has
 [=sticky activation=].
 
-{{AudioContext}} has an internal slot:
+{{AudioContext}} has two internal slots:
 
 <dl dfn-type=attribute dfn-for="AudioContext">
 	: <dfn>[[suspended by user]]</dfn>
 	::
 		A boolean flag representing whether the context is suspended by user code.
 		The initial value is <code>false</code>.
+
+	: <dfn>[[sink ID]]</dfn>
+	::
+		A DOMString represending the ID of the current audio output device. The
+		initial value is <code>""</code>.
 </dl>
 
 <h4 id="AudioContext-constructors">
@@ -1592,36 +1620,71 @@ Constructors</h4>
 			<span class="synchronous">When creating an {{AudioContext}},
 			execute these steps:</span>
 
-			1. Set a {{[[control thread state]]}} to <code>suspended</code> on the {{AudioContext}}.
+			1. Let |context| be a new {{AudioContext}} object.
 
-			2. Set a {{[[rendering thread state]]}} to <code>suspended</code> on the {{AudioContext}}.
+			1. Let |document| be the current settings object's
+				[=associated Document=].
 
-			3. Let <dfn attribute for="AudioContext">[[pending resume promises]]</dfn> be a
-				slot on this {{AudioContext}}, that is an initially empty ordered list of
+			1. Set a {{[[control thread state]]}} to <code>suspended</code> on
+				|context|.
+
+			1. Set a {{[[rendering thread state]]}} to <code>suspended</code> on
+				|context|.
+
+			1. Let 
+				<dfn attribute for="AudioContext">[[pending resume promises]]</dfn> be
+				a slot on |context|, that is an initially empty ordered list of
 				promises.
 
-			4. If <code>contextOptions</code> is given, apply the options:
+			1. If <code>contextOptions</code> is given, perform the following
+				substeps:
 
-				1. Set the internal latency of this {{AudioContext}}
-					according to <code>contextOptions.{{AudioContextOptions/latencyHint}}</code>, as described
-					in {{AudioContextOptions/latencyHint}}.
+				1. If {{AudioContextOptions/sinkId}} is specified, set |sinkId| to
+					<code>contextOptions.{{AudioContextOptions/sinkId}}</code> and run
+					the following substeps:
 
-				2. If <code>contextOptions.{{AudioContextOptions/sampleRate}}</code> is specified,
-					set the {{BaseAudioContext/sampleRate}}
-					of this {{AudioContext}} to this value. Otherwise, use
-					the sample rate of the default output device. If the
-					selected sample rate differs from the sample rate of the
-					output device, this {{AudioContext}} MUST resample the
-					audio output to match the sample rate of the output device.
+					1. If |document| is not allowed to use the feature identified by
+						<code>“speaker-selection”</code>, |sinkId| is the empty string,
+						or |sinkId| does not match any audio output device identified by
+						the result that would be provided by
+						{{MediaDevices/enumerateDevices()}}, abort these substeps.
+					
+					1. Set {{AudioContext/[[sink ID]]}} to |sinkId|.
 
-					Note: If resampling is required, the latency of the
-					AudioContext may be affected, possibly by a large
-					amount.
+				1. Set the internal latency of |context| according to
+					<code>contextOptions.{{AudioContextOptions/latencyHint}}</code>, as
+					described in {{AudioContextOptions/latencyHint}}.
 
-			5. If the context is <a>allowed to start</a>, send a
-				<a>control message</a> to start processing.
+				1. If <code>contextOptions.{{AudioContextOptions/sampleRate}}</code> is
+					specified, set the {{BaseAudioContext/sampleRate}} of |context| to
+					this value. Otherwise, use the sample rate of the default output
+					device. If the selected sample rate differs from the sample rate of
+					the output device, the user agent MUST resample the audio output to
+					match the sample rate of the output device.
 
-			6. Return this {{AudioContext}} object.
+					Note: If resampling is required, the latency of |context|	may be
+					affected, possibly by a large amount.
+
+			1. Run the following substeps to prepare an audio output device for
+					rendering:
+	
+				1. If {{AudioContext/[[sink ID]]}} is
+					<code>"{{AudioContextSinkCategory/none}}"</code>, acquire system
+					resources to render the audio graph without an actual audio output
+					device. Abort these substeps.
+
+				1. If {{AudioContext/[[sink ID]]}} is the empty string, acquire system
+					resources to render the audio graph with a default audio output
+					device. Abort these substeps.
+
+				1. If {{AudioContext/[[sink ID]]}} is not the empty string, acquire
+					system resources to render the audio graph with the underlying audio
+					output device identified by {{AudioContext/[[sink ID]]}}.
+
+			1. If the context is <a>allowed to start</a>, send a
+					<a>control message</a> to start processing.
+	
+			1. Return |context| object.
 		</div>
 
 		<div algorithm="sending a control message to start processing">
@@ -1708,6 +1771,12 @@ Attributes</h4>
 
 	</ins>
 </div>
+
+	: <dfn>sinkId</dfn>
+	::
+		This attribute contains the ID of the current audio output device, or the
+		empty string for the user-agent default device. The attribute MUST return
+		the value of {{AudioContext/[[sink ID]]}} internal slot upon getting.
 </dl>
 
 <h4 id="AudioContext-methods">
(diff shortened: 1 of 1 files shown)


hoch:
@padenot PTAL - I've made an attempt on setSinkId() based on our discussion. I'll add `onsinkchange` event handler soon, but just wanted to kick off the review first.

index.bs
```diff
@@ -2073,6 +2151,69 @@ Methods</h4>
...
+				1. Let |sinkId| be the method's first argument.
+
+				1. If |sinkId| is equal to |context|'s {{AudioContext/[[sink ID]]}}
+					internal slot, return a promise resolved with {{undefined}}.
+
+				1. Let |p| be a new promise.
+
+				1. Run the following substeps:
+
+					1. If |sinkId| is the empty string, reject |p| with a new
```
hoch, line 2180: If the context is already closed, we should reject the promise as well.

index.bs
```diff
@@ -1592,36 +1621,71 @@ Constructors</h4>
...
+					1. Set {{AudioContext/[[sink ID]]}} to |sinkId|.
+
+				1. Set the internal latency of |context| according to
+					<code>contextOptions.{{AudioContextOptions/latencyHint}}</code>, as
+					described in {{AudioContextOptions/latencyHint}}.
+
+				1. If <code>contextOptions.{{AudioContextOptions/sampleRate}}</code> is
+					specified, set the {{BaseAudioContext/sampleRate}} of |context| to
+					this value. Otherwise, use the sample rate of the default output
+					device. If the selected sample rate differs from the sample rate of
```
chrisguttandin, line 1662: "Otherwise, use the sample rate of the default output device." I think this sentence needs to be updated to pick the sample rate of the device selected with the `sinkId` if present.

index.bs
```diff
@@ -1567,13 +1591,18 @@ allows the context state to transition from "{{AudioContextState/suspended}}" to
...
 
 <dl dfn-type=attribute dfn-for="AudioContext">
 	: <dfn>[[suspended by user]]</dfn>
 	::
 		A boolean flag representing whether the context is suspended by user code.
 		The initial value is <code>false</code>.
+
+	: <dfn>[[sink ID]]</dfn>
+	::
+		A DOMString represending the ID of the current audio output device. The
```
chrisguttandin, line 1604:
```suggestion
		A DOMString representing the ID of the current audio output device. The
```

index.bs
```diff
@@ -1592,36 +1621,71 @@ Constructors</h4>
...
+
+				1. If {{AudioContextOptions/sinkId}} is specified, set |sinkId| to
+					<code>contextOptions.{{AudioContextOptions/sinkId}}</code> and run
+					the following substeps:
+
+					1. If |document| is not allowed to use the feature identified by
+						<code>“speaker-selection”</code>, |sinkId| is the empty string,
+						or |sinkId| does not match any audio output device identified by
+						the result that would be provided by
+						{{MediaDevices/enumerateDevices()}}, abort these substeps.
```
chrisguttandin, line 1651: I wonder if there also needs to be a special case for the "none" value. At least as far as I understand it it would otherwise cause the algorithm to be aborted here.

index.bs
```diff
@@ -1511,6 +1511,28 @@ enum AudioContextLatencyCategory {
...
+	<thead>
+		<tr>
+			<th scope="col" colspan="2">
+				Enumeration description
+	<tbody>
+		<tr>
+			<td>
+				"<dfn>none</dfn>"
+			<td>
+				The audio graph will be rendered by a worker thread without using an
```
padenot, line 1531: I don't think we want to be prescriptive here by saying "worker". Let's just say the audio will be processed without being output to an audio output device.

index.bs
padenot, line 1604: Probably need a bit of markup on `DOMString`.

index.bs
```diff
@@ -1592,36 +1621,71 @@ Constructors</h4>
...
-					audio output to match the sample rate of the output device.
+			1. If <code>contextOptions</code> is given, perform the following
+				substeps:
+
+				1. If {{AudioContextOptions/sinkId}} is specified, set |sinkId| to
+					<code>contextOptions.{{AudioContextOptions/sinkId}}</code> and run
+					the following substeps:
+
+					1. If |document| is not allowed to use the feature identified by
+						<code>“speaker-selection”</code>, |sinkId| is the empty string,
```
padenot, line 1648: You probably want regular double-quotes here (ASCII `0x22`), and not Unicode left quotation mark.

index.bs
```diff
@@ -1592,36 +1621,71 @@ Constructors</h4>
...
-					set the {{BaseAudioContext/sampleRate}}
-					of this {{AudioContext}} to this value. Otherwise, use
-					the sample rate of the default output device. If the
-					selected sample rate differs from the sample rate of the
-					output device, this {{AudioContext}} MUST resample the
-					audio output to match the sample rate of the output device.
+			1. If <code>contextOptions</code> is given, perform the following
+				substeps:
+
+				1. If {{AudioContextOptions/sinkId}} is specified, set |sinkId| to
```
padenot, line 1643: I think the preferred wording for this is something like "let |sinkId| be the value of <code>contextOptions.{{AudioContextOptions/sinkId}}</code>".

index.bs
padenot: Yes.

index.bs
padenot: Yes.

index.bs
```diff
@@ -1592,36 +1621,71 @@ Constructors</h4>
...
-					amount.
+				1. If {{AudioContext/[[sink ID]]}} is the empty string, acquire system
+					resources to render the audio graph with a default audio output
+					device. Abort these substeps.
 
-			5. If the context is <a>allowed to start</a>, send a
-				<a>control message</a> to start processing.
+				1. If {{AudioContext/[[sink ID]]}} is not the empty string, acquire
+					system resources to render the audio graph with the underlying audio
+					output device identified by {{AudioContext/[[sink ID]]}}.
```
padenot, line 1683: "Acquiring system resources" historically happens when starting, not when initializing. It's asynchronous to the `AudioContext` construction. It's lazy, so to speak. I think those steps are best spelled out in the control message part.

index.bs
padenot: Agreed.

index.bs
```diff
@@ -2073,6 +2151,69 @@ Methods</h4>
...
+						1. Set context's {{AudioContext/[[sink ID]]}} internal slot to
+							|sinkId|.
+
+						1. If {{AudioContext/[[sink ID]]}} is <code>"{{AudioContextSinkCategory/none}}"</code>, acquire system
+							resources to render the audio graph without an actual audio
+							output device. Abort these substeps.
+
+						1. If {{AudioContext/[[sink ID]]}} is not the empty string, acquire system
+							resources to render the audio graph with the underlying audio
+							output device identified by {{AudioContext/[[sink ID]]}}.
```
padenot, line 2209: What would be cool here is to suspend the graph, so that authors can use `onstatechange` to determine when the audio is flowing again, because it can take some time for this to happen, and the only option is to resort to polling `currentTime`.

We should be able to reuse the `suspend` and `resume` methods. This has also the advantage of making it clear that this will happen in between two render quantums.

index.bs
hoch: Let's discuss this again in light of the new proposal:
https://github.com/WebAudio/web-audio-api/issues/2400#issuecomment-1230466457

index.bs
hoch: That's a good catch! :)

index.bs
hoch: Agreed. Done.

index.bs
hoch: Not sure I am following here - so setSinkId() will cause to suspend, and then automatically resume later? So the user code use the `onstatechange` signal for the renderer's readiness?

I think `onsinkchange` might be a better signal for it.

I guess this should be another topic for discussion?

index.bs
```diff
@@ -1511,6 +1511,28 @@ enum AudioContextLatencyCategory {
...
+	<thead>
+		<tr>
+			<th scope="col" colspan="2">
+				Enumeration description
+	<tbody>
+		<tr>
+			<td>
+				"<dfn>none</dfn>"
+			<td>
+				The audio graph will be processoed without being played through an
```
otherdaniel, line 1531: [drive-by comment; feel free to ignore.] processoed -> processed

hoch: Thank you!

index.bs
```diff
@@ -1511,6 +1511,28 @@ enum AudioContextLatencyCategory {
...
+		"none"
+};
+</pre>
+	
+<div class="enum-description">
+<table class="simple" dfn-type=enum-value dfn-for="AudioContextSinkCategory">
+	<thead>
+		<tr>
+			<th scope="col" colspan="2">
+				Enumeration description
```
padenot, line 1525: Can you rebase this over the changes: https://github.com/WebAudio/web-audio-api/commit/840e23e00e00297e764e6b93ea60c86ac82e0fa2#diff-5e793325cd2bfc452e268a4aa2f02b4024dd9584bd1db3c2595f61f1ecf7b985R793-R796 ?

index.bs
```diff
@@ -1708,6 +1773,20 @@ Attributes</h4>
 
 	</ins>
 </div>
+
+	: <dfn>sinkId</dfn>
+	::
+		This attribute contains the ID of the current audio output device, or the
+		empty string for the user-agent default device. The attribute MUST return
+		the value of {{AudioContext/[[sink ID]]}} internal slot upon getting.
```
padenot, line 1781: I think that it's better to only say that this returns the value of `[[sink ID]]` and not have too much prose, to not have to repeat what the values can be (e.g., this can also now return `"none"`).

index.bs
```diff
@@ -1708,6 +1773,20 @@ Attributes</h4>
...
+
+	: <dfn>onsinkchange</dfn>
+	::
+		An EventHandler for {{AudioContext/setSinkId()}}. This event will be
```
padenot, line 1785: `{{}}` on `EventHandler` probably.

index.bs
```diff
@@ -1708,6 +1773,20 @@ Attributes</h4>
...
+		dispatched when changing the output device is completed. NOTE: This is not
```
padenot, line 1786: `NOTE: ...` should be in its own paragraph so that it appears in a green box.

index.bs
```diff
@@ -2073,6 +2152,74 @@ Methods</h4>
...
+					1. If the {{BaseAudioContext/state}} attibutes of |context| is
+						"{{AudioContextState/closed}}", reject |p| with a new
+						{{DOMException}} whose name is "{{InvalidStateError}}" and abort
+						these substeps.
+
+				  1. If |sinkId| is the empty string, reject |p| with a new
+						{{DOMException}} whose name is "{{NotFoundError}}" and abort these
+						substeps.
+
+					1. If |sinkId| does not match any audio output device identified by
```
padenot, line 2190: This must be indented one level less, I think.

index.bs
```diff
@@ -2073,6 +2152,74 @@ Methods</h4>
...
+						{{MediaDevices/enumerateDevices()}}, reject |p| with a new
+						{{DOMException}} whose name is "{{NotFoundError}}" and abort these
+						substeps.
+
+					1. If the device specified by |sinkId| is not the user agent's
+						default device or is not permitted to play audio through, reject |p|
+						with a new {{DOMException}} whose name is {{NotAllowedError}} and
+						abort these substeps.
+
+					1. Stop playing audio out of the current output device.
```
padenot, line 2201: We probably need to integrate this to the graph processing algorithm. Maybe mentioning it happens between render quantum or something?

index.bs
padenot: Does `onsinkchange` fire if the default audio output changes? i.e. you're just playing on the default audio output device, and it's being unplugged, so that the device has changed? It would be useful: something like the sample-rate or optimal render quantum size could be different, authors might want to rebuild their audio graph.

Other than that, my proposal was that:
- The promise from `setSinkID(...)` is resolved after the audio is flowing again (i.e. it's resolved after the operation has succeeded)
- The `AudioContext.state` is `"suspended"` while the graph is not being rendered because the underlying audio output device is changing
- `onstatechange` fires normally.

It would go like this:
- `setSinkID(...)` is called with a valid device identifier, returning a promise `p`
- The rendering is paused, possibly after the current rendering quantum
- `"statechanged"` is fired, the state is now `"suspended"`
- The old device's stream is closed
- A new system-level audio stream is opened, on the new device
- Some time passes, opening the device can take time
- The new stream initialization has finished
- Queue a task to fire `"sinkchange"` the `sinkID` is now the new `sinkID`, resolve `p`
- Queue a task to fire `"statechange"`, the state is now `"running"`

index.bs
hoch: Yes, now I can make this simplification.

index.bs
hoch: This is now handled by a new sub-algorithm introduced at https://github.com/WebAudio/web-audio-api/pull/2498/commits/1d6292645873ddb39cf7635634fd7a5233447f4f.

index.bs
hoch: I'll consider this change in the next round of editing.

index.bs

hoch:
PTAL again @padenot and @chrisguttandin - I think the last revision covers what @padenot suggested in https://github.com/WebAudio/web-audio-api/pull/2498#discussion_r974618408.

index.bs
```diff
@@ -2166,8 +2237,125 @@ Methods</h4>
...
+		<code>"speaker-selection"</code>, return a new {{DOMException}} whose name
+		is "{{NotAllowedError}}".
+	1. If |sinkIdArg| is a type of {{AudioSinkOptions}} and its
+		{{AudioSinkOptions/type}} property is not equal to
+		<code>"{{AudioSinkType/none}}"</code>, return a new {{DOMException}} whose
+		name is "{{NotSupportedError}}".
+	1. If |sinkIdArg| is a type of {{DOMString}} but it does not match any audio
+		output device identified by the result that would be provided by
+		{{MediaDevices/enumerateDevices()}}, return a new {{DOMException}} whose
+		name is "{{InvalidStateError}}".
```
padenot, line 2354: Or `NotFoundError` maybe.

index.bs
```diff
@@ -2166,8 +2237,125 @@ Methods</h4>
...
+
+							1. Set the {{BaseAudioContext/state}} attribute of the
+								{{AudioContext}} to "{{AudioContextState/suspended}}".
+							
+							1. <a spec="dom" lt="fire an event">Fire an event</a> to
+								{{BaseAudioContext/onstatechange}} {{EventHandler}} at the
+								associated {{AudioContext}}.
+
+				1. Attempt to <a href="#acquiring">acquire system resources</a> to use
+					an audio output device identified {{AudioContext/[[sink ID]]}} for
```
padenot, line 2300: missing a word here? `identified by` maybe?

index.bs
```diff
@@ -2166,8 +2237,125 @@ Methods</h4>
...
+					rendering. In case of failure, reject |p| with
+					"{{InvalidAccessError}}" abort the following steps.
+				
+				1. <a href="https://html.spec.whatwg.org/multipage/media.html#queue-a-media-element-task">
+					Queue a media element task</a> to execute the follwoing steps:
```
padenot, line 2305: `s/follwoing/following/`

index.bs
```diff
@@ -2166,8 +2237,125 @@ Methods</h4>
...
+				1. Set |wasRunning| to false if the {{[[rendering thread state]]}} on
+					the {{AudioContext}} is <code>"suspended"</code>.
+
+				1. Pause the renderer after processing the current render quantum.
+
+				1. Attempt to <a>release system resources</a>.
+
+				1. If |wasRunning| is true:
+
+					1. Set the {{[[rendering thread state]]}} on the {{AudioContext}} to
```
padenot, line 2283: It feels like this (and it's "resume" counterpart below) should be spun up in their own algorithm, but we can do it later.

padenot (approved):
A couple nits, but looks good, please merge at your convenience!

index.bs
```diff
@@ -1602,36 +1638,61 @@ Constructors</h4>
...
-						amount.
+					1. If <code>contextOptions</code> is given, perform the following
+						substeps:
+		
+						1. If {{AudioContextOptions/sinkId}} is specified, let |sinkId| be
+							the value of
+							<code>contextOptions.{{AudioContextOptions/sinkId}}</code>.
+		
+						1. Let |validationResult| be the return value of
+							<a href="#validating-sink-identifier">sink identifier	validation
```
chrisguttandin, line 1657: It looks like a tab or some special whitespace slipped in here.

```suggestion
							<a href="#validating-sink-identifier">sink identifier validation
```

index.bs
```diff
@@ -1602,36 +1638,61 @@ Constructors</h4>
...
+							
+							1. If |sinkId| is a {{DOMString}}, use the sample rate of the
+								output device identified by |sinkId|. Abort these substeps.
+							
+							If <code>contextOptions.{{AudioContextOptions/sampleRate}}</code>
+							differs from the sample rate of the output device, the user agent
+							MUST resample the audio output to match the sample rate of the
+							output device.
+		
+							Note: If resampling is required, the latency of |context|	may be
```
chrisguttandin, line 1689: Same as above.

```suggestion
							Note: If resampling is required, the latency of |context| may be
```

index.bs
```diff
@@ -1692,26 +1749,27 @@ Constructors</h4>
 			Sending a <a>control message</a> to start processing means
 			executing the following steps:
 
-			1. Attempt to <a href="#acquiring">acquire system resources</a>.
-				In case of failure, abort the following steps.
+			1. Attempt to <a href="#acquiring">acquire system resources</a> to use an
+				audio output device for rendering.	In case of failure, abort the
```
chrisguttandin, line 1753:
```suggestion
				audio output device for rendering. In case of failure, abort the
```

chrisguttandin (approved):
Thanks for pinging me again. I only spotted a few "tabs".

While reading the text I was wondering what would be the `sampleRate` of an `AudioContext` with `{ type: 'none' }` as the `sinkId`? Do we need to define a default `sampleRate` for that case?

```js
const audioContext = new AudioContext({ sinkId: { type: 'none' } });

console.log(audioContext.sampleRate); // ?
```

I was also wondering if the algorithm for changing the output device should be aborted if the the device doesn't change. As far as I understand it the following code (which is of course a silly example) would still trigger the algorithm although it's basically not changing anything.

```js
audioContext.setSinkId(audioContext.sinkId);
```

index.bs
hoch: ACK

hoch:
@chrisguttandin Yes. I thought about it as well. Here's my proposal:

1. If {type: "none"}, use the sample rate of the default audio device if unspecified.
2. For your second question, I think the "do nothing" case only happens when |sinkId| is the same with the current one. All other case will be handled by the rejected promise.

chrisguttandin:
Since I often run the Web Audio API in tests on CI servers I had to learn the hard way that not every computer (or VM) has an audio device. :-( I'm not sure though if that rare edge case needs to be handled.

I think you're right, the second question can probably be ignored. Calling `setSinkId()` with exactly the same `sinkId` as the one of the currently used device can be easily avoided by a user of the API with a simple comparison.

hoch:
Hmm. That's an interesting case - but it easily leads into a different question. setSinkId() won't even be a problem because the AudioContext construction will fail on such system. IMO it needs a new spec issue for further discussion.

hoch:
@padenot While cleaning up and preparing for mark up, I made some changes to improve the algorithm:

- When |sinkId| is the same with the current one, exit early (on constructor) or resolve it right away (on setSinkId()).
- Now the ID validation subroutine returns a DOMException or null. (Previously it was returning a DOMException or boolean. I thought that could be confusing)

Other than that changes are editorial. Please take one more look - in the mean time I'll start marking things up for landing.

hoch:
Okay, the markup is done and this PR is ready to land.

PR #2516, opened 2022-10-05
Link event handlers IDL attributes to event types

tidoust:
Changes in #2498 introduce a weird phrasing for "fire an event", e.g. "Fire an event to onstatechange EventHandler". Correct phrasing should rather be "Fire an event named statechange".

This made me realize that the spec never associates the event handler IDL attributes that it defines with the actual event handler event type that gets fired. The spec also seems to assume that there will be at most one event handler (e.g. when it says that "This AudioBuffer is only valid while in the scope of the onaudioprocess function"), whereas `onxxx` is just one way to listen to `xxx` events, `addEventListener` may also be used.

Specs typically make the association between event handler IDL attributes and event handler event types explicit, e.g. as done in HTML through tables such as: https://html.spec.whatwg.org/multipage/webappapis.html#event-handlers-on-elements,-document-objects,-and-window-objects

... or in WebRTC through text in the definition of the IDL attribute: https://w3c.github.io/webrtc-pc/#ref-for-event-datachannel-bufferedamountlow-1

This pull request adds definitions of event types next to the definitions of the `onxxx` IDL attributes with which they are associated and uses references to these event types whenever the spec fires an event or mentions event handlers.

I tried to keep the changes minimal. I was going to argue that these changes are editorial in nature as they merely clarify something that did not create ambiguities for implementations. Now, you seem to track all changes with "proposed corrections", regardless of whether they're editorial or substantive so I suspect you cannot accept these changes as-is.

At a minimum, I think that the occurrences of "fire an event to onxxx" should be fixed (they only appear in proposed corrections so that seems doable without creating additional proposed corrections). I can prepare a separate PR to that effect. I can also look into creating appropriate "proposed corrections" structures for the other changes if they seem useful. That may not be worth the hassle ;)

***
<a href="https://pr-preview.s3.amazonaws.com/tidoust/web-audio-api/pull/2516.html" title="Last updated on Oct 5, 2022, 10:22 AM UTC (11b6897)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/2516/cc718a3...tidoust:11b6897.html" title="Last updated on Oct 5, 2022, 10:22 AM UTC (11b6897)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -930,17 +930,18 @@ Attributes</h4>
 
 	: <dfn>onstatechange</dfn>
 	::
-		A property used to set the <code>EventHandler</code> for an
+		A property used to set an [=event handler=] for an
 		event that is dispatched to
 		{{BaseAudioContext}} when the state of the
 		AudioContext has changed (i.e. when the corresponding promise
-		would have resolved). An event of type
-		{{Event}} will be dispatched to the event
+		would have resolved). The event type of this event handler is
+		<dfn event>statechange</dfn>. An event that uses the
+		{{Event}} interface will be dispatched to the event
 		handler, which can query the AudioContext's state directly. A
 		newly-created AudioContext will always begin in the
 		<code>suspended</code> state, and a state change event will be
 		fired whenever the state changes to a different state. This
-		event is fired before the {{oncomplete}} event
+		event is fired before the {{complete}} event
 		is fired.
 
 	: <dfn>sampleRate</dfn>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1185,7 +1186,7 @@ Methods</h4>
...
... parameter determines the buffer size in units of sample-frames. If it's not passed in, or if the value is 0, then the implementation will choose the best buffer size for the given environment, which will be constant power of 2 throughout the lifetime of the node. Otherwise if the author explicitly specifies the bufferSize, it <em class="rfc2119" title="MUST">MUST</em> be one of the following values: 256, 512, 1024, 2048, 4096, 8192, 16384. This value controls how frequently the {{ScriptProcessorNode/audioprocess}} event is dispatched and how many sample-frames need to be processed each call. Lower values for {{ScriptProcessorNode/bufferSize}} will result in a lower (better) <a href="#latency">latency</a>. Higher values will be necessary to avoid audio breakup and <a href="#audio-glitching">glitches</a>. It is recommended for authors to not specify this buffer size and allow the implementation to pick a good buffer size to balance between <a href="#latency">latency</a> and audio quality. If the value of this parameter is not one of the allowed power-of-2 values listed above, <span class="synchronous">an {{IndexSizeError}} <em class="rfc2119" title="MUST">MUST</em> be thrown</span>.
```
hoch, line 1189: This is not relevant to the most recent change. Do we still want to do that? If so, shouldn't we wrap this with the proposed addition markup?

hoch (approved):
LGTM. Thanks for this change, @tidoust!

My only question is how to handle the propose addition markup on this. If you believe it's not necessary because this is a non-substantive change, we can simply merge this.

tidoust:
> My only question is how to handle the propose addition markup on this. If you believe it's not necessary because this is a non-substantive change, we can simply merge this.

Yes, it's up to you but I would argue that these changes are [corrections that do not affect conformance](https://www.w3.org/2021/Process-20211102/#class-2): they are meant to align the spec with the wording used in most other specs but I would be extremely surprised if the previous text had created any ambiguity for implementations that this update would resolve.

I don't think that the W3C process requires "proposed addition" markup for such changes. Unless the Audio Working Group resolved that all changes need to be wrapped in "proposed addition" markup, I would keep things simple, be it only to improve readability and ease maintenance on your end.

index.bs
tidoust: I went through all the events defined in the spec. As said, I think these changes are editorial improvements in nature so do not require proposed addition markup. But that's up to you ;)

hoch:
>  I would keep things simple, be it only to improve readability and ease maintenance on your end.

Agreed.

hoch:
@padenot Friendly ping - I don't think this change is controversial. Would like to land if you're okay! :)

padenot (approved):
Sorry I missed this, this looks very good, thanks for doing it François!

PR #2558, opened 2023-10-12
Specify `start()` and `stop()` in AudioRenderCapacity

### LINKED ISSUE #2526: Calling AudioRenderCapacity.start multiple times
**Describe the issue**

Reading the [AudioRenderCapacity](https://webaudio.github.io/web-audio-api/#AudioRenderCapacity) spec, it is unclear to me what the behaviour should be when calling `start` repeatedly. Should it:
- ignore subsequent calls
- update the update interval to the latest setting
- spawn another metric collection callback

**Where Is It**

https://webaudio.github.io/web-audio-api/#ref-for-dom-audiorendercapacity-start

> Starts metric collection and analysis. This will repeatedly [fire an event](https://dom.spec.whatwg.org/#concept-event-fire) named [update](https://webaudio.github.io/web-audio-api/#eventdef-audiorendercapacity-update) at [AudioRenderCapacity](https://webaudio.github.io/web-audio-api/#audiorendercapacity), using [AudioRenderCapacityEvent](https://webaudio.github.io/web-audio-api/#audiorendercapacityevent), with the given update interval in [AudioRenderCapacityOptions](https://webaudio.github.io/web-audio-api/#dictdef-audiorendercapacityoptions).

**Additional Information**

None

hoch:
Fixes #2526.

FYI @padenot - the current draft is without the proposed addition markup. I'll add markups once the PR is reviewed.

***
<a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/pull/2558.html" title="Last updated on Oct 12, 2023, 8:00 PM UTC (2931820)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/2558/0c5f2df...2931820.html" title="Last updated on Oct 12, 2023, 8:00 PM UTC (2931820)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2852,6 +2852,20 @@ Dictionary {{AudioTimestamp}} Members</h5>
         {{AudioContext}}. In order to calculate them, the renderer collects a
         <a>load value</a> per <a>system-level audio callback</a>.
 
+        This interface has two internal slots:
+
+        <dl dfn-type=attribute dft-for="AudioRenderCapacity">
+            : <dfn>[[is running]]</dfn>
+            :: 
+                A boolean flag indicating whether the metric reporting loop is 
+                running. The initial value is <code>false</code>.
+
+            : <dfn>[[update interval]]</dfn>
+            ::
+                A double number that represents the interval of the metric
+                reporting loop in seconds. The initial value is 1.
+        </dl>
+
         <h5 id="AudioRenderCapacity-attributes">
         Attributes</h5>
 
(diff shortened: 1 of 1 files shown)


hoch:
@padenot Friendly ping!

index.bs
```diff
@@ -2874,13 +2888,47 @@ Dictionary {{AudioTimestamp}} Members</h5>
...
 
+        <h6 id="AudioRenderCapacity-metric-reporting-loop">
+        Metric Reporting Loop</h6>
+
+        This algorithm can be invoked by {{AudioRenderCapacity/start}} method
+        to report the performance metric repreatedly.
+
+        1. If {{[[is running]]}} is <code>false</code>, abort this algorithm.
+        2. Create |E|, which is a new instance of {{AudioRenderCapacityEvent}},
+            with computed performance metrics.
```
padenot, line 2924: We might need to have something that stores the load values on the context as a private slot (e.g. a list) and perform the average/peak/underrun calculation, resets the list, fire the event, etc. so it's clear what's going on.

Besides, do we want this to be clocked off of the rendering clock or the main thread?

I can see things going both ways:
- after a delay of `[[update interval]]` in the audio context clock domain, the rendering algorithm queues a task with all the data to the main thread, continuously, while it's running. No event is fired when not running (e.g. context suspended).
- after a delay of `[[update interval]]` in the system clock clock-domain, the main thread gathers the performance data, computes and fires the event. If the `AudioContext` is suspended, we can decide to fire or not fire the event.

I'd lean towards the second solution, _not_ firing the event when suspended. I think it's more useful: even if underrunning horribly, authors can know.

But what happens if the graph is so big that not a single rendering quantum was completed in the interval? That easy to simulate, and we should make it so that it reports 100% of load or something like that (some software report > 100%, that's also well-defined what it means).

hoch: So you proposal is to expand "computed performance metrics" with more detailed steps? Right?

+1 on the second option.

For your last question, the API doesn't really specify how the renderer can recover from such problem. So we might see different behavior depending on the implementation. Some will drop subsequent system audio callbacks until the renderer becomes responsive while others will continue to process subsequent callbacks and accumulate more buffer underruns.

All that aside, 1.0 (100%) is practical. I am not sure what users can do with a number beyond 1.0.

index.bs
```diff
@@ -2874,13 +2888,47 @@ Dictionary {{AudioTimestamp}} Members</h5>
...
+                </div>
         </dl>
 
+        <h6 id="AudioRenderCapacity-metric-reporting-loop">
+        Metric Reporting Loop</h6>
+
+        This algorithm can be invoked by {{AudioRenderCapacity/start}} method
+        to report the performance metric repreatedly.
+
+        1. If {{[[is running]]}} is <code>false</code>, abort this algorithm.
```
chrisguttandin, line 2922: I think if `is running` is only checked here it could result in multiple Metric Reporting Loops running in parallel.

I think the following could cause two parallel loops.

```
start();
stop();
start();
```

Since `stop()` only sets the `is running` flag to `false` but doesn't stop the loop both loops started by calling `start()` could be kept alive. It seems to be possible that by the time the first loop checks `is running` it's true again as a result of calling `start()` for the second time.

hoch:
This effort is currently on hold because of other blockers (TAG review, Privacy/Security review).

hoch:
Teleconf 5/30/2024:

- The development of this API in the Chromium project is on hold.
- There still are strong developer interest on this feature, and we need more input from other browser vendors than Chromium.

Issue #2532, opened 2023-01-12
Fire an event when the audio device that's in use in an AudioContext changes


padenot:
We can probably use `"devicechange"`. This is useful, because changing the device can have problematic performance issue, because the block size, latency and channel count can change.

hoch:
How is different from `onsinkchange`?

https://webaudio.github.io/web-audio-api/#dom-audiocontext-onsinkchange

padenot:
This would be fired when the underlying device changes, not necessarily in reaction to `setSinkId(...);` having been called on the `AudioContext`.

It might be possible to reuse the even `"sinkchange"`. It might be desirable to be able to distinguish a sink change because of `setSinkId(...)` and something external though, we need to think about it.

hoch:
In 02/09 Teleconf:

@padenot suggested that there might be a tricky case where setSinkId() user call could happen almost at the same time of user plugging out an audio device from the system. We need to think about the order of events and how to handle this case properly.

hoch:
2023 TPAC Audio WG Discussion:
- Use `sinkchange` event for a platform-orignated callback when there is a change in the system device registry. 
- Set `sinkId` property under AudioContext to the empty string ("") when there’s no audio device to fallback.

PR #2580, opened 2024-03-27
AudioContext.onerror

### LINKED ISSUE #2567: Device-related error reporting via AudioContext.onerror
There were several requests in the past to expose an event handler for device-related error reporting.

Currently AudioContext does not expose this type of error; it just silently fails and falls back to browser-specific default. The spec says: 

> Note: It is unfortunately not possible to programatically notify authors that the creation of the AudioContext failed. User-Agents are encouraged to log an informative message if they have access to a logging mechanism, such as a developer tools console.

One proposal would be having `onrendererror` under AudioContext. This can be used for cases like:

1. Selecting device via the context constructor options fails.
2. The current device gets lost for some reasons. (e.g. losing a bluetooth connection)
3. The system does not have any audio device available.

hoch:
This PR resolves #2567.

- Add the `onerror` event handler under the `AudioContext` interface.
- Add a step to fire `onerror` event in the AudioContext constructor steps on the render thread.
- Remove the non-normative notes on the notification being impossible.
- Add a prose to specify the `onerror` case on a running AudioContext.

***
<a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/pull/2580.html" title="Last updated on Jul 9, 2024, 6:42 PM UTC (8776c75)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/2580/22d71ec...8776c75.html" title="Last updated on Jul 9, 2024, 6:42 PM UTC (8776c75)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1997,7 +1997,7 @@ Constructors</h4>
         </ins>
     </div>
 
-        <div class="correction proposed" id="c2567-2">
+        <div class="correction proposed" id="c2567-3">
             <span class="marker">Proposed Correction</span>
             <a href="https://github.com/WebAudio/web-audio-api/issues/2567">
                 Issue 2567</a> Device-related error reporting via AudioContext
--- a/index.bs
+++ b/index.bs
@@ -12931,7 +12931,7 @@ audio system resource error.
         1. <a spec="dom" lt="fire an event">Fire an event</a> named
             {{AudioContext/error}} at the {{AudioContext}}.
 
-    1. Abort these substeps.
+    1. Abort these steps.
 
 1. If the {{[[rendering thread state]]}} of the {{AudioContext}} is
     <code>suspended</code>.
@@ -12941,9 +12941,6 @@ audio system resource error.
 
         1. <a spec="dom" lt="fire an event">Fire an event</a> named
             {{AudioContext/error}} at the {{AudioContext}}.
-    
-    1. Abort these substeps.
-
 
 Note: An example of system audio resource errors would be when an external or wireless audio
     device becomes disconnected during the active rendering of the AudioContext.
--- a/index.bs
+++ b/index.bs
@@ -2189,13 +2189,14 @@ Attributes</h4>
     <ins cite=#2567>
     : <dfn>onerror</dfn>
     ::
-        An [=event handler=] for any <dfn event>error</dfn> {{Event}} dispatched
-        from an {{AudioContext}}. The user agent can dispatch this event in the following cases:
+        An [=event handler=] for the {{ErrorEvent}} dispatched from an {{AudioContext}}. The event
+            type of this handler is <dfn event for=AudioContext>error</dfn> and the user agent can
+            dispatch this event in the following cases:
 
-            * The process of initializing and activiating the selected audio device encounters an
-                issue.
-            * The audio output from a {{AudioContextState/running}} {{AudioContext}} unexpectedly
-                interrupted or the device is disconnected.
+            * When initializing and activating a selected audio device encounters failures.
+            * When the audio output device associated with an {{AudioContext}} is disconnected while
+                the context is {{AudioContextState/running}}.
+            * When the operating system reports an audio device malfunction.
 
     </ins>
 </div>
--- a/index.bs
+++ b/index.bs
@@ -182,6 +182,7 @@ window.addEventListener('load', (event) => {
   ListAmendments("c2400", "Proposed Addition", "change-list-2400");
   ListAmendments("c2512", "Proposed Addition", "change-list-2512");
   ListAmendments("c2450", "Proposed Addition", "change-list-2450");
+  ListAmendments("c2567", "Proposed Addition", "change-list-2567");
 });
 </script>
 <style>
@@ -1694,8 +1695,8 @@ enum AudioSinkType {
         </a> Add AudioRenderCapacity Interface <br>
         <a href="https://github.com/WebAudio/web-audio-api/issues/2400" id="c2400-2">
             Issue 2400</a> Access to a different output device <br>
-        <a href="https://github.com/WebAudio/web-audio-api/issues/2567" id="c2567-2">
-            Issue 2567</a> Device-related error reporting via AudioContext
+        <a href="https://github.com/WebAudio/web-audio-api/issues/2567" id="c2567-1">
+            Issue 2567</a>. Device-related error reporting via AudioContext
     <div class="amendment-buttons"></div>
     <del cite="#c2444 #c2400 #c2567">
         <xmp class="idl extract" data-no-idl>
@@ -1946,61 +1947,73 @@ Constructors</h4>
         Sending a <a>control message</a> to start processing means
         executing the following steps:
     
-        <div class="addition proposed" id="c2400-5">
-        <span class="marker">Proposed Addition</span><br>
-            <a href="https://github.com/WebAudio/web-audio-api/issues/2400">Issue
-                2400</a> Access to a different output device <br>
-            <a href="https://github.com/WebAudio/web-audio-api/issues/2567" id="c2567-1">
-                Issue 2567</a> Device-related error reporting via AudioContext
-        <div class="amendment-buttons"></div>
-        <del cite="#2400 #2567">
-            1. Attempt to <a href="#acquiring">acquire system resources</a>.
-                In case of failure, abort the following steps.
+        <div class="addition proposed">
+            <div class="addition proposed" id="c2400-5" style="border:0; margin:0; padding:0">
+                <span class="marker">Proposed Addition
+                    <a href="https://github.com/WebAudio/web-audio-api/issues/2400">Issue 2400</a>.
+                </span>
+                Access to a different output device
+                <div class="amendment-buttons"></div>
+            </div>
+            <div class="addition proposed" id="c2567-2" style="border:0; margin:0; padding:0">
+                <span class="marker">Proposed Addition
+                    <a href="https://github.com/WebAudio/web-audio-api/issues/2567">Issue 2567</a>.
+                </span>
+                Device-related error reporting via AudioContext
+                <div class="amendment-buttons"></div>
+            </div>
 
-            3. Set the {{[[rendering thread state]]}} to <code>running</code> on the {{AudioContext}}.
+            <del cite="#2400 #2567">
+                1. Attempt to <a href="#acquiring">acquire system resources</a>.
+                    In case of failure, abort the following steps.
 
-            1. <a href="https://html.spec.whatwg.org/multipage/media.html#queue-a-media-element-task">
-                queue a media element task</a> to execute the following steps:
-
-                1. Set the {{BaseAudioContext/state}} attribute of the {{AudioContext}} to "{{AudioContextState/running}}".
+                3. Set the {{[[rendering thread state]]}} to <code>running</code> on the {{AudioContext}}.
 
                 1. <a href="https://html.spec.whatwg.org/multipage/media.html#queue-a-media-element-task">
-                    queue a media element task</a> to <a spec="dom" lt="fire an event">fire an event
-                    </a> named {{BaseAudioContext/statechange}} at the {{AudioContext}}.
-        </del>
-        <ins cite="#2400 #2567">
-            1. Attempt to <a href="#acquiring">acquire system resources</a> to use a
-                following audio output device based on {{AudioContext/[[sink ID]]}} for
-                rendering:
+                    queue a media element task</a> to execute the following steps:
 
-                * The default audio output device for the empty string.
-                * A audio output device identified by {{AudioContext/[[sink ID]]}}.
-            
-                Upon failure of resource acquisition,
-                    <a href="https://html.spec.whatwg.org/multipage/media.html#queue-a-media-element-task">
-                    queue a media element task</a> to
-                    <a spec="dom" lt="fire an event">fire an event</a> named
-                    {{AudioContext/error}} at the {{AudioContext}}, and abort
-                    the following steps.
+                    1. Set the {{BaseAudioContext/state}} attribute of the {{AudioContext}} to "{{AudioContextState/running}}".
 
-            1. Set the {{[[rendering thread state]]}} to <code>running</code> on the
-                {{AudioContext}}.
+                    1. <a href="https://html.spec.whatwg.org/multipage/media.html#queue-a-media-element-task">
+                        queue a media element task</a> to <a spec="dom" lt="fire an event">fire an event
+                        </a> named {{BaseAudioContext/statechange}} at the {{AudioContext}}.
+            </del>
+            <ins cite="#2400 #2567">
+                1. Attempt to <a href="#acquiring">acquire system resources</a> to use a
+                    following audio output device based on {{AudioContext/[[sink ID]]}} for
+                    rendering:
 
-            1. <a href="https://html.spec.whatwg.org/multipage/media.html#queue-a-media-element-task">
-                Queue a media element task</a> to execute the following steps:
+                    * The default audio output device for the empty string.
+                    * A audio output device identified by {{AudioContext/[[sink ID]]}}.
+                
+                    Upon failure of resource acquisition,
+                        <a href="https://html.spec.whatwg.org/multipage/media.html#queue-a-media-element-task">
+                        queue a media element task</a> to
+                        <a spec="dom" lt="fire an event">fire an event</a> named
+                        {{AudioContext/error}} at the {{AudioContext}}, and abort
+                        the following steps.
 
-                1. Set the {{BaseAudioContext/state}} attribute of the {{AudioContext}}
-                    to "{{AudioContextState/running}}".
+                1. Set the {{[[rendering thread state]]}} to <code>running</code> on the
+                    {{AudioContext}}.
 
-                1. <a spec="dom" lt="fire an event">Fire an event</a> named
-                    {{BaseAudioContext/statechange}} at the {{AudioContext}}.
-        </ins>
+                1. <a href="https://html.spec.whatwg.org/multipage/media.html#queue-a-media-element-task">
+                    Queue a media element task</a> to execute the following steps:
+
+                    1. Set the {{BaseAudioContext/state}} attribute of the {{AudioContext}}
+                        to "{{AudioContextState/running}}".
+
+                    1. <a spec="dom" lt="fire an event">Fire an event</a> named
+                        {{BaseAudioContext/statechange}} at the {{AudioContext}}.
+            </ins>
+
+        </div>
     </div>
 
-        <div class="correction proposed" id="c2567-3">
-            <span class="marker">Proposed Correction</span>
-            <a href="https://github.com/WebAudio/web-audio-api/issues/2567">
-                Issue 2567</a> Device-related error reporting via AudioContext
+        <div class="proposed addition" id="c2567-3">
+            <span class="marker">Proposed Addition
+                <a href="https://github.com/WebAudio/web-audio-api/issues/2567">Issue 2567</a>.
+            </span>
+            Device-related error reporting via AudioContext
             <div class="amendment-buttons"></div>
             <del>
             Note: It is unfortunately not possible to programatically notify authors
@@ -2150,10 +2163,9 @@ Attributes</h4>
     </ins>
 </div>
 
-<div class="addition proposed" id="c2567-4">
+<div class="proposed addition" id="c2567-4">
     <span class="marker">Proposed Addition
-        <a href="https://github.com/WebAudio/web-audio-api/issues/2567">
-            Issue 2567</a>.
+        <a href="https://github.com/WebAudio/web-audio-api/issues/2567">Issue 2567</a>.
     </span>
     Device-related error reporting via AudioContext
     <div class="amendment-buttons"></div>
--- a/index.bs
+++ b/index.bs
@@ -2004,16 +2004,23 @@ Constructors</h4>
 
             </del>
             <ins cite="#2400 #2567">
+                1. Let |document| be the [=current settings object=]'s [=relevant global object=]'s
+                        [=associated Document=].
+
                 1. Attempt to [=acquire system resources=] to use a following audio output device
                     based on {{AudioContext/[[sink ID]]}} for rendering:
 
                     * The default audio output device for the empty string.
                     * A audio output device identified by {{AudioContext/[[sink ID]]}}.
                 
-                    Upon failure of resource acquisition,
-                        [=queue a media element task=] to [=fire an event=] using {{ErrorEvent}}
-                        at the {{AudioContext/error}} of the {{AudioContext}}, and abort the
-                        following steps.
+                    1. If resource acquisition fails, execute the following steps:
+                        
+                        1. If |document| is not allowed to use the feature identified by
+                            <code>"speaker-selection"</code>, abort these substeps.
+
+                        1. [=queue a media element task=] to [=fire an event=] using {{ErrorEvent}}
+                            at the {{AudioContext/error}} of the {{AudioContext}}, and abort the
+                            following steps.
 
                 1. Set [=this=] {{[[rendering thread state]]}} to <code>running</code> on the
                     {{AudioContext}}.
--- a/index.bs
+++ b/index.bs
@@ -2145,7 +2145,7 @@ Attributes</h4>
         following cases:
 
             * The process of initializing and activiating the selected audio
-                devie encounters an issue.
+                device encounters an issue.
             * The audio output from a {{AudioContextState/running}}
                 {{AudioContext}} unexpectedly interrupted or the device is
                 disconnected.
@@ -12870,6 +12870,34 @@ running the algorithm for this {{AudioNode}} to produce
 running the algorithm for an {{AudioNode}}, using an <a>input
 buffer</a> and the value(s) of the {{AudioParam}}(s) of this
 {{AudioNode}} as the input for this algorithm.
+
+
+<h3 id="error-handling-on-a-running-audio-context">
+    Error Handling for System Audio Resources on a running AudioContext</h3>
+
+In the event of an audio system resource error, the <a>rendering thread</a>
+performs the following steps upon receiving the error:
+
+1. Set {{[[rendering thread state]]}} of the {{AudioContext}} to
+    "{{AudioContextState/suspended}}".
+
+1. <a href="https://html.spec.whatwg.org/multipage/media.html#queue-a-media-element-task">
+    Queue a media element task</a> to execute the following steps:
+
+    1. <a spec="dom" lt="fire an event">fire an event</a> named
+        {{AudioContext/error}} at the associated {{AudioContext}}.
+    
+    1. If the {{BaseAudioContext/state}} attribute of the {{AudioContext}} is
+        not already "{{AudioContextState/suspended}}":
+
+        1. Set the {{BaseAudioContext/state}} attribute of the {{AudioContext}}
+            to "{{AudioContextState/suspended}}".
+
+        1. <a spec="dom" lt="fire an event">Fire an event</a> named
+            {{BaseAudioContext/statechange}} at the associated {{AudioContext}}.
+
+Note: An example of such a case would be when an external or wireless audio
+    device becomes disconnected during the active rendering of the AudioContext.
 
 <h3 id="unloading-a-document">Unloading a document</h3>
     Additional <a href=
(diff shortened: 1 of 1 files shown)


hoch:
@padenot Ready for the first pass. Let me know what you think!

gsinafirooz:
Hi @padenot; I just wanted to bump this.

index.bs
```diff
@@ -2121,7 +2128,32 @@ Attributes</h4>
...
-    </dl>
+</div>
+
+<div class="addition proposed" id="c2567-3">
+    <span class="marker">Proposed Addition
+        <a href="https://github.com/WebAudio/web-audio-api/issues/2567">
+            Issue 2567</a>.
+    </span>
+     Device-related error reporting via AudioContext
+    <div class="amendment-buttons"></div>
```
Korilakkuma, line 2139: [nits] It seems to render empty `HTMLDivElement`. Maybe, I think that it is better to render buttons as expected HTML. Or, if unnecessary buttons, it would be better to delete this `HTMLDivElement`.

index.bs
```diff
@@ -12839,6 +12871,34 @@ running the algorithm for an {{AudioNode}}, using an <a>input
...
+        not already "{{AudioContextState/suspended}}":
+
+        1. Set the {{BaseAudioContext/state}} attribute of the {{AudioContext}}
+            to "{{AudioContextState/suspended}}".
+
+        1. <a spec="dom" lt="fire an event">Fire an event</a> named
+            {{BaseAudioContext/statechange}} at the associated {{AudioContext}}.
+
+Note: An example of such a case would be when an external or wireless audio
+    device becomes disconnected during the active rendering of the AudioContext.
```
padenot, line 12900: Do we want to explicitly mention the fallbacks that we've all implemented here?

index.bs
```diff
@@ -12839,6 +12871,34 @@ running the algorithm for an {{AudioNode}}, using an <a>input
...
+performs the following steps upon receiving the error:
+
+1. Set {{[[rendering thread state]]}} of the {{AudioContext}} to
+    "{{AudioContextState/suspended}}".
+
+1. <a href="https://html.spec.whatwg.org/multipage/media.html#queue-a-media-element-task">
+    Queue a media element task</a> to execute the following steps:
+
+    1. <a spec="dom" lt="fire an event">fire an event</a> named
+        {{AudioContext/error}} at the associated {{AudioContext}}.
```
gsinafirooz, line 12888: onerror?

gsinafirooz (changes requested):

index.bs
padenot: The event is called `"error"`. This in turns call `onerror` on the context via normal even `EventTarget` rules.

hoch: @gsinafirooz If the comment/question has more detailed, it would be more helpful to others.

index.bs
hoch: Can you elaborate a bit more? Do you mean the fallback to the default device when no sinkID is given?

index.bs
```diff
@@ -12839,6 +12871,34 @@ running the algorithm for an {{AudioNode}}, using an <a>input
...
+    
+    1. If the {{BaseAudioContext/state}} attribute of the {{AudioContext}} is
+        not already "{{AudioContextState/suspended}}":
```
gsinafirooz, line 12891: Per our brief discussion; I suggest changing this line to:

`"{{AudioContextState/running}}":`

gsinafirooz (changes requested):

index.bs
padenot: I meant that when no devices are present on the system, as things stand today, Firefox (and I think Chrome, but I could have misheard) will run the graph on a regular thread, periodically attempting to grab an audio device should one appear.

hoch: SG. I'll add that to the note.

index.bs
hoch: Correct.

That brings another question: do we want to fire `onerror` even when the context is closed? What does that mean and how's that useful to developers?

I actually prefer not to do anything when the context is closed. The closed context should be of no use.

To summarize:
- If the context is suspended, don't change anything and just fire `onerror`.
- If the context is running, suspend the context and then fire `onerror`.
- if the context is closed, just return.

And yet another question - if the app tries to resume the suspended context (due to `onerror`) what should happen? Also do we have to change the previous sinkId (e.g. to a default device) because it was a problematic device?

cc @padenot

chrisguttandin: Does that mean the `'error'` event will replace the `'sinkchange'` event proposed here (https://github.com/WebAudio/web-audio-api/issues/2532#issuecomment-1717151574)?

hoch: @chrisguttandin If the group's answer to my last question above is yes, then yes.

statechange (suspended) -> error -> sinkchange

The flow makes sense, but perhaps too many events are triggered?

gsinafirooz: If a developer wants to use a non-default sink, they should know what they are doing. They should know it may fail. I'd rather leave the decision of switching to a different sink (default or else) to the developers.

index.bs
hoch: I gave more thoughts on this and the first part makes sense (`will run the graph on a regular thread`). However, the second part seems difficult to spec and expect consistent behavior between different implementations.

How does it "attempt" and how "periodic" should it be?

hoch: By the way, I took a stab on this at L2012 above.

index.bs
Korilakkuma: > The buttons will be rendered automatically once the page is fully rendered.

Sorry ... I didn't know that.
Thanks for your review.

hoch: No - thanks so much for your interest here. The Web Audio API specification is quite messy and more eyes on a PR is always better!

index.bs
padenot: I have no idea, but I can tell you what Firefox does.

If it cannot open a system level audio stream, it's going to attempt in 10ms. If that fails, it's going to backoff a bit and retry. The backoff progression is exponential but is capped at trying once a second in the end. While the retries are happening, the web audio rendering graph runs with a quantum of 10ms at the graph sample-rate, on a regular thread (not an audio IO thread).

index.bs
padenot: I'd like to not have `"error"` fire in nominal scenario. `"statechange"` is for state changes that aren't an error.

index.bs
hoch: This issue is fixed now.

index.bs
hoch: I see. 10ms is rather quite fast for retry! In Chromium, we simply spawn up a thread to emulate the device callback. The automatic retry seems like a a nice service, but I don't think we need to include in our spec language.

IMO, as a common ground what I have above makes sense: "the User-Agent must attempt to silently render the audio graph using a mechanism that emulates an audio output device."

WDYT?

index.bs
hoch: I am okay with that. To summarize:

When the running AudioContext encounters an error from the renderer, it suspends the context (but with no statechange event) and fires an error event.

WDYT?

index.bs
padenot: Yeah, we've found that trying very often and then backing off ended up with a good UX. Your phrasing sounds good to me.

index.bs
padenot: We usually pair handlers with an attribute. We could do the same as `HTMLMediaElement`, that's fairly flexible. It's probably that `setSinkId` + `resume` or just `setSinkId` would allow restarting a failed context.

index.bs
hoch: Great. Resolving.

index.bs
mjwilson-google: Regarding not firing `statechang` on an error:

At least the current [wording for onStateChange](https://webaudio.github.io/web-audio-api/#dom-baseaudiocontext-onstatechange) should be changed as well in this case. Maybe adding a note that the `onStateChange` handler won't be notified of state changes that are the result of an error?  Or perhaps say that the `onStateChange` handler is only notified in response to requested state changes or automatic playing?

I feel like it's probably simpler for implementers and developers to always fire `statechange` when the state changes. Are there use cases that become more difficult or impossible if we do it the simpler way?

chrisguttandin: I'm not sure if we should blindly copy the `MediaRecorder` but it follows a different pattern. I always fires an `'error'` event followed by a `'stop'` event. The `'stop'` event would also fire when the `MediaRecorder` gets stopped without an error.

It's defined in step 15.3., step 15.4., and 15.5. of this algorithm: https://w3c.github.io/mediacapture-record/#dom-mediarecorder-start.

marcoscaceres:
@jernoble, could you take a look at this also by chance?

index.bs
hoch: Two comments above support firing both events. The benefits are a) it's consistent, b) easier to understand/implement, and b) developers can easily filter out events they do not need.

@padenot What's the benefit of not firing `statechange` upon the suspension caused by the system audio error?

padenot: I think it's fine, in restrospect.

hoch: Teleconf May02: Developer opinions are positive on both events being fired.

hoch: @padenot Could you clarify your last comment? Are we supportive of firing both events (onerror & onstatechange) upon a system audio failure?

padenot: Yes, both events.

hoch: Great. Thanks for confirmation. Resolving.

hoch:
@padenot It looks like we resolved all discussions so far. Let me know if you have more questions!

gsinafirooz:
> @jernoble, could you take a look at this also by chance?

@jernoble, We'd like to hear your opinion :)

padenot (approved):

hoch:
@jernoble @marcoscaceres Can any of you comment on this pull request? We've been waiting for three weeks, and would like to merge this PR so that we can make progress on the browser implementation.

The PR itself is not really controversial; we want to add an event handler for audio device/platform errors.

index.bs
```diff
@@ -1943,53 +1947,88 @@ Constructors</h4>
...
+            </div>
 
-            1. <a href="https://html.spec.whatwg.org/multipage/media.html#queue-a-media-element-task">
-                queue a media element task</a> to execute the following steps:
+            <del cite="#2400 #2567">
+                1. Attempt to <a href="#acquiring">acquire system resources</a>.
+                    In case of failure, abort the following steps.
 
-                1. Set the {{BaseAudioContext/state}} attribute of the {{AudioContext}} to "{{AudioContextState/running}}".
+                3. Set the {{[[rendering thread state]]}} to <code>running</code> on the {{AudioContext}}.
```
marcoscaceres, line 1970:
```suggestion
                1. Set the {{[[rendering thread state]]}} to <code>running</code> on the {{AudioContext}}.
```

index.bs
```diff
@@ -1943,53 +1947,88 @@ Constructors</h4>
...
-                * A audio output device identified by {{AudioContext/[[sink ID]]}}.
-            
-                In case of failure, abort the following steps.
+                    1. Set the {{BaseAudioContext/state}} attribute of the {{AudioContext}} to "{{AudioContextState/running}}".
 
-            1. Set the {{[[rendering thread state]]}} to <code>running</code> on the
-                {{AudioContext}}.
+                    1. <a href="https://html.spec.whatwg.org/multipage/media.html#queue-a-media-element-task">
+                        queue a media element task</a> to <a spec="dom" lt="fire an event">fire an event
+                        </a> named {{BaseAudioContext/statechange}} at the {{AudioContext}}.
```
marcoscaceres, line 1979:
```suggestion
                    1.  [=queue a media element task=] to <a spec="dom" lt="fire an event">fire an event
                        </a> named {{BaseAudioContext/statechange}} at the {{AudioContext}}.
```

index.bs
```diff
@@ -1943,53 +1947,88 @@ Constructors</h4>
...
-                Queue a media element task</a> to execute the following steps:
+                    * The default audio output device for the empty string.
+                    * A audio output device identified by {{AudioContext/[[sink ID]]}}.
+                
+                    Upon failure of resource acquisition,
+                        <a href="https://html.spec.whatwg.org/multipage/media.html#queue-a-media-element-task">
+                        queue a media element task</a> to
+                        <a spec="dom" lt="fire an event">fire an event</a> named
+                        {{AudioContext/error}} at the {{AudioContext}}, and abort
+                        the following steps.
```
marcoscaceres, line 1994:
```suggestion                
                        [=queue a media element task=] to
                        [=fire an event=] named
                        {{AudioContext/error}} at the {{AudioContext}}, and abort
                        the following steps.
```

index.bs
```diff
@@ -1943,53 +1947,88 @@ Constructors</h4>
...
+                    1. Set the {{BaseAudioContext/state}} attribute of the {{AudioContext}} to "{{AudioContextState/running}}".
 
-            1. Set the {{[[rendering thread state]]}} to <code>running</code> on the
-                {{AudioContext}}.
+                    1. <a href="https://html.spec.whatwg.org/multipage/media.html#queue-a-media-element-task">
+                        queue a media element task</a> to <a spec="dom" lt="fire an event">fire an event
+                        </a> named {{BaseAudioContext/statechange}} at the {{AudioContext}}.
+            </del>
+            <ins cite="#2400 #2567">
+                1. Attempt to <a href="#acquiring">acquire system resources</a> to use a
```
marcoscaceres, line 1982:
```suggestion
                1. Attempt to [=acquire system resources=] to use a
```

index.bs
```diff
@@ -1943,53 +1947,88 @@ Constructors</h4>
...
-                In case of failure, abort the following steps.
-
-            3. Set the {{[[rendering thread state]]}} to <code>running</code> on the {{AudioContext}}.
+        <div class="addition proposed">
+            <div class="addition proposed" id="c2400-5" style="border:0; margin:0; padding:0">
+                <span class="marker">Proposed Addition
+                    <a href="https://github.com/WebAudio/web-audio-api/issues/2400">Issue 2400</a>.
+                </span>
+                Access to a different output device
+                <div class="amendment-buttons"></div>
```
marcoscaceres, line 1956: do you need these? don't these get added by the fixup.js script?

index.bs
```diff
@@ -1943,53 +1947,88 @@ Constructors</h4>
...
-            <a href="https://github.com/WebAudio/web-audio-api/issues/2400">Issue
-                2400</a> Access to a different output device
-        <div class="amendment-buttons"></div>
-        <del cite=#2400>
-            1. Attempt to <a href="#acquiring">acquire system resources</a>.
-                In case of failure, abort the following steps.
-
-            3. Set the {{[[rendering thread state]]}} to <code>running</code> on the {{AudioContext}}.
+        <div class="addition proposed">
+            <div class="addition proposed" id="c2400-5" style="border:0; margin:0; padding:0">
```
marcoscaceres, line 1951: See
Seems odd to have nested divs here?  
```suggestion
        <div class="addition proposed">
            <div class="addition proposed" id="c2400-5" style="border:0; margin:0; padding:0">
```

index.bs
marcoscaceres, line 1970: Will the number there be weird? like 1..3?

index.bs
```diff
@@ -1943,53 +1947,88 @@ Constructors</h4>
...
+                    Upon failure of resource acquisition,
+                        <a href="https://html.spec.whatwg.org/multipage/media.html#queue-a-media-element-task">
+                        queue a media element task</a> to
+                        <a spec="dom" lt="fire an event">fire an event</a> named
+                        {{AudioContext/error}} at the {{AudioContext}}, and abort
+                        the following steps.
 
-                1. Set the {{BaseAudioContext/state}} attribute of the {{AudioContext}}
-                    to "{{AudioContextState/running}}".
+                1. Set the {{[[rendering thread state]]}} to <code>running</code> on the
```
marcoscaceres, line 1996: Shouldn't this say (same with the others... internal slots are for instances, but this never talks about what "this" is): 

```suggestion
                1. Set [=this=] {{[[rendering thread state]]}} to <code>running</code> on the
```

index.bs
```diff
@@ -1943,53 +1947,88 @@ Constructors</h4>
...
+                    {{AudioContext}}.
 
-                1. <a spec="dom" lt="fire an event">Fire an event</a> named
-                    {{BaseAudioContext/statechange}} at the {{AudioContext}}.
-        </ins>
+                1. <a href="https://html.spec.whatwg.org/multipage/media.html#queue-a-media-element-task">
+                    Queue a media element task</a> to execute the following steps:
```
marcoscaceres, line 2000:
```suggestion
                1. [=Queue a media element task=] to execute the following steps:
```

index.bs
```diff
@@ -1943,53 +1947,88 @@ Constructors</h4>
...
+            Device-related error reporting via AudioContext
+            <div class="amendment-buttons"></div>
+            <del>
             Note: It is unfortunately not possible to programatically notify authors
             that the creation of the {{AudioContext}} failed. User-Agents are
             encouraged to log an informative message if they have access to a logging
             mechanism, such as a developer tools console.
+            </del>
+            <ins>
+            NOTE: In cases where an AudioContext is constructed with no
```
marcoscaceres, line 2025:
```suggestion
            NOTE: In cases where an {{AudioContext}} is constructed with no
```

index.bs
```diff
@@ -1943,53 +1947,88 @@ Constructors</h4>
...
+            arguments and resource acquisition fails, the User-Agent must 
+            attempt to silently render the audio graph using a mechanism that
+            emulates an audio output device.
```
marcoscaceres, line 2028:
```suggestion
            arguments and resource acquisition fails, the User-Agent will 
            attempt to silently render the audio graph using a mechanism that
            emulates an audio output device.
```

index.bs
```diff
@@ -2121,7 +2158,31 @@ Attributes</h4>
...
+<div class="proposed addition" id="c2567-4">
+    <span class="marker">Proposed Addition
+        <a href="https://github.com/WebAudio/web-audio-api/issues/2567">Issue 2567</a>.
+    </span>
+    Device-related error reporting via AudioContext
+    <div class="amendment-buttons"></div>
+    <ins cite=#2567>
+    : <dfn>onerror</dfn>
+    ::
+        An [=event handler=] for any <dfn event>error</dfn> {{Event}} dispatched from an
```
marcoscaceres, line 2172: I don't think we want to redefine error... don't you want HTMLMediaElement/error?  

```suggestion
        An [=event handler=] for any <a event for="HTMLMediaElement">error</a> {{Event}} dispatched from an
```

marcoscaceres: Or at least, it should be defined as a AudioContext error.

index.bs
```diff
@@ -12839,6 +12900,62 @@ running the algorithm for an {{AudioNode}}, using an <a>input
...
+<div class="addition proposed" id="c2567-5">
+    <span class="marker">Proposed addition
+        <a href="https://github.com/WebAudio/web-audio-api/issues/2567">Issue 2567</a>.
+    </span>
+    Device-related error reporting via AudioContext
+    <div class="amendment-buttons"></div>
+    <ins>
+
+<h3 id="error-handling-on-a-running-audio-context">
+    Handling an error from System Audio Resources on the {{AudioContext}}</h3>
```
marcoscaceres, line 12913:
```suggestion
    Handling an error from System Audio Resources on the and {{AudioContext}} |audioContext|</h3>
```

index.bs
```diff
@@ -12839,6 +12900,62 @@ running the algorithm for an {{AudioNode}}, using an <a>input
...
+
+The <a>rendering thread</a> performs the following steps in the event of an
+audio system resource error.
+
+1. If the {{[[rendering thread state]]}} of the {{AudioContext}} is
```
marcoscaceres, line 12918: I think you want something like this: 

```suggestion
1. If the |audioContext|'s {{[[rendering thread state]]}} is
```

index.bs
```diff
@@ -12839,6 +12900,62 @@ running the algorithm for an {{AudioNode}}, using an <a>input
...
+    1. Abort these steps.
+
+1. If the {{[[rendering thread state]]}} of the {{AudioContext}} is
+    <code>suspended</code>.
+
+    1. <a href="https://html.spec.whatwg.org/multipage/media.html#queue-a-media-element-task">
+        Queue a media element task</a> to execute the following steps:
+
+        1. <a spec="dom" lt="fire an event">Fire an event</a> named {{AudioContext/error}} at the
+            {{AudioContext}}.
```
marcoscaceres, line 12951:
```suggestion
    1. [=Queue a media element task=] to execute the following steps:

        1. [=Fire an event=] named {{AudioContext/error}} at |audioContext|.
```

index.bs
```diff
@@ -12839,6 +12900,62 @@ running the algorithm for an {{AudioNode}}, using an <a>input
...
+
+Note: An example of system audio resource errors would be when an external or wireless audio
+    device becomes disconnected during the active rendering of the AudioContext.
```
marcoscaceres, line 12954:
```suggestion
    device becoming disconnected during the active rendering of the {{AudioContext}}.
```

jernoble:
> > @jernoble, could you take a look at this also by chance?
> 
> @jernoble, We'd like to hear your opinion :)

As a comment about the general capability added here, I see no problem with exposing AudioContext error states. This PR appears to be narrowly around handling what happens when the AudioContext loses exclusive access to the current audio sink / device. That either because that device is no longer available _permanently_ (a plug was pulled, a wireless device disappeared, etc.) or because that device is no longer available _temporarily_ (another process / service received exclusive access to that device, etc.). And there's no way to distinguish between those two states here.

In the "permanent" error case, there's nothing for the application to do save choose a different sink. In the "temporary" error case, this is recoverable by simply un-suspending.

Is there some other mechanism available to distinguish these two states? If not, should the `onerror` handler take a custom `Error` object with a payload?

hoch:
> that device is no longer available temporarily (another process / service received exclusive access to that device, etc.). And there's no way to distinguish between those two states here.

This is somewhat related `interrupted' status from the [AudioSession](https://github.com/w3c/audio-session/blob/main/explainer.md). Also I am a bit confused by the expression "no longer available" and "temporarily" in one sentence. We have to consider two cases separately.

- If it's **temporarily unavailable** (as in `interrupted`) it is not an error. It's an intentional state change initiated by the platform.
- If it's **permanently unavailable**, that's an error and should be reported by `onerror`.

index.bs
hoch: Done.

index.bs
hoch: Done.

index.bs
hoch: Done.

index.bs
hoch: Certain sections needed multiple "addition proposed" entries. This is the way to make it happen. (This is a new style/method that can only be seen in Web Audio API spec)

cc @svgeesus

index.bs
hoch: Fixed, but the numbering is automatically handled by Bikeshed.

index.bs
hoch: Thanks for the suggestion! Done.

index.bs
hoch: Done.

index.bs
hoch: Done.

index.bs
hoch: Thanks! Done.

index.bs
hoch: I am not sure what this suggestion means. Can you clarify?

index.bs
hoch: Done.

index.bs
hoch: Done.

index.bs
hoch: Done.

index.bs
hoch: This needs a bit more work to define `error` under AudioContext. I'll look for the HTMLMediaElement error spec for examples.

index.bs
hoch: I've been following other existing examples of this spec. As far as I can tell, I don't see any reference to `fixup.js` in the spec source file. Perhaps I am missing something?

index.bs
hoch: I am assuming `and` is typo, and `{{AudioContext}}|audioContext|` means that "audioContext, an instance of AudioContext interface".

index.bs
marcoscaceres: Bikeshed adds fixup.js when generating the static html file. Just confirm that the purple boxes are all ok (and if possible, see if removing the div makes any difference).

index.bs
marcoscaceres: Oh, ok! that's super interesting... cc'ing @fantasai, as we are looking at how people use all this.

index.bs
marcoscaceres: yes, sorry! oops sorry about the typo! it's to match styles of other specs. For example:

https://www.w3.org/TR/geolocation/#request-a-position

jernoble:
> > that device is no longer available temporarily (another process / service received exclusive access to that device, etc.). And there's no way to distinguish between those two states here.
> 
> This is somewhat related `interrupted' status from the [AudioSession](https://github.com/w3c/audio-session/blob/main/explainer.md). Also I am a bit confused by the expression "no longer available" and "temporarily" in one sentence. We have to consider two cases separately.

That was inelegantly phrased on my part. :)

I was referring to this new language:

> * The audio output from a [running](https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/pull/2580.html#dom-audiocontextstate-running) [AudioContext](https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/pull/2580.html#AudioContext) unexpectedly interrupted or the device is disconnected.

What does "unexpectedly interrupted" mean here?

index.bs
hoch: With this div, the button shows up:

Screenshot 2024-05-31 at 8 40 25 PM


When I remove the div, the button disappears:

Screenshot 2024-05-31 at 8 39 09 PM

index.bs
hoch: Having a variable within the title looks a bit weird, so I rephrase the prose to add `|audioContext|` variable in there.

hoch:
> What does "unexpectedly interrupted" mean here?

Yes, now I can see why you used the word "interrupted".  Perhaps I can intentionally move away from that language so we can reserve such case for AudioSession API. My example of "unexpectedly interrupted" cases would be any kind of audio infra malfunction reported by the operating system including (but not limited to):
  - A wireless device getting disconnected (e.g. too far away from the host machine)
  - An external audio device getting disconnected (e.g. USB cable plugged out)
 
IMO one significant distinction from the AudioSession is:

> interrupted: the AudioSession is not playing sound, but can resume when it will get uninterrupted.

The interrupted status can be resumed automatically after the interruption (please correct me if I am mistaken). So I think this can be useful for situations like notification, phone call on mobile device, etc.

index.bs
hoch: Feel free to reopen if this is an issue! I would love to provide more feedback on this "proposed addition" markup.

hoch:
@marcoscaceres @jernoble If you don't have other outstanding concerns, can we land this PR?

Also if you would like to have deeper technical discussion on this topic, please consider joining our bi-weekly teleconference. :)

index.bs
```diff
@@ -1968,53 +1972,90 @@ Constructors</h4>
...
-                1. <a spec="dom" lt="fire an event">Fire an event</a> named
-                    {{BaseAudioContext/statechange}} at the {{AudioContext}}.
-        </ins>
+                    * The default audio output device for the empty string.
+                    * A audio output device identified by {{AudioContext/[[sink ID]]}}.
+                
+                    1. If resource acquisition fails, execute the following steps:
+                        
+                        1. If |document| is not allowed to use the feature identified by
+                            <code>"speaker-selection"</code>, abort these substeps.
```
hoch, line 2019: @padenot Please take a look at this step.

Per Chromium's privacy discussion, this step is necessary to avoid brute-forcing UUID in a drive-by fashion. If this looks good to you, I'll merge this PR.

padenot: Sure thing, looks good.

hoch:
Thanks @padenot! Merging.

GorbachovMaina1 (commented):
```
Tut mir leid aber ich kenne mich nicht aus das funktioniert ich bin bitte - [ ] weiterhelfen damit
```

GorbachovMaina1:
> > Kannst du mir bitte auf Deutsch oder bulgarisch schreiben damit ich verstehe und was es dir ansonsten ist mir schwierig weil ich verstehe leider keine Englisch ich bedanke mich voraus

PR #2629, opened 2025-03-19
Throw onerror instead of exception for AudioContext constructor

### LINKED ISSUE #2435: Use algorithm to throw exception when constructing Web Audio API objects, where it makes sense.
Related to WebAudio/web-audio-api#1805, we should probably throw the exceptions in an algorithm and not the Option object, to make it clear that there is an order, and that the throwing the exception happens when the ctor or factory method is being called, and not when setting the member on the option object, or constructing the option object from an object literal.

mjwilson-google:
This affects #2435, but may not fix everything in that issue.

We discussed at TPAC 2024 that, because device enumeration is asynchronous, it will be easier for implementers and also more robust to use onerror to signal problems with the device at construction time.

This PR does the following:
- Move the sink validation to before system resources are acquired
- Change the sink validation algorithm to return boolean values
- Update both uses of the sink validation algorithm to use the new return value

***
<a href="https://pr-preview.s3.amazonaws.com/mjwilson-google/web-audio-api/pull/2629.html" title="Last updated on Mar 20, 2025, 10:41 PM UTC (00a23b0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/2629/0be282c...mjwilson-google:00a23b0.html" title="Last updated on Mar 20, 2025, 10:41 PM UTC (00a23b0)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1591,13 +1591,6 @@ Constructors</h4>
                         {{AudioSinkOptions/type}} in |sinkId| and {{AudioSinkInfo/type}}
                         in {{AudioContext/[[sink ID]]}} are equal, abort these substeps.
 
-                    1. Let |validationResult| be the return value of
-                        <a href="#validating-sink-identifier">sink identifier validation
-                        </a> of |sinkId|.
-    
-                    1. If |validationResult| is a type of {{DOMException}}, throw an
-                        exception with |validationResult| and abort these substeps.
-
                     1. If |sinkId| is a type of {{DOMString}}, set
                         {{AudioContext/[[sink ID]]}} to |sinkId| and abort these
                         substeps.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1643,14 +1636,22 @@ Constructors</h4>
         1. Let |document| be the [=current settings object=]'s [=relevant global object=]'s
                 [=associated Document=].
 
+        1. Let |validationResult| be the return value of
+                <a href="#validating-sink-identifier">sink identifier validation
+                </a> of {{AudioContext/[[sink ID]]}}.
+
+        1. If |validationResult| is not <code>true</code>, [=Queue a
```
hoch, line 1643: `not <code>true</code>` 

to

`<code>false</code>`

index.bs
```diff
@@ -2171,9 +2172,10 @@ Methods</h4>
             1. Let |validationResult| be the return value of
                 <a href="#validating-sink-identifier">sink identifier validation</a>
                 of |sinkId|.
-            
-            1. If |validationResult| is not <code>null</code>, return a promise
-                rejected with |validationResult|. Abort these steps.
+
+            1. If |validationResult| is not <code>true</code>, return a promise
```
hoch, line 2176: Ditto for `not true`.

hoch (commented):
I think this is what we discussed offline, but I am curious if there are places we might need to spot check,

index.bs
mjwilson-google: Done, thanks!

index.bs
mjwilson-google: Done.

index.bs
```diff
@@ -1643,14 +1636,22 @@ Constructors</h4>
         1. Let |document| be the [=current settings object=]'s [=relevant global object=]'s
                 [=associated Document=].
 
+        1. Let |validationResult| be the return value of
+                <a href="#validating-sink-identifier">sink identifier validation
+                </a> of {{AudioContext/[[sink ID]]}}.
+
+        1. If |validationResult| is <code>false</code>, [=Queue a
```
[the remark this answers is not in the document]
mjwilson-google: Good catch, thank you.  Fixed.  The auto-linking seems to still work.

mjwilson-google:
After the working group meeting this PR needs some rework, I will comment again when it is ready for review.

mjwilson-google:
I think the current patch covers what we talked about at the meeting:
- Add private internal slot for sinkid, to hold the constructor parameter until we are sure the sinkid is usable
- Then we can update the javascript-visible slot with the valid ID once it has been opened
- If validation fails, fall back to default device, fire onerror, and do not transition to running
- For acquisition failure, always fire onerror, and we can remove the speaker-selection step since it will be handled in the validation step

I'm not sure about the naming of the private slot, and we also have two things called sending a control message to start processing (one in the constructor and one for setsinkid) but hopefully it's clear from context which one is which.

hoch (approved):

PR #2641, opened 2025-08-15
Specify DynamicsCompressorNode makeup gain control

### LINKED ISSUE #2639: Add adjustable makeup gain to DynamicsCompressorNode
**Describe the feature**
The `DynamicsCompressorNode` specification currently has a fixed makeup gain derived from its other parameters: https://webaudio.github.io/web-audio-api/#computing-the-makeup-gain

While automatic makeup gain is convenient in some use cases, some users of the Web Audio API want more control over the makeup gain.  For instance see discussions here:
- https://github.com/Tonejs/Tone.js/issues/1353
- https://crbug.com/372198644

We could add a way for developers to control the makeup gain behavior.

**Is there a prototype?**
No.

**Describe the feature in more detail**
I propose allowing users to control makeup gain directly in `DynamicsCompressorNode` by adding the following:
- An `AudioParam`, `makeupGain`, which represents the makeup gain value in dB
- An addition to `DynamicsCompressorOptions`, `makeupGain`, of type `(float or MakeupGainOptions)`, where `MakeupGainOptions` can be an `enum` with the value `"automatic"`
  - If `makupGain` is `"automatic"`, the default, then the current behavior in the spec is used
  - If `makeupGain` is a `float`, then that is the fixed makeup gain in dB (with 0 meaning no makeup gain)

I think the basic idea of the above would work well, but we can also change the details or discuss alternative approaches.

mjwilson-google:
This fixes #2639.

I picked the parameter bounds to be the same as for the filter node gain.

I'm not sure if this is the best way to specify the behavior:
- Is there a better way to specify that the parameter value can change than the current `default-notes`?
- Is it ok to reference an AudioParam directly in an algorithm step?

***
<a href="https://pr-preview.s3.amazonaws.com/mjwilson-google/web-audio-api/pull/2641.html" title="Last updated on Aug 15, 2025, 11:03 PM UTC (4057738)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/2641/7f58ac7...mjwilson-google:4057738.html" title="Last updated on Aug 15, 2025, 11:03 PM UTC (4057738)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -7894,9 +7894,9 @@ Dictionary {{DynamicsCompressorOptions}} Members</h5>
 
     : <dfn>makeupGain</dfn>
     :: The initial value for the {{DynamicsCompressorNode/makeupGain}}
-       AudioParam. If set to {{MakeupGainOptions/automatic}}, the
-       makeup gain will be calculated as described in [=Computing the
-       makeup gain=].
+         AudioParam. If set to {{MakeupGainOptions/automatic}}, the
+         makeup gain will be calculated as described in [=Computing the
+         makeup gain=].
 </dl>
 
 <h4 id="DynamicsCompressorOptions-processing">
--- a/index.bs
+++ b/index.bs
@@ -7689,6 +7689,27 @@ macros:
 </pre>
 
 <pre class="idl">
+enum MakeupGainOptions {
+    "automatic"
+};
+</pre>
+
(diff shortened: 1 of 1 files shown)


hoch (approved):
LGTM

padenot (approved):

PR #2666, opened 2026-01-16
Use <xmp> for examples

padenot:
This is more comfortable to write e.g. comparisons (< instead of `&lt;`
in the code).

This fixes https://github.com/WebAudio/web-audio-api/issues/2429.

I found that while inspecting ready for editing issues.


hoch (approved):
Not sure if it's ready for review, but you have my LGTM.

Issue #2663, opened 2026-01-08
Add renderSizeHint handling to AudioContext constructor


mjwilson-google:
Currently, there is specific text about renderSizeHint in the OfflineAudioContext constructor here: https://webaudio.github.io/web-audio-api/#OfflineAudioContext-constructors

However, text to handle this is missing from the AudioContext constructor here: https://webaudio.github.io/web-audio-api/#AudioContext-constructors

We should add text to AudioContext constructor as well.

PR #2667, opened 2026-01-21
Specify render quantum size range

### LINKED ISSUE #2664: Constrain valid renderSizeHint to inclusive range [1 sample, 6 seconds]
As discussed at TPAC 2025, we will accept renderSizeHints in the range of 1 sample to 6 seconds at the context sample rate, and reject all other values.  This should be specified in the text.

The reasoning for this range is that we want to support the largest ScriptProcessorNode buffer size (16384) at the lowest sample rate (3000 Hz as sepcified in https://webaudio.github.io/web-audio-api/#sample-rates), to allow existing ScriptProcessorNode applications to be ported to AudioWorklet.  This reasoning should be included in a non-normative note in the spec.

mjwilson-google:
Fixes #2664 

Specify the valid range, and include references in the relevant places.  Also require that sample rate is set before the render quantum size in OfflineAudioContext.  Don't modify AudioContext constructor yet, since we have a separate issue tracking that (#2663).

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2381,6 +2381,8 @@ Dictionary {{AudioContextOptions}} Members</h5>
         <code>"default"</code> is passed, or to ask the User-Agent to pick a good
         <a>render quantum size</a> if <code>"hardware"</code> is specified.
 
+        See [[#render-quantum-sizes]] for the required supported range.
+
         It is a hint that might not be honored.
 </dl>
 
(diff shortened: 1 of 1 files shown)


mjwilson-google:
Note that WPTs have already landed:
- https://github.com/web-platform-tests/wpt/blob/master/webaudio/the-audio-api/the-audiocontext-interface/audiocontext-rendersizehint.html
- https://github.com/web-platform-tests/wpt/blob/master/webaudio/the-audio-api/the-offlineaudiocontext-interface/offlineaudiocontext-rendersizehint.html

hoch (approved):

padenot (approved):
We noticed during the 2026-02-17 call that there's a name discrepancy: `renderSizeHint` in the `Option` object, but `renderQuantumSize` on the `BaseAudioContext` and `Worklet` scope, but that might be too late to fix.

This PT looks good in any case

mjwilson-google:
Thank you!

> name discrepancy

I think this is because `renderQuantumSize` is the actual size as an integer, whereas `renderSizeHint` can take other values like 'default' and 'hardware'.  If we change the API to only accept integer size hints then I think we should also change the `Option` object name to `renderQuantumSize`.

As far as I know, Chromium is the only implementation experimenting with this API right now and we're still in origin trial.  So I don't think it's too late to make fixes yet.

Issue #2445, opened 2020-01-26
Incremental delivery of gargantuan amounts of audio data from OfflineAudioContext rendering


JohnWeisz:
**Describe the issue**

Currently, `OfflineAudioContext.startRendering` will unconditionally and _always_ result in the creation of an AudioBuffer containing the rendered audio, regardless of whether it is actually used or not.

This is a problem when used to render a massive duration, as it results in a gargantuan memory impact. Take the following settings for example:

 - Sample rate: 48kHz
 - Channels: 4
 - Duration: 4 hours

Rendering this monster, we have `48000 * 4 * 4 * 60 * 60 = 2_764_800_000` float32 values total, which amounts to more than **11 *Gigabytes*** of in-memory data (more specifically, we don't, because no end-user browser will allow this allocation without flags, some even _with_ flags, not speaking of systems which don't even have this much RAM in the first place).

**Where Is It**

https://webaudio.github.io/web-audio-api/#dom-offlineaudiocontext-startrendering, more specifically in step 4 of the OfflineAudioContext.startRendering process:

 > Create a new AudioBuffer, with a number of channels, length and sample rate equal respectively to the numberOfChannels, length and sampleRate values passed to this instance’s constructor in the contextOptions parameter. Assign this buffer to an internal slot [[rendered buffer]] in the OfflineAudioContext.

**Additional Information**

To avoid the above issue, OfflineAudioContext should "somehow" allow skipping this preliminary _forced_ memory allocation, and be able to deliver data at regular intervals.

The way I see it, there are several options here, and I personally think the [MediaRecorder.ondataavailable](https://w3c.github.io/mediacapture-record/#dom-mediarecorder-ondataavailable) is a good starting point, which is an already established example of incremental data delivery.

 1. Having OfflineAudioContext expose a ReadableStream that's pushed as data is available from rendering 
    - _Note: this has the potential to synergize _extremely_ well with the proposed [WebCodecs API](https://github.com/WICG/web-codecs/blob/master/explainer.md), which has a proposal to work out of a ReadableStream -- if the OfflineAudioContext exposed a ReadableStream, it would be possible to plug it into the WebCodecs API directly and acquire formatted audio files in an efficient and elegant manner, which is a very strong point for this option_
 2. Emitting some sort of a `dataavailable` event, possibly with an AudioBuffer for each event containing a chunk of the rendered data
    - _Probably unnecessary, as AudioWorklet is a viable alternative already_
[...]

PR #2675, opened 2026-06-05
OfflineAudioContext Incremental Rendering

gabrielsanbrito:
This PR proposes changes to the Web Audio spec to enable rendering audio from `OfflineAudioContexts` in chunks.

Github issue: https://github.com/WebAudio/web-audio-api/issues/2445
Feature explainer: https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/OfflineAudioContext/explainer.md

***
<a href="https://pr-preview.s3.amazonaws.com/gabrielsanbrito/web-audio-api/pull/2675.html" title="Last updated on Aug 6, 2026, 2:39 AM UTC (257a31c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-audio-api/2675/ad5d2ed...gabrielsanbrito:257a31c.html" title="Last updated on Aug 6, 2026, 2:39 AM UTC (257a31c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2668,9 +2668,12 @@ Methods</h4>
             occasion.
 
             <ol>
+                <li>Let <var>nFrames</var> be a number initialized to the value
+                of the length of {{[[rendered buffer]]}}.
+                
                 <li>Given the current connections and scheduled changes, start
-                rendering {{[[rendered buffer/length]]}} sample-frames of audio into
-                {{[[rendered buffer]]}}
+                rendering <var>nFrames</var> sample-frames of audio into
+                {{[[rendered buffer]]}}.
 
                 <li>For every <a>render quantum</a>, check and
                 {{OfflineAudioContext/suspend()|suspend}}
--- a/index.bs
+++ b/index.bs
@@ -2608,10 +2608,11 @@ Attributes</h4>
 <dl dfn-type=attribute dfn-for="OfflineAudioContext">
     : <dfn>length</dfn>
     ::
-        The total size of the audio render in sample-frames. This is the same as the
-        value of the <code>length</code> parameter for the constructor.
+        The total size of the audio render in sample-frames. This is the same
+        as the value of the <code>length</code> parameter for the constructor.
 
-        For undefined-length rendering, this attribute SHOULD be set to <code>null</code>.
+        For undefined-length rendering, this attribute SHOULD be set to
+        <code>null</code>.
 
     : <dfn>oncomplete</dfn>
     ::
@@ -2648,16 +2649,28 @@ Methods</h4>
 
                 <li>Let <var>promise</var> be a new promise.
 
-                <lI>Let <var>bufferLength</var> be a number initialized to the value of {{OfflineAudioContext/length}}.
+                <lI>Let <var>bufferLength</var> be a number initialized to the
+                value of {{OfflineAudioContext/length}}.
 
                 <ol>
-                    <li> If {{OfflineAudioContext/startRendering(chunkSize)/chunkSize}} is provided:
+                    <li> If {{OfflineAudioContext/startRendering(chunkSize)/chunkSize}}
+                    is provided:
                     <ol>
-                        <li> Let <var>renderedFrames</var> be the number of sample-frames already rendered by the {{OfflineAudioContext}}.
-                        <li> If <var>renderedFrames</var> + {{OfflineAudioContext/startRendering(chunkSize)/chunkSize}} is less than or equal to {{OfflineAudioContext/length}}, set <var>bufferLength</var> to {{OfflineAudioContext/startRendering(chunkSize)/chunkSize}}.
-                        <li> Otherwise, set <var>bufferLength</var> to {{OfflineAudioContext/length}} - <var>renderedFrames</var>.
+                        <li> Let <var>renderedFrames</var> be the number of
+                        sample-frames already rendered by the
+                        {{OfflineAudioContext}}.
+                        <li> If <var>renderedFrames</var> +
+                        {{OfflineAudioContext/startRendering(chunkSize)/chunkSize}}
+                        is less than or equal to {{OfflineAudioContext/length}},
+                        set <var>bufferLength</var> to
+                        {{OfflineAudioContext/startRendering(chunkSize)/chunkSize}}.
+                        <li> Otherwise, set <var>bufferLength</var> to
+                        {{OfflineAudioContext/length}} -
+                        <var>renderedFrames</var>.
                     </ol>
-                    <li> Otherwise, if {{OfflineAudioContext/length}} is <code>null</code>, set <var>bufferLength</var> to the <a>render quantum size</a>.
+                    <li> Otherwise, if {{OfflineAudioContext/length}} is
+                    <code>null</code>, set <var>bufferLength</var> to the
+                    <a>render quantum size</a>.
                 </ol>
 
                 <li>Create a new {{AudioBuffer}}, with a number of
@@ -2688,8 +2701,8 @@ Methods</h4>
             occasion.
 
             <ol>
-                <li>Let <var>numberOfFrames</var> be a number initialized to the value
-                of the length of {{[[rendered buffer]]}}.
+                <li>Let <var>numberOfFrames</var> be a number initialized to
+                the value of the length of {{[[rendered buffer]]}}.
                 
                 <li>Given the current connections and scheduled changes, start
                 rendering <var>numberOfFrames</var> sample-frames of audio into
@@ -2710,11 +2723,12 @@ Methods</h4>
                         <li>Resolve the <var ignore>promise</var> created by {{startRendering()}}
                         with {{[[rendered buffer]]}}.
 
-                        <li>If {{OfflineAudioContext/length}} is not <code>null</code>, 
-                        [=Queue a media element task=] to [=fire an event=] named
-                        {{OfflineAudioContext/complete}} at the {{OfflineAudioContext}} using
-                        {{OfflineAudioCompletionEvent}} whose `renderedBuffer` property is set to
-                        {{[[rendered buffer]]}}.
+                        <li>If {{OfflineAudioContext/length}} is not
+                        <code>null</code>, [=Queue a media element task=] to
+                        [=fire an event=] named {{OfflineAudioContext/complete}}
+                        at the {{OfflineAudioContext}} using
+                        {{OfflineAudioCompletionEvent}} whose `renderedBuffer`
+                        property is set to {{[[rendered buffer]]}}.
 
                     </ol>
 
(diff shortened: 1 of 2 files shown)


gabrielsanbrito:
@hoch @padenot @mjwilson-google, this is the first draft for `OfflineAudioContext` incremental rendering. PTAL when you have some time.

While writing this, we had some open questions that might be better to discuss here:
- In this draft I am proposing using `Infinity` to denote the undefined-length render scenario. However, `OfflineAudioContextOptions.length` is currently an unsigned long, which cannot be `Infinity`. What would be preferrable?
  - Swap all `Infinity` usage for `null`; or
  - Convert `OfflineAudioContextOptions.length` to an union type like `(long | float)` that only accept the `Infinity` value for `float`.
- When `OfflineAudioContextOptions.length` is `Infinity` and `chunksSize` is not provided to `startRendering()`, `chunkSize` defaults to the render quantum size. But the render quantum size is very small. Maybe, should we consider a more optimal default?
- Should `OfflineAudioContextOptions.length` be a multiple of `chunkSize`? I am assuming it doesn't need to.

(unfortunately, there is no preview available, but I think there might be something actually broken in the repo's CI pipeline)

gabrielsanbrito:
Audio WG meeting (06/11/2026) discussion:
1. Using `null` is less controversial given that the `length` property is a `long`
2. The render quantum size can be configured when the OfflineAudioContext is being constructed via the [renderSizeHint](https://webaudio.github.io/web-audio-api/#dom-offlineaudiocontextoptions-rendersizehint) parameter.
3. It's not trivial to enforce since `chunkSize` can change with every call. At the end of the rendering, if the `chunkSize` is larger than the number of remaining samples, the `OfflineAudioContext` just returns a smaller `AudioBuffer`.

index.bs
```diff
@@ -2706,9 +2701,9 @@ Methods</h4>
             </ol>
         </div>
 
-        <div>
-            <em>No parameters.</em>
-        </div>
+        <pre class=argumentdef for="OfflineAudioContext/startRendering()">
+            chunkSize: The size of the desired {{AudioBuffer}}.
```
mjwilson-google: I think we should say something like the number of sample-frames to render during this call.

index.bs
```diff
@@ -2500,9 +2480,10 @@ returned promise with the rendered result as an
 interface OfflineAudioContext : BaseAudioContext {
     constructor(OfflineAudioContextOptions contextOptions);
     constructor(unsigned long numberOfChannels, unsigned long length, float sampleRate);
-    Promise<AudioBuffer> startRendering();
+    Promise<AudioBuffer> startRendering(unsigned long chunkSize);
```
mjwilson-google: I think this should be `optional unsigned long chunkSize`

index.bs
```diff
@@ -2500,9 +2480,10 @@ returned promise with the rendered result as an
 interface OfflineAudioContext : BaseAudioContext {
     constructor(OfflineAudioContextOptions contextOptions);
     constructor(unsigned long numberOfChannels, unsigned long length, float sampleRate);
```
mjwilson-google, line 2503: If we make `length` nullable I think it would become `length?`

index.bs
```diff
@@ -2640,21 +2623,29 @@ Methods</h4>
...
                 {{OfflineAudioContext}} to <em>true</em>.
 
                 <li>Let <var>promise</var> be a new promise.
 
+                <lI>Let <var>bufferLength</var> be a number initialized to the value of {{OfflineAudioContext/length}}.
+
+                <ol>
+                    <li> If {{OfflineAudioContext/startRendering(chunkSize)/chunkSize}} is provided:
+                    <ol>
+                        <li> Let <var>renderedFrames</var> be the number of sample-frames already rendered by the {{OfflineAudioContext}}.
```
mjwilson-google: We may want renderedFrames to be an internal slot; then we can also specify when and how it is updated.

mjwilson-google (commented):
I added some comments.  Also, I think I fixed the Git workflow issue if you rebase the change on the latest commit.

mjwilson-google:
It looks like the git checks are now passing but the PR preview still isn't working; sorry.  I'm not sure why.

index.bs
```diff
@@ -2677,9 +2688,12 @@ Methods</h4>
             occasion.
 
             <ol>
+                <li>Let <var>nFrames</var> be a number initialized to the value
```
hoch: `numberOfFrames`

index.bs
hoch: Based on the discussion in the spec issue, was using `length` the final resolution?

index.bs
```diff
@@ -2607,9 +2608,11 @@ Attributes</h4>
 <dl dfn-type=attribute dfn-for="OfflineAudioContext">
     : <dfn>length</dfn>
     ::
-        The size of the buffer in sample-frames. This is the same as the
+        The total size of the audio render in sample-frames. This is the same as the
         value of the <code>length</code> parameter for the constructor.
 
+        For undefined-length rendering, this attribute SHOULD be set to positive infinity.
```
hoch: IMO - since we have an explicit `close()` in this change, rendering indefinitely can be done by calling `startRendering(chunkSize)` repeatedly.

Probably this is aligned with what @karlt proposed?

index.bs
```diff
@@ -2640,21 +2643,29 @@ Methods</h4>
...
 
                 <li>Let <var>promise</var> be a new promise.
 
+                <lI>Let <var>bufferLength</var> be a number initialized to the value of {{OfflineAudioContext/length}}.
+
+                <ol>
+                    <li> If {{OfflineAudioContext/startRendering(chunkSize)/chunkSize}} is provided:
+                    <ol>
+                        <li> Let <var>renderedFrames</var> be the number of sample-frames already rendered by the {{OfflineAudioContext}}.
+                        <li> If <var>renderedFrames</var> + {{OfflineAudioContext/startRendering(chunkSize)/chunkSize}} <= {{OfflineAudioContext/length}}, set <var>bufferLength</var> to {{OfflineAudioContext/startRendering(chunkSize)/chunkSize}}.
```
hoch: nit: Let's explain this logic in descriptive prose rather than using abbreviated math.

index.bs
```diff
@@ -2803,6 +2818,63 @@ Methods</h4>
         <div>
             <em>Return type:</em> {{Promise}}&lt;{{undefined}}&gt;
         </div>
+    
+    : <dfn>close()</dfn>
+    ::
+        Closes the {{AudioContext}}. This will not automatically release
```
hoch: Perhaps `OfflineAudioContext`?

hoch (commented):
Generally looks good, but I do have some suggestions. Please take a look at my comments.

Also it would be great if you can wrap the body text at column 80.

index.bs
gabrielsanbrito: The discussion has stalled since my last comment. AFAIU, it seems that everyone agreed on using `length`, but there is some divergence on *how* it should be used. Should it be `optional` or nullable?

I'd rather have `length` be nullable instead of `optional`. This way callers need to be a little more intentional when creating indefinite-length `OfflineAudioContexts`. What do you think?

index.bs
gabrielsanbrito: I am not sure if understand this comment :/ Would you mind clarifying? 

When I wrote this, I intended to have `length` be `Infinity` (after the last update it should be `null`) because an undefined render session has no defined length.

Do you mean that we shouldn't set `length` to `null` in the undefined length scenario and this should be tracked via an internal slot? For example, we need this info when determining the size of the returned `AudioBuffer`. The `OfflineAudioContext` needs to know whether we are in an undefined-length render or not:

```html
<li> Otherwise, if {{OfflineAudioContext/length}} is
<code>null</code>, set <var>bufferLength</var> to the
<a>render quantum size</a>.
```

index.bs
gabrielsanbrito: Thanks! Does it look good now? I was not sure if it was needed to make changes to the addition/subtraction.

index.bs
hoch: Then let's clarify the resolution first. I'd really liked to have a WG decision on this:

IIUC your design is to make `length` required, and the proposal from @karlt is to make it optional?

@karlt @padenot Can you all chime in here so we can make a decision?

index.bs
```diff
@@ -2500,9 +2500,10 @@ returned promise with the rendered result as an
 interface OfflineAudioContext : BaseAudioContext {
     constructor(OfflineAudioContextOptions contextOptions);
     constructor(unsigned long numberOfChannels, unsigned long length, float sampleRate);
-    Promise<AudioBuffer> startRendering();
+    Promise<AudioBuffer> startRendering(optional unsigned long chunkSize);
```
hoch: @karlt @padenot Please chime in here so we can make a group decision.

I don't think this argument is as controversial as `length` above, but I'd like to record the group's choice.

hoch (commented):
I'd like to have input from @karlt and @padenot on this.

index.bs
karlt: Confirming that I was proposing that indefinite repeated rendering be available even when an unsigned long value for the `length` member is explicitly provided.

Even an optional member needs a (default) value, for which `null` seems a reasonable choice to correspond to "unspecified", but we need not necessarily require a specific or explicit value for indefinite repeated rendering.

index.bs
karlt: The new `chunkSize` parameter to `startRendering()` now provides sufficient control over how much is rendered.
That makes the `length` dictionary member of OfflineAudioContextOptions redundant, but we need to keep the member in some form for backward compat.
We can choose whether it becomes optional or stays required.

I think allowing the length to be omitted is more convenient for developers than having to explicitly state `length: null`.
To me, a missing `length` member also seems better representative of unspecified length than an explicit null value for a required `length` member.
Still, I guess `null` would be the best default value for an optional `length` member, so nullable seems appropriate for the type.

IIUC @gabrielsanbrito proposed making the `length` member required to save the developer from accidentally creating an unlimited/unspecified length context.
I don't feel that an unspecified length context would be a problem.
I see the potential of a simpler API without an unnecessary length member as more valuable than any potential resource recovery immediately on rendering having reached a specified total length.

---

Now that startRendering() may be called multiple times, there is the question of which length the legacy `length` OfflineAudioContextOptions dictionary member should describe:

I see the the default length of one rendering chunk (if not specified as a startRendering() parameter) as being consistent with the legacy `length` member.
IIUC @gabrielsanbrito sees the limit on the sum of the lengths of rendering chunks as being more consistent.
Perhaps this choice is less important if the `length` member can be optional.

I would prefer that `startRendering(chunkSize)` have control over how much is rendered, as I would see a limit from a limiting OfflineAudioContextOptions member as action at a distance.

karlt (commented):
Thank you for putting this together and for the opportunity to comment.

index.bs
karlt: Is this asking about the `chunkSize` parameter?
I assume we want this to provide for variable chunk sizes.
I think this is better than a mutable attribute on the OfflineAudioContext, because the startRendering() parameter is directly associated with the rendering while the context attribute would be action at a distance.

This new parameter would need to be optional for backward compat.

index.bs
gabrielsanbrito: Thanks for writing this summary @karlt!

>I think allowing the length to be omitted is more convenient for developers than having to explicitly state length: null. To me, a missing length member also seems better representative of unspecified length than an explicit null value for a required length member.
Still, I guess null would be the best default value for an optional length member, so nullable seems appropriate for the type.

I am fine with this. Let's make `length` optional with a default value of `null`. Does it sound good? @hoch @padenot

index.bs
gabrielsanbrito: Quoting @karlt message from [here](https://github.com/WebAudio/web-audio-api/pull/2675#discussion_r3527848625): 

>I see the the default length of one rendering chunk (if not specified as a startRendering() parameter) as being consistent with the legacy length member.
IIUC @gabrielsanbrito sees the limit on the sum of the lengths of rendering chunks as being more consistent.
Perhaps this choice is less important if the length member can be optional.

>I would prefer that startRendering(chunkSize) have control over how much is rendered, as I would see a limit from a limiting OfflineAudioContextOptions member as action at a distance.

I want to point out that, if, moving forward, the only way to close an `OfflineAudioContext` is via calling `close()`, this is potentially breaking. Currently, after the render is finished, the `OfflineAudioContext` goes to the `"closed"` state and fires a `"statechange"` event - i.e. after rendering `length` sample-frames. This means that pages that expect this state change may not work properly since we won't automatically close the `OfflineAudioContext`. This would be prevented by my proposal.

_On a side note: I couldn't find this behavior in the spec (moving `OfflineAudioContext` to the `"closed"` state after the rendering is done), but this is what I noticed happening in both Chromium and Gecko._

Other concern is that, by allowing contexts to be solely manually closed, web pages could end up leaking resources - especially older ones unaware of `close()`. I recognize that this could be a minor concern since most WebAudio usage on the web is done through libraries and pages don't depend directly on WebAudio.

I am fine with adopting @karlt suggestions, although I think my concerns are valid. If we think that the breaking risk is small and that the performance implications are not that impactful, I feel like this is basically a matter of editorial preference. Both proposals should work for incremental and one-shot rendering scenarios. @hoch @padenot what is your opinion?

karlt: Ah, thank you for pointing that out. I had looked only at the spec and hadn't noticed that browsers were transitioning to "closed".
Even webkit is [doing similarly](https://searchfox.org/wubkat/rev/d221d71a6a560cb45bb3e9b23e526b6bcfa9e4b3/Source/WebCore/Modules/webaudio/OfflineAudioContext.cpp#259).
Treating non-null `length` as a total limit allows maintaining that behavior.
I understand the consensus is to stay with the `length` as limit approach.

padenot (approved):
Looks good!

mjwilson-google (approved):

hoch (approved):
Thanks for this contribution, @gabrielsanbrito! 

LGTM

